PR #15, opened 2014-01-29
Fixed #9, added option to open browser and default to yes

### LINKED ISSUE #9: Option to automatically open browser to server location
This should default to true

anglee:
Added a flag --do-not-open-browser, if the flag is absent (default), the default browser should open the beaker main entry page now.


anglee:
See #16

PR #39, opened 2014-02-06
Open ipython nb

anglee:
Added improvements to #19 base on suggestions from @jchendy.

### DIFF
--- a/core/src/main/web/plugin/menu/file_filesys.js
+++ b/core/src/main/web/plugin/menu/file_filesys.js
@@ -19,7 +19,7 @@
  */
 (function () {
     'use strict';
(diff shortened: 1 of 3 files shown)


core/src/main/web/plugin/menu/file_filesys.js
```diff
@@ -56,6 +68,7 @@
         if (!path) {
             return;
         }
+        var load = path.indexOf("http") === 0 ? loadFromHttp : loadFromFile;
```
scottdraves, line 71: i don't think that's a reliable enough way to identify URLs.  how about searching for http:// or https://

PR #82, opened 2014-03-10
A basic node.js plugin that executes javascript code in a remote node ap...

planetmcd:
...plication.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -75,3 +75,4 @@ task makeDist(type:Exec) {
 makeDist.dependsOn(":core:installApp")
 makeDist.dependsOn(":plugin:r:installApp")
 makeDist.dependsOn(":plugin:groovy:installApp")
+makeDist.dependsOn(":plugin:node:installApp")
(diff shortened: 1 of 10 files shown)


scottdraves:
this is a big improvement.  two remaining problems:
- still need the installApp target to run "npm install".  need to add your subdir to settings.gradle
- undefined produces an error.  try `var x=1;`

planetmcd:
On the first point, I added the npm install command to the shell script, with the thinking that it would remove an additional dependency on Java.  I can move it to gradle if that makes more sense.

scottdraves:
yea def move to gradle this is a build-time event.

scottdraves:
and when an error is reported, the background of the result cell should be red.
the innertype of the modelOutput.result should be set to "Error".

jchendy:
I'm excited to see how this turns out! FYI, I just published our user-facing expectations for new plugins on the wiki: https://github.com/twosigma/beaker-notebook/wiki/Eval-plugin-spec

scottdraves:
i just made these fixes myself on master.
i also just noticed syntax highlighting is not working, will get that too.

planetmcd:
Sorry I saw the note on the gradle changes, made the changes, and pushed before I saw your second note.

PR #98, opened 2014-03-26
Prepare for repo separation

anglee:


core/build.gradle
```diff
@@ -13,14 +31,15 @@ task (copyNginx, type: Exec) {
...
+bowerInstall.dependsOn npmInstall
 build.dependsOn bowerInstall
 run.dependsOn bowerInstall
 run.dependsOn jar
 run.dependsOn copyNginx
 
-dependencies {
-  compile project(':shared')
+compileJava {
+  options.compilerArgs << '-Xlint:deprecation'
```
scottdraves, line 41: eliminate duplication.  right now this is in core and every plugin.  it should only be in one place, at the top-level gradle file, or in shared.

shared/build.gradle
```diff
@@ -0,0 +1,17 @@
+apply plugin: 'java'
```
scottdraves, line 1: was this file meant to be deleted?

plugin/README.md
```diff
@@ -0,0 +1 @@
+Beaker Evaluate Plugins
```
scottdraves, line 1: need to at least point people at the core repo.

plugin/build.gradle
```diff
@@ -0,0 +1,19 @@
...
+if (hasProperty('evalPluginDir')) {
+  evalPluginDir = project.evalPluginDir;
+}
+if (System.properties['evalPluginDir']) {
+  evalPluginDir = System.properties['evalPluginDir'];
+}
+
+subprojects {
+  ext.evalPluginDir=evalPluginDir
+}
```
scottdraves: need newlines at the end of all lines.

plugin/groovy/build.gradle
```diff
@@ -1,10 +1,24 @@
...
 }
+
+compileJava {
+  options.compilerArgs << '-Xlint:deprecation'
+  options.compilerArgs << '-Xlint:unchecked'
+}
+
+if (hasProperty('evalPluginDir')) {
+  installApp.into evalPluginDir + '/groovy'
+}
```
scottdraves: newlines for all lines

plugin/groovy/build.gradle
```diff
@@ -1,10 +1,24 @@
+apply plugin: 'java'
+
+repositories {
```
scottdraves, line 3: remove duplication, this is core and every plugin, it used to be in just one place.

plugin/groovy/settings.gradle
```diff
@@ -0,0 +1,3 @@
+include 'plugin:jvm', 'shared'
+project(':plugin:jvm').projectDir = new File(settingsDir, "../jvm")
+project(':shared').projectDir = new File(settingsDir, "../../shared")
```
scottdraves, line 3: is setting the projectDir like this good practice?
this looks pretty weird, who else does it?

plugin/iruby/src/dist/vendor/ipython/kernel.js
```diff
@@ -0,0 +1,508 @@
+//----------------------------------------------------------------------------
```
scottdraves: remove duplication with ipython

plugin/jvm/build.gradle
```diff
@@ -1,3 +1,25 @@
...
+
+
+compileJava {
+  options.compilerArgs << '-Xlint:deprecation'
+  options.compilerArgs << '-Xlint:unchecked'
 }
+
+task installApp {
+  // do nothing, this is a library
+}
```
scottdraves: newline for all lines

plugin/jvm/settings.gradle
```diff
@@ -0,0 +1,2 @@
+include 'shared'
+project(':shared').projectDir = new File(settingsDir, "../../shared")
```
scottdraves: see previous similar comment.

plugin/node/src/dist/node.js
```diff
@@ -21,9 +21,9 @@
  */
 (function () {
     'use strict';
-    var url = "./plugin/evaluator/node.js";
+    var url = "./plugins/eval/node/node.js";
     var PLUGIN_NAME = "Node";
-    var COMMAND = "nodePlugin";
+    var COMMAND = "node/nodePlugin";
```
scottdraves, line 26: since every plugin has its own directory there's no need to pass in the name of this command.
they should all have the same name "startPlugin" or "<pluginname>Plugin".
no need to pass a parameter up.

shared/build.gradle
```diff
@@ -0,0 +1,17 @@
...
+  compile group: 'org.cometd.java', name: 'cometd-java-annotations', version: '2.7.0'
+  compile group: 'org.codehaus.jackson', name: 'jackson-jaxrs', version: '1.9.13'
+  compile group: 'org.apache.cxf', name: 'cxf-bundle-jaxrs', version: '2.7.7'
+  compile group: 'com.sun.jersey.contribs', name: 'jersey-guice', version: '1.17.1'
+}
+
+compileJava {
+  options.compilerArgs << '-Xlint:deprecation'
+  options.compilerArgs << '-Xlint:unchecked'
+}
```
scottdraves: newline for all lines.

scottdraves:
See above for the detailed comments.  At a high level, this PR cannot be merged until it has a working makeDist target.  I think the way plugins install into ~/.beaker/v1 is pretty weird though, and is going to make that harder.  Normally an application has a plugin dir of its own, and then users can add additional plugins in ~/.beaker.  That way the application can run out of a self-contained section of the filesystem.  With this PR, when you install the app it has to copy files to ~/.beaker/v1 and you can't run two versions at once without changing that v1 string.  RIght now we don't even need to "install" the app.

PR #146, opened 2014-04-16
Refactoring of evaluator managers

anglee:
This batch of changes includes:
- [evaluatorManager](https://github.com/twosigma/beaker-notebook/blob/release-refact/core/src/main/web/js/module/evaluatorManager.js) has be separated and completely rewritten into [bkEvaluatorManager](https://github.com/twosigma/beaker-notebook/blob/dev-refact-evaluator-managers/core/src/main/web/js/module/bkEvaluatorManager.js) and [bkEvaluatePluginManager](https://github.com/twosigma/beaker-notebook/blob/dev-refact-evaluator-managers/core/src/main/web/js/module/bkEvaluatePluginManager.js)
- the evaluator plugins format is changed to use AMD module style (requireJS)
- when adding a plugin by clicking button in the [bkNotebookEvaluators](https://github.com/twosigma/beaker-notebook/blob/dev-refact-evaluator-managers/core/src/main/web/js/module/bkNotebookEvaluators.js) (plugin manager), the notebook model store the plugin as plugin name instead of plugin urls, also, all the default notebooks and tutorial notebooks are changed to refer plugin with name instead of url


plugin/ipythonPlugins/src/dist/ipython/ipython.js
```diff
@@ -271,4 +269,20 @@
         onSuccess, onFail);
   };
   init();
-})();
+
+  exports.getEvaluatorFactory = function(settings) {
```
scottdraves: this looks like a pretty complicated piece of boilerplate.
can it be factored into bkHelper?

core/src/main/web/js/module/bkEvaluatePluginManager.js
```diff
@@ -0,0 +1,100 @@
...
+              loadingInProgressPlugins = _(loadingInProgressPlugins).filter(function(it) {
+                return it.url !== url;
+              });
+            });
+
+            return deferred.promise;
+          }
+        },
+        createEvaluatorThenExit: function(settings) {
+          console.log("createEvaluatorThenExit", settings.plugin);
```
scottdraves: pls remove debug console logging

plugin/ipythonPlugins/src/dist/ipython/ipython.js
anglee: I have simplified the code.

scottdraves: OK that's better but it's still the same 9 lines of code in every plugin.
Why not just a 1-line call to a function?

PR #183, opened 2014-04-25
Fixed #97 by adding spin check ensuring nginx restart before return

### LINKED ISSUE #97: race condition reconfiguring nginx
when we HUP nginx, must wait until reload is complete.
since the reconfig happens asynchronously and could take arbitrarily long,
it might not be ready when the plugin goes to try to use it.
maybe can spin on loading the URL in the plugins namespace until it doesn't 404 or something?

https://github.com/twosigma/beaker-notebook/blob/master/core/src/main/java/com/twosigma/beaker/core/rest/PluginServiceLocatorRest.java#L191

anglee:


core/nginx/nginx.conf.template
```diff
@@ -60,6 +60,11 @@ http {
         location /beaker/ {
             proxy_pass http://127.0.0.1:%(port_beaker)s/;
         }
+
+        location = /restart%(restart_id)s {
+            alias   html/50x.html;
```
scottdraves: using 50x.html is fine for testing but is confusing.  please create a separate 0-length file just called "stub" or "present" or something like that.

core/src/main/java/com/twosigma/beaker/core/rest/PluginServiceLocatorRest.java
```diff
@@ -56,6 +58,8 @@
 @Produces(MediaType.APPLICATION_JSON)
 @Singleton
 public class PluginServiceLocatorRest {
+  private static final int RESTART_ENSURE_RETRY_MAX_COUNT = 60;
+  private static final int RESTART_ENSURE_RETRY_INTERVAL = 500; //ms
```
scottdraves: interval should be 50, and count 600.

Issue #179, opened 2014-04-23
Require JRE instead of JDK for end users


jchendy:
The CLI aliases we use to launch our Java app don't get updated by a JRE install/upgrade, so users have to install JDK. This process is more confusing and error prone, and is unnecessary for users who already have JRE 7. Let's see if we can eliminate the JDK dependency.

scottdraves:
we might want to consider 1.6 as well, we use little of 1.7

jchendy:
Yeah, it would be a big improvement for users if we could remove the 1.7 dependency.

scottdraves:
on mac to make it work with JRE instead of JDK basically we just need to do
export JAVA_HOME=/Library/Internet\ Plug-Ins/JavaAppletPlugin.plugin/Contents/Home
in the beaker.command script

scottdraves:
fixed on mac https://github.com/twosigma/beaker-notebook/pull/217

scottdraves:
done on windows too

PR #217, opened 2014-05-04
fix for Issue #179

scottdraves:
tested on 2 macs one with jdk, one with jre.
needs testing on ubuntu.


scottdraves:
tested on ubuntu, no issues (but does not fix it there yet).

PR #249, opened 2014-05-08
Fix a bug preventing creating new notebook when a new noteobok is opened

anglee:


jchendy:
Hooray, this has been bothering me!

PR #262, opened 2014-05-13
Notebook save/open dialog boxes

### LINKED ISSUE #247: Improve save/load behavior
**Save as file dialog**
- Show full file path below input box (see mockups)
  - If user has input a relative path, expand PWD
- Append .bkr if user doesn’t include it
- If user selects a directory, show dialog with “{selected path} is a directory. Please choose a different location” Include a close button that takes user back to ‘save as dialog’ with the same state as right before the clicked save.
- If the file exists, show dialog with “{selected path} already exists. Overwrite?” with buttons ‘Overwrite’ and ‘Cancel’ Cancel should take user back to ‘save as dialog’

**Open file dialog**
- Checkbox to show only .bkr files, checked by detault (see mockup)

**Save errors**
- Show dialog with title ‘Save failed’, text save failed + short exception, scroll pane with full exception, and a ‘close’ button.

**Open errors**
- For errors accessing file, use the same type of dialog as save errors
- If the content isn’t a notebook, show a dialog to that effect (currently this is done with alert(), let’s use a Bootstrap dialog for consistency)

save-as-file-pwd

save-as-file-pwd-2

anglee:
Fix #247

### DIFF
--- a/core/src/main/web/app/mainapp/mainapp.js
+++ b/core/src/main/web/app/mainapp/mainapp.js
@@ -235,7 +235,7 @@
                 },
                 function() {
                   deferred.reject();
-                }, "Overwrite", "Cancel");
+                }, "Overwrite", "Cancel", "btn-danger");
             return deferred.promise;
           };
 
--- a/core/src/main/web/app/mainapp/mainapp.js
+++ b/core/src/main/web/app/mainapp/mainapp.js
@@ -302,7 +302,7 @@
           };
           var saveDone = function(ret) {
             bkSessionManager.setNotebookModelEdited(false);
-            bkSessionManager.updateNotebookUri(ret.uri, ret.uriType, false);
+            bkSessionManager.updateNotebookUri(ret.uri, ret.uriType, false, "bkr");
             document.title = bkSessionManager.getNotebookTitle();
             showTransientStatusMessage("Saved");
           };
--- a/core/src/main/web/app/mainapp/mainapp.js
+++ b/core/src/main/web/app/mainapp/mainapp.js
@@ -160,7 +160,7 @@
                 } else {
                   bkCoreManager.showErrorModal("Failed to open " + notebookUri
                       + " because format " + format
-                      + " was not recognized.");
+                      + " was not recognized.", "Open Failed");
                 }
               }
               var fileLoader = bkCoreManager.getFileLoader(uriType);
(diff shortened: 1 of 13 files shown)


core/src/main/web/app/mainapp/mainapp.js
```diff
@@ -206,51 +206,142 @@
         };
 
         var _impl = (function() {
-          var _saveNotebook = function() {
-            showStatusMessage("Saving");
+
+          var promptUriChooser = function(initUri) {
```
anglee, line 214: see https://www.lucidchart.com/documents/view/d2c35b6f-cbb6-48bd-be0d-80c86bb97a04

jchendy:
Help and settings menus seem to be missing from control panel.

jchendy:
Also 'open -> .bkr' is missing.

jchendy:
I found out I had the same problem on master. Ang suggested deleting beaker.conf.json, which worked.

jchendy:
As per the spec, the buttons on the 'file exists' dialog should say 'Overwrite' and 'Cancel' instead of yes/no. Please also use Bootstrap button styles with btn-danger for 'Overwrite' and the default class for 'Cancel'

jchendy:
This may have been an omission in the spec, but names like '~/file' don't expand properly in the save dialog. Can we have it translate into '/User/jeff/file.bkr' etc

jchendy:
Two issues related to opening .ipynb:
- Open an IPython file and click save. It wipes out the file with a Beaker file. We had previously decided that we don't want to overwrite converted files, so this should bring up the 'save as' dialog
- Open an IPython file, save it somewhere else, and load that file. I get 'TypeError: Cannot read property '0' of undefined'

jchendy:
Also, for the overwrite dialog, can you please make 'overwrite' the right button and add the class btn-danger?

jchendy:
Please swap the order of 'overwrite' and 'cancel' buttons. We should always have the 'action' or 'go forward' button on the right.

jchendy:
1) Open a notebook
2) Try to open an invalid file
3) After you close the dialog, you end up with a corrupted version of the original notebook that can't run any code

Let's open the control panel instead.

jchendy:
On the 'save failed' dialog, can we please have the blue 'close' button in the bottom right for consistency with our other dialogs?

jchendy:
Sorry, found another thing. The 'show .bkr files only' label should be clickable and toggle the checkbox.

PR #312, opened 2014-06-03
initial code cell style pass

samccone:
Implements a basic version of the codecell styling
that does not remove any functionality however brings it closer
inline with the implemented design.


jchendy:
Sweet, it's looking good! I'd recommend a couple things before merging:
- Get rid of the background colors for different languages
- Add some other indication of when the evaluator is done loading (I don't remember if Ljubo mocked this up; something like a spinner next to the language name would work)
- Add the hamburger menu to each cell with the same contents that the bracket menus currently have

samccone:
> Get rid of the background colors for different languages

I did not drop the colors yet because there is a bunch of language spread across the app saying "when i turn <color> I am ready".

This goes along with 

> Add some other indication of when the evaluator is done loading (I don't remember if Ljubo mocked this up; something like a spinner next to the language name would work)

I am not 100% what this indicator is yet, I will need to ask Lub about this.

> Add the hamburger menu to each cell with the same contents that the bracket menus currently have

This is the next step in the iteration from me, I will just keep working on this branch if you would like. This PR is more of a small visible step with low complexity cost that seems like a high value add.

thanks for the feedback @jchendy

anglee:
Hi, @samccone. Looks like a good start!
Not sure if it is considered part of this round of changes, I noticed the output cell menu still shows up when clicked on the white space to the right of `<pre></pre>`

samccone:
per feedback here is the next iteration.



I have omitted the color change since currently there is no way to know when a cell is ready to be evaluated. 

The next steps for me is going to be to add some tests around the menu functionality.

jchendy:
Cool. It's looking really good. I found one small bug: when you uncheck 'show input cell' from the hamburger menu, it hides both the input and output.

Also a usability thing: it looks like you can click about 20 pixels to the left of the hamburger to open the menu, but you can't click to the right at all. It should be even on both sides. And there should probably be some kind of mouseover effect on the menu icon (either darken the color or draw a button around it).

jchendy:
Also, if you open the tutorial notebook, the first R cell has two menu icons. I think there's an HTML cell (which I accidentally hid completely before saving the notebook) that's somehow being merged into the R cell.

samccone:
Cool, thanks for the comments @jchendy 
I have updated the branch per your comments and added test coverage to prevent regressions for these features.

https://github.com/twosigma/beaker-notebook/commit/0dce75dbbe4dc136879a08158e0fcb76c4382133#diff-8f72f81380db6def73bc17de6869d12aR31

https://github.com/twosigma/beaker-notebook/commit/23816e4689f1041ea334f79a232507aa1df26bf2#diff-8f72f81380db6def73bc17de6869d12aR20

jchendy:
I did a fresh clone, and I'm seeing some issues with your branch. Master seems to work fine.

screen shot 2014-06-05 at 11 54 07 am

samccone:
Ah yes indeed, looks like my browser was fooling me due to chrome liking to aggressively cache the templates.

Digging in.

samccone:
just FYI here was the breaking line
https://github.com/twosigma/beaker-notebook/commit/0dce75dbbe4dc136879a08158e0fcb76c4382133#diff-3505310d2776329a2110f5be8aab8cfeR18

Since codemirror eagerly inits and the element was not on the page everything fell apart. Updating now. Great catch @jchendy

samccone:
great all updated :)

jchendy:
When trying to pull your two most recent commits, I've gotten "Automatic merge failed; fix conflicts and then commit the result." both times. I've never seen that before. Any idea why it would be happening here?

samccone:
yep, because i force pushed over the branch, you will want to `git fetch && git reset --hard origin/sjs/restyle-code-menu`

jchendy:
Cool. Got it running again. Just a couple more comments:

With the removal of the bracket, we lost the visual indication of 'initialization cells.' I think Ljubo had some sort of icon to represent that. We should be sure to get that in place before we do a release.

The vertical alignment of the menu icon looks a little off:

menu-align

samccone:
Here is just an update for what it looks like now


The missing piece is the cell grouping indicator on the left.

samccone:
Just an update the new base header styles are in for now


Obviously this will be iterated on further.

Waiting for the nesting styles next and then hopefully we can get this branch into main.

samccone:
Updated again with the cell depth indicator

samccone:
I think this is all we need for this to be able to be merged without breaking anyones experience.

samccone:
codecell loading indicators are now in here

samccone:
closing for now until we are ready to merge.

PR #314, opened 2014-06-03
Beaker notebook sharing v2

anglee:


scottdraves:
why are the plugins changed?

anglee:
Scott, you were right. I have moved the logic from individual plugins into the evaluator manager.

Issue #316, opened 2014-06-05
gradle can do incomplete build


scottdraves:
~/beaker-notebook/core(public-server)$ gradle installApp
:shared:compileJava UP-TO-DATE
:shared:processResources UP-TO-DATE
:shared:classes UP-TO-DATE
:shared:jar UP-TO-DATE
:compileJava
:processResources UP-TO-DATE
:classes
:jar
:startScripts
:installApp

BUILD SUCCESSFUL

Total time: 7.412 secs
~/beaker-notebook/core(public-server)$ cd config/builds/dev
~/beaker-notebook/core/config/builds/dev(public-server)$ gradle installApp
:shared:compileJava UP-TO-DATE
:shared:processResources UP-TO-DATE
:shared:classes UP-TO-DATE
:shared:jar
:core:compileJava
:core:processResources UP-TO-DATE
:core:classes
:core:jar
:core:startScripts
:core:installApp
:plugin:jvm:compileJava
:plugin:jvm:processResources UP-TO-DATE
:plugin:jvm:classes
:plugin:jvm:jar UP-TO-DATE
:plugin:groovy:compileJava
:plugin:groovy:processResources UP-TO-DATE
:plugin:groovy:classes
:plugin:groovy:jar UP-TO-DATE
:plugin:groovy:startScripts
:plugin:groovy:installApp
:plugin:ipythonPlugins:installApp UP-TO-DATE
:plugin:jvm:installApp UP-TO-DATE
:plugin:node:syncFiles UP-TO-DATE
:plugin:node:installNodePackagesnpm WARN package.json beaker-node-plugin@0.0.0 No repository field.
npm WARN package.json beaker-node-plugin@0.0.0 No README data

:plugin:node:installApp
:plugin:r:compileJava
:plugin:r:processResources UP-TO-DATE
:plugin:r:classes
:plugin:r:jar UP-TO-DATE
:plugin:r:startScripts
:plugin:r:installApp

BUILD SUCCESSFUL

Total time: 11.513 secs

scottdraves:
to clarify:

If you "gradle run" from the core directory, the plugins are not compiled.  This could result in executing the project in an inconsistent state.

Another problem is from the root of the project "gradle run" does nothing:

```
$ git clone https://github.com/twosigma/beaker-notebook.git
$ cd beaker-notebook
$ gradle run

FAILURE: Build failed with an exception.

* What went wrong:
Task 'run' not found in root project 'beaker-notebook'.

* Try:
[...]

PR #329, opened 2014-06-12
Add top-level gradle build file that just redirects to our existing buil...

scottdraves:
...d system.  This addresses part of Issue #316, enough to get working with Travis.  Also add a .travis.yml file that should get us building there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: java
+jdk:
+ - oraclejdk7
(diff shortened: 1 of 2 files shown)


anglee:
So is the build script in core/config/builds deprecated now?

scottdraves:
no i just made the root redirect there as a quick fix.

scottdraves:
but yes, after this goes in i will simplify the documentation now that cd down there is not necessary.

anglee:
OK. This is fine, given the benefit from integration with travis.

I do want to point out the original design was there to make repositories separation trivial.

And yes, please get rid of core/config/builds scripts soon, otherwise, we need docs or comments in files to ensure they are sync'ed when developers touch them

PR #335, opened 2014-06-13
Update dill version per chromedriver bug.

samccone:


samccone:
thanks @scottdraves :+1:

PR #355, opened 2014-06-24
Fix #314, Add tick event sending to ga track

anglee:


scottdraves:
don't forget to update the wiki page.

PR #358, opened 2014-06-25
URL behavior changes

anglee:
Make hitting '/open' and '/session/new' URLs automatically be redirected to '/session/:sessionId' so refresh the page will not create new sessions

Also, make sure the search(query) parameters from the previous navigation are cleared when starting a new location change.

### DIFF
--- a/core/src/main/web/app/beaker.js
+++ b/core/src/main/web/app/beaker.js
@@ -101,28 +101,19 @@
       'bk.debug'
     ]);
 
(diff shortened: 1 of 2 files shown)


scottdraves:
please explain what the "minor change" is and why it was done.

anglee:
Move the function into the function to minimize its scope.
Use underscore.js to shorten and improve code readability.

PR #376, opened 2014-07-09
Fix #101, implement new way to handle beaker conf with user prefs

### LINKED ISSUE #101: do not copy beaker.conf.json into ~/.beaker
right now it cannot be updated.
the user should have a file like this that is merged with the copy in the application.
the merge process should allow them to remove items as well as add items.
it could have negative lists in addition to positive lists, or it could have a list where items are optionally noted as negative like ["addx", "addy", {"remove": "z"}]

anglee:
Before this change the beaker.conf.json file will only be copied
to the .beaker directory, when one file doesn't exist already. After
this change, such copying operation won't take place anymore. Instead, a
beaker.pref.json file in the dot directory will be used to record user
preferences. And the final configuration beaker uses would be the merged
result of the beaker.conf.json (in install dir) overridden by
beaker.pref.json (in .beaker). Also, removing elements from arrays is
possible by by prefixing array property name with "_remove-".


scottdraves:
can you give an example showing how to add and remove files?

anglee:
I wrote up an wiki page: [Configuration-and-Preference](https://github.com/twosigma/beaker-notebook/wiki/Configuration-and-Preference)

scottdraves:
thanks that looks great.  i don't understand what the underscore is for though, seems like the files would be easier to understand without it, especially the combination of underscore and dash is hard for people to process.

scottdraves:
I started to expand that wiki page and was confused because it uses "cell-menu-plugin" but the code seems to use "notebook-cell-menu-plugins", which is right?

Also, I think it would be really helpful to explain here how to add a new file.
Where would a user put it so it will be served and what path do they use to refer to it in the prefs?

anglee:
The code is right. The wiki page is just using a random example showing how removal works. I should have used 'foo' there. The wiki page is meant for devs in answering questions like the one you raised.

I completely agree that we should have helpful documentation for users too, We want to improve our documentation situation in general for many other things.

scottdraves:
Thanks, OK that's a step in the right direction but I am not satisfied.
Must we really subject our users to writing configuration files like this?

```
  "notebook-cell-menu-plugins": ["D"],
  "remove--notebook-cell-menu-plugins": ["B"]
```

This is how it should be (note without a double dash):

```
  "notebook-cell-menu-plugins": ["D"],
  "remove-notebook-cell-menu-plugins": ["B"]
```

Users will just be confused by a double dash and it serves no purpose in the code.  I am not aware of any other software configuration files that do something like this.  Note that this will be hard to change once we release it.

core/src/main/java/com/twosigma/beaker/core/rest/UtilRest.java
```diff
@@ -131,60 +132,111 @@ public String getDefaultNotebook() {
...
+
+    // TODO, assume the url is a file path for now.
+    java.nio.file.Path configFile = Paths.get(configFileUrl);
+    java.nio.file.Path preferenceFile = Paths.get(preferenceFileUrl);
+    try {
+      JSONParser parser = new JSONParser();
+      JSONObject configJsonObject =
+          (JSONObject) parser.parse(this.utils.readFile(configFile));
+      JSONObject preferenceJsonObject =
+          (JSONObject) parser.parse(this.utils.readFile(preferenceFile));
```
scottdraves, line 223: can we call clean() on the files (as defined in UtilRest.java) so they can have comments like in the default notebook?

anglee:
The prefix indicates a property is not a normal(no-prefix) property, which will cause the default behavior (append to existing list). Having the prefix standout so that it is clear that it doesn't belong to the rest of the property name make this fact harder to miss.

PR #382, opened 2014-07-12
V0.55 -- bring release branch back to master

scottdraves:
now that we have rc1:
this closes Issue #367 by including nginx, jre, and python in the distro.  python is checked into the repo, nginx is copied from developer config, jre is downloaded from oracle manually at build time.

also includes script to configure ubuntu from stock to run.

build process is documented in the README.mds

base version string off git branch instead of tag.

one commit to note is: https://github.com/twosigma/beaker-notebook/commit/3315c11dbab0f2139b7f708fd3735814de6f0231
on mac, beaker.command has to wrap running the server in "bash -cl" to create a login shell.  Needs to be added because when run from the Finder, no login shell has yet been created and so the user's .bash_profile hasn't been read, and that's where anaconda sets up the PATH.  Also, who knows what other settings they may have in their and want.

Also on mac, move where DYLD_LIBRARY_PATH from beaker.command to Java like the rest of the envars.


anglee:
Merging after confirmed gradle check passed locally

PR #410, opened 2014-07-27
Swap underscore.js for lo-dash

anglee:
A difference btw underscore and lo-dash is that lo-dash uses _() for chaining, while with underscore, _.chain() needs to be called explicitly. Replaced a bunch of uses of _().foo to _.foo() to avoid confusion.


scottdraves:
Let's wait until after 1.0 before making nonessential improvements.

Issue #89, opened 2014-03-19
notebook namespace


scottdraves:
We need a way to communicate between cells from different evaluators.  I propose this be done with a notebook namespace: a collection of variables that are part of the document model.  Their values would be JSON.

Access to the variables would be by a module "beaker" written for each plugin and loaded automatically.  As a baseline we would use strings for the variables with a basic get/set API:  `beaker.get("x")` and `beaker.set("x", 1)`.  Some languages in particular Javascript might be able to better and you could just say `beaker.x` and `beaker.x = 1`.  Maybe, but we could live without it.

The big question is where the namespace is stored.  Since the namespace is part of the document model (including for example be saved in the notebook), naturally it should primarily exist there, in the JS in the web browser.  But if code in plugin X accesses a value, it would have to make a REST call to Beaker core, which would then use comet to get the value.  This round-trip could be expensive.  So I propose keeping a copy of the namespace in Beaker Core, and using comet to keep the two copies synchronized.  REST calls are not as fast as in-memory access but avoiding going over the WAN could be a huge win.

Mostly this synchronization can be done after the caller returns.  The danger is that you will write to one side and then evaluate a cell on the other side that reads it before it has been updated.  Without some kind of locking this would be possible because different channels are used for each plugin and the core.  So when accessing the namespace you would have to block until you were guaranteed to get the right value.  Along with the cell evaluation results you could return a flag that said there might be namespace changes coming.  Similarly with the cell evaluation call you can have a flag that there might be changes coming.  In both cases you only block if you need access.

This is just a performance optimization, so a first pass at implementation should probably stick with keeping one copy in the JS.

It would be possible to keep it just in the Core and not in the client.  I think they are most likely to be accessed from JS though (for application development and for saving the document).  Another possibility is to keep a copy in every plugin, but that would be expensive in memory, though it does give very fast access.

The big alternative to this is using cell evaluation results as the bus.  Ie in one cell referring to the output of another cell.  This has two problems, one is there's no obvious name to use, and two because values are not explicitly shared the optimization of duplicating the namespace becomes impractical.

scottdraves:
On the client the minimum JS that completes this should be pretty straightforward, just allowing the extra information in the notebook model and responding to the cometd requests to access it.

[...]

PR #416, opened 2014-07-28
implement Notebook Namespaces (Issue #89)

scottdraves:
This work followed the proposal closely: the notebook data model is optionally extended with a section containing a namespace (dictionary) of JSON values.  The names are restricted to safe identifiers (alphanumeric plus underscore, sans initial digit). The server exports new REST endpoints for manipulating these values.  Each Plugin language is extended with a module ("beaker") that exports functions to do the work.  The module is initialized by the plugin JS, and the session ID is provided to it.  The plugin code uses HTTP to communicate with the REST endpoints on the Core server (session must be provided to each call to specify which notebook it refers to).  The implementation in the core server uses Comet to access the namespace on the client where the notebook lives.  Since comet is a one-way protocol, in order to get data back from the client we use a SynchronousQueue to get the reply to the right place, crossing threads.

The functions that are exported are:

```
set(var, val)
set_fast(var, val) # set and return without waiting for write to complete
unset(var) # remove from the namespace
get(var) # and return it
```

Includes implementations for Python, R, Groovy, and Javascript.  The groovy implementation is actually in Java and is in the JVM module so it should be accessible from other JVM languages like Scala or Clojure.

I updated the Tutorial notebook to demonstrate the use of this feature, and especially used it to simplify the final d3.js example.

Remaining to do:
- more languages
- make user-extensible types
- identify dataframes (and other types) automatically.  right now in python you have to call a constructor to convert the json.

In addition, I discovered a few files were missing the copyright/license banner or had incorrect copyright banners.  These are fixed.

Here are the examples from the new section of the tutorial:
screen shot 2014-07-28 at 6 07 12 pm


plugin/groovy/src/dist/groovy.js
```diff
@@ -76,7 +76,7 @@ define(function(require, exports, module) {
             console.log("failed to create shell", arguments);
           });
     },
-    evaluate: function(code, modelOutput) {
+    evaluate: function(code, modelOutput, cb) {
```
anglee: evaluate() uses promise to do async, it shouldn't take a callback. In other words, the action to follow should be done like: evaluate(...).then(action) instead of evaluate(..., action)

plugin/groovy/src/dist/groovy.js
```diff
@@ -173,9 +176,12 @@ define(function(require, exports, module) {
           }
           settings.shellID = id;
           self.settings = settings;
-          if (doneCB) {
-            doneCB(self);
-          }
+	  var initCode = "import com.twosigma.beaker.NamespaceClient\n" +
```
anglee: This line is using tab instead of whitespace. Please do a global check about whitespace in other files too.

core/src/main/java/com/twosigma/beaker/core/rest/NamespaceRest.java
```diff
@@ -0,0 +1,72 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package com.twosigma.beaker.core.rest;
+
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import org.codehaus.jackson.map.ObjectMapper;
+import java.io.IOException;
+import java.util.List;
```
anglee: Remove unused import

plugin/groovy/src/dist/groovy.js
scottdraves: thanks, done.

plugin/groovy/src/dist/groovy.js
scottdraves: done

core/src/main/java/com/twosigma/beaker/core/rest/NamespaceRest.java
scottdraves: done

core/src/main/java/com/twosigma/beaker/core/rest/NamespaceService.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  public NamespaceBinding get(String session, String name)
+    throws RuntimeException, InterruptedException
+  {
+    Map<String, Object> data = new HashMap<String, Object>(1);
+    data.put("name", name);
+    ServerChannel channel = getChannel(session);
+    if (null == channel) {
+	System.err.println("channel not found for session " + session);
```
anglee: please check whitespace

scottdraves: done

shared/src/main/java/com/twosigma/beaker/shared/NamespaceBinding.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+package com.twosigma.beaker.shared;
+
+import org.codehaus.jackson.annotate.JsonAutoDetect;
+
+@JsonAutoDetect
+public class NamespaceBinding {
+
+  public String name;
```
anglee: Can these member variables be private?

scottdraves: done

plugin/groovy/src/dist/groovy.js
anglee: I think you forgot to remove the last parameter 'cb' in evaluate functions in groovy.js and r.js

plugin/r/src/dist/beaker/R/beaker.r
```diff
@@ -0,0 +1,62 @@
...
+}
+
+set4 <- function(var, val, unset, sync) {
+  req = paste('http://127.0.0.1:',Sys.getenv("beaker_core_port"),
+           '/rest/namespace/set', sep='')
+  opts = list(userpwd=pwarg, httpauth = AUTH_BASIC)
+  if (unset) {
+    reply = postForm(req, style='POST', name=var, session=session_id, sync=sync, .opts=opts)
+  } else {
+    reply = postForm(req, style='POST', name=var, value=toJSON(val), session=session_id, sync=sync, .opts=opts)
```
anglee, line 33: `toJSON()` turns type "double" and "character" into list. For example,
one does this in R:
`beaker::set("foo", "bar")`
this appears in the name space :
`{...   "foo": ["bar"], .... }`

where the expected is
`{...   "foo": "bar", .... }`

scottdraves, line 33: yea that's just how R works.  There is no distinction between a scalar and a vector of length 1.

plugin/groovy/src/dist/groovy.js
scottdraves: done

plugin/r/src/dist/beaker/R/beaker.r
anglee, line 33: Got it. Thanks

PR #433, opened 2014-08-04
Move landing page new notebook to the bottom

samccone:


anglee:
It seems there is a sizing issue when displaying a table:
image
I am using Chrome, Zoom=100%

anglee:
Insert Cell should be hidden when notebook is locked
pasted_image_8_6_14__9_44_am

anglee:
We want to distinguish an absent output (deleted or not evaluated) from a empty or hidden output.
When a output cell is deleted, it should be completely gone with no output cell box and no text 'Output Hidden’.

core/src/main/web/app/mainapp/components/notebook/cell-directive.js
```diff
@@ -32,7 +32,19 @@
   module.directive('bkCell', function(bkUtils, bkSessionManager, bkCoreManager) {
     return {
       restrict: 'E',
-      template: '<div class="bkcell">' +
+      template: '<div ng-class="isLocked() && \'locked\' " class="bkcell {{cellmodel.type}}" index="$index">' +
```
anglee: What is `index="$index"` for? Was it used anywhere?

samccone: it is used to generate the hierarchy index

anglee: Can you provide an example of how hierarchy index is used?

samccone: Sure thing

see here
https://github.com/twosigma/beaker-notebook/commit/76a088ae1dd847873b1c317bcb4e9533d2159f70

core/src/main/web/app/mainapp/components/notebook/cell-directive.js
```diff
@@ -83,6 +104,14 @@
           }
         };
 
+        $scope.getFullIndex = function() {
+          if ($scope.$parent.getNestedLevel) {
```
anglee, line 84: Nothing needs to change at this point if there is going to be much work. I just want to point out that,
Ideally the cell hierarchy should come from bkNotebookCellModelManager (bkSessionManager.getNotebookCellOp()), that is, the model (as in MVC). $scope.getNestedLevel (should have been deprecated) and $scope.getParentId() should have been refactored to the cell model manager instead of relying on the view logic.

Also, note although this works, this is finding getNestedLevel through the prototypal chain. As commented in the definition of getNestedLevel(), the parent is $scope.$parent.$parent (hasOwnProperty("getNestedLevel") === true).

core/src/main/web/app/mainapp/components/notebook/cell-directive.js
anglee: I see. The so the hierarchy index was referred as $scope.index in the isolated scope? If so, it should have already been made available with `scope: { index: "="}` and `<bk-cell ng-repeat="..." index="$index">`  (e.g. in sectioncell.html) and shouldn't need to add `index="$index"` to the `div` in the template. Or am I missing something?

core/src/main/web/app/mainapp/components/notebook/cell-directive.js
samccone, line 84: :+1: you are right, all fixed

core/src/main/web/app/mainapp/components/notebook/cell-directive.js
```diff
@@ -147,32 +194,20 @@
...
-        div.on("click", clickHandler);
-        var mouseMoveHandler = function(event) {
-          if (bkUtils.getEventOffsetX(div, event) >= div.width()) {
-            div.css('cursor', 'pointer');
-          } else {
-            div.css('cursor', 'default');
-          }
-        };
-        div.mousemove('mousemove', mouseMoveHandler);
+      link: function($scope, $element, attrs) {
```
anglee: Please name arguments of link functions (scope, element, attrs) without `$` prefix.

core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
```diff
@@ -361,34 +376,18 @@
...
-          menu.css("left", clicked ? event.clientX - 150 : "");
           menu.find('.dropdown-toggle').first().dropdown('toggle');
         };
-        var inputClickHandler = function(event) {
-          if (bkUtils.getEventOffsetX(inputMenuDiv, event) >= inputMenuDiv.width()) {
-            scope.popupMenu(event);
-            event.stopPropagation();
-          }
-        };
-        inputMenuDiv.click(inputClickHandler);
```
anglee, line 383: Since these event handlers are removed and not used, they should also be removed in the `$destroy` handler, (i.e. line [417](https://github.com/twosigma/beaker-notebook/blob/9aafc8f01b3d9b90631e2ce2e5115efca57917bc/core/src/main/web/app/mainapp/components/notebook/codecell-directive.js#L417), [418](https://github.com/twosigma/beaker-notebook/blob/9aafc8f01b3d9b90631e2ce2e5115efca57917bc/core/src/main/web/app/mainapp/components/notebook/codecell-directive.js#L418))

anglee:
There is a gap between menu and a submenu, which prevents items in the submenu being selected.
pasted_image_8_6_14__1_46_pm

core/src/main/web/app/mainapp/components/notebook/sectioncell.html
```diff
@@ -14,11 +14,12 @@
     limitations under the License.
 -->
 <div ng-hide='cellmodel.hideTitle'>
-  <i class='fa fa-plus-square-o bksectiontoggleplus' ng-click='toggleShowChildren()' ng-hide='isShowChildren()'></i>
-  <i class='fa fa-minus-square-o bksectiontoggleminus' ng-click='toggleShowChildren()' ng-show='isShowChildren()'></i>
+  <span class='bksectiontoggleplus section-toggle' ng-click='toggleShowChildren()' ng-hide='isShowChildren()'>+</span>
```
anglee: This is probably minor, but can we use icon for the plus/minus sign? (fa-plus, fa-minus). because '+' is wider then '-' for non-monospace fonts, there is a slight horizontal shifting on the section title text when toggling.

anglee:
I found if I start with the default notebook (New Notebook), no output cell to begin with, and eval a code cell. I got 2 horizontal lines between input and output:
image

anglee:
And sometime when I delete the output cell, the bottom border is gone
image

anglee:
hmm... looks like the border issue is a Chrome thing(v 36.0.1985.125), I didn't see the same behavior in Firefox.

samccone:
yeah it is a chrome render bug  :(

anglee:
Note this one is not Chrome specific.
In advanced mode, looking at Ljubo's design (_Advanced mode.pdf_ on basecamp). There should be a horizontal line between the input cell and the output cell for example:
image
Right now it looks like:
image

anglee:
Right now, the view state of 'show hierarchy' and 'Advanced mode' is not kept when user reload the page.The view mode should be recorded in the **Session model**(not the notebook model) and used when restoring a session (e.g. reload page, or 'go to' from the start page). You might check out the use 'notebook edited' state as an example of states kept in session model: [sessionmanager.js#L207](https://github.com/twosigma/beaker-notebook/blob/master/core/src/main/web/app/mainapp/services/sessionmanager.js#L207).

samccone:
Do we want to keep this in the session?
I thought of it more as just a view mode and not something to persist

.gitignore
```diff
@@ -17,3 +17,6 @@ nbproject/
 build.xml
 manifest.mf
 .DS_Store
+/core/src/main/web/app/app.css
```
anglee, line 20: What do you think if we move the compiled results into a folder (`dist`?) and just gitignore the whole folder?

anglee:
Can we change the button style in the track settings box to be more consistent with other buttons? Also, maybe add some room btw it and the New Notebook button.
image
One can make the box show by going to start page, 'nav bar' -> 'settings' -> 'Set Anonymous tracking permission'

anglee:
About persisting view settings, I can see it as a application global user preference setting. Let's keep it as it is for now. I want to know what Ljubo and Mel's opinions are before doing anything.

core/src/main/java/com/twosigma/beaker/core/rest/PluginServiceLocatorRest.java
```diff
@@ -569,8 +569,8 @@ private String generateNginxConfig() throws IOException, InterruptedException {
                  Paths.get(htmlDir.toString() + "/login.html"));
       Files.copy(Paths.get(this.nginxStaticDir + "/present.html"),
                  Paths.get(htmlDir.toString() + "/present.html"));
-      Files.copy(Paths.get(this.nginxStaticDir + "/favicon.ico"),
-                 Paths.get(htmlDir.toString() + "/favicon.ico"));
+      Files.copy(Paths.get(this.nginxStaticDir + "/favicon.png"),
```
anglee, line 572: Since the file ext changed from 'ico' to 'png', do we need `<link rel="shortcut icon" href="../favicon.png" />` in `index.html`?

anglee:
I talked to Mel about view state persisting.
Ideally, the 'Advanced Mode' state should be persisted at the user pref level. And the 'Show Hierarchy' state should be persisted at the session level. But for this release (V1), they can both be User pref level or session level. I think this falls out of the scope of this PR. I created a separate issue(#437) to have this fixed in V1.

core/src/main/java/com/twosigma/beaker/core/rest/PluginServiceLocatorRest.java
samccone, line 572: :+1:

core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
samccone, line 383: :+1:

anglee:
After a code cell is run, the whitespace btw the Run button and the bottom border of the input cell increased unexpectedly.
pasted_image_8_7_14__10_30_am

.gitignore
samccone, line 20: OK i will change that right now.

samccone:
ok @anglee All of your comments have been addressed :tada:

anglee:
Changing the header level made all cells gone. Open a new notebook, click on the the cell menu of the section cell and try to change the header level by click on the menu item
image
Then the cells are gone:
image
Was able to reproduce this in both Chrome and FireFox. I noticed the notebook model seems to be updated correctly (so that after reloading the session, the cells appeared)

samccone:
Oh man that is no good :| looking into it

anglee:
hmmm. Just confirmed that the same issue also happens in master, I will also look into this.

samccone:
ah ok well looking into still.

anglee:
I am almost sure this is caused by my change https://github.com/twosigma/beaker-notebook/commit/9569e61b33fc411803f33db2dd63a8780e31ddf8. I will fix it. Sorry about the trouble.

anglee:
Hey Sam, the issue about changing section header level makes all cells go away was fixed. Please help review and maybe merge https://github.com/twosigma/beaker-notebook/pull/442 (to master as well as your branch). Thanks

anglee:
In 'Advanced mode', we were saying we want to display some indicator that tells which evaluator(language) is used. But I couldn't find it. Did I miss something?
image

samccone:
Yeah it was marked as something to add in the future

anglee:
Hmmm, ok. I don't remember that. I think this will make advanced mode not useful at all. I created https://github.com/twosigma/beaker-notebook/issues/443 and I think it should be given a very high priority.

samccone:
:tada: :tada:

scottdraves:
:dolphin:

samccone:
:ocean: :ocean: :ocean:

PR #453, opened 2014-08-12
extract templates

samccone:


anglee:
We want to follow the [Angular App Structure guide](https://docs.google.com/document/d/1XXMvReO8-Awi1EZXAXS4PzDzdNvV6pGcuaF4Q9821Es/pub). The idea being each directory is self-contained, reusable component. The template should be put next to the directive files. (Moreover, we should even try to separate the styles). If there are difficulties or reasons to not followed the agreed guide, please raise them and we can revisit the decision.

Also, there should be gradle tasks that starts beaker as well as start the watch task for developers so that on making changes to the templates, no manual steps are needed to regenerate the compile targets.

samccone:
> We want to follow the Angular App Structure guide. The idea being each directory is self-contained, reusable component. The template should be put next to the directive files. (Moreover, we should even try to separate the styles). If there are difficulties or reasons to not followed the agreed guide, please raise them and we can revisit the decision.

I can move the templates back into the directories if you would like. Having a single template just makes it simple from a watcher / compile / build / phase because I am only watching a single directory. It also is handy because the keys for each template are relative to their location in the template folder structure, which makes it quite nice to use on the front end, because the structure mirrors the component paths (just in the template directory) 

Centralizing all of the templates into a single compile pass also lets us minify, mutate, and compress all of the html into a very small payload for the user. Currently there is a http request for each template. On the notebook page for instance there are multiple http requests that have to occur just to  render the page. There is a notable speed boost that we gain by having all of the templates precompiled and packed into a single javascript file that we can call.

I have used this strategy on all of my JS apps over the past 2 years with much success. If the location of the file is what is throwing you off I have no problem moving them back. Just let me know.

Ah as a note for why we need to compile the templates, When we do the integration and make the notebook interface more pluggable across apps, we need to prefix all of the DOM nodes so that we can also prefix the CSS selectors, to prevent global CSS scope bleed. This compile task lets us do just that.

Again, let me know what your thoughts are.

anglee:
Thanks, Sam.

I agree with you that concating/compiling the template to reduce the count of http is nice. About the location of the html files, what about using `/**/*.html` insteadof `/template/**/*.html` just like you are already doing for the `scss` files? If there is a worry about we might include some html files that are not templates, I think we could affix `-template.html` and gulp watch the pattern `/**/*-template.html`. This way, we got the benefit of centralizing as well as having self-contained component code in directories following the structure guide.

samccone:
yep sounds good @anglee I will do that pattern.

I will ping you when I update this branch

PR #460, opened 2014-08-12
Fix spelling of file

samccone:


scottdraves:
need to fix it everywhere, eg inside that file it says "heirarchy-mode"

samccone:
k :+1:  thanks @scottdraves

PR #476, opened 2014-08-14
Add header menu spacing for markdown cells

### LINKED ISSUE #449: cell button menu aligned too low
there is whitespace above it, and it runs into stuff below it.  should be moved up.

screen shot 2014-08-12 at 11 19 31 am

samccone:
fixes #449

### DIFF
--- a/core/src/main/web/app/styles/cells.scss
+++ b/core/src/main/web/app/styles/cells.scss
@@ -77,6 +77,10 @@ bk-code-cell .code-cell-output {
     padding:    0 20px;
     position:   relative;
     border-top: solid 1px #CCC;
+
+    &.input-hidden {
+        margin-top: 35px;
+    }
 }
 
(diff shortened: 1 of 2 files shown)


anglee:
In the comment of #449, you were saying this is happening on a html (code) cell. The commit is about markdown. When I checkout the code and opened the tutorial notebook, I still see the same behavior to the first code cell (html cell with input cell hidden).

samccone:
@anglee did you just checkout the branch and not run build?

anglee:
I did run build. I guess what I was saying is I thought the problem is about the hidden input cells, but the class you changed looks like is for div in bk-markdown-cell.
pasted_image_8_14_14__4_07_pm

samccone:
ah ah you are right, shall get this one resolved :+1:  should be simple

samccone:
cool thanks for the :eyes:  @anglee  all updated

PR #483, opened 2014-08-15
Fix file open dialogs regressed in a4146e8

anglee:


samccone:
:+1: good catch I will add a test for this

Issue #469, opened 2014-08-14
Plugin Manager design needs to be simplified


scottdraves:
- Remove the "menus" panel (this can be set from the prefs file).
- Get rid of the text field next to "Add plugin from URL".
- When a user clicks one of the plugin buttons, it just loads.  ideally it would have a status light that cycled from gray to yellow to green.  even more ideally, if you click it again, and it is already loaded, then it unloads.
- The "Add plugin from URL" button becomes just another button in the big list (on the far right) with label "Open from URL...", which opens a dialog where you enter the URL and confirm.
- change the accordion list into tabbed panels.  plugins with no parameters do not create a panel.

anglee:
The "menus" panel let you let you load menu plugins dynamically. This cannot be done from the prefs file.

PR #485, opened 2014-08-18
Spot/469 (simplify plugin manager)

scottdraves:
Addresses Issue #469:
- Remove the "menus" panel (this can be set from the prefs file).
- Get rid of the text field next to "Add plugin from URL".
- When a user clicks one of the plugin buttons, it just loads.  It has a status light that cycles from gray to yellow to green.
- The "Add plugin from URL" button becomes just another button in the big list (on the far right) with label "Open from URL...", which opens a field where you enter the URL and confirm.
- change the accordion list into tabbed panels.  plugins with no parameters do not create a panel.

### DIFF
--- a/core/src/main/web/app/helpers/evaluatepluginmanager.js
+++ b/core/src/main/web/app/helpers/evaluatepluginmanager.js
@@ -18,7 +18,7 @@
  */
 (function() {
   'use strict';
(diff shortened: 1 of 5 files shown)


scottdraves:
thanks!

PR #500, opened 2014-08-25
Remove goto button and update tests

### LINKED ISSUE #496: Remove goto button on dashboard and make the entire notebook row clickable

samccone:
Fixes #496

### DIFF
--- a/core/src/main/web/app/controlpanel/table.jst.html
+++ b/core/src/main/web/app/controlpanel/table.jst.html
@@ -23,7 +23,6 @@
       </div>
     </div>
     <div class='btn-group right'>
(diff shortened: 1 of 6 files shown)


samccone:
Looks like we are getting a timeout waiting for the python eval to load

anglee:
Looks like this change made only the notebook title (Text) part clickable, but not the entire row as #496 describes. @samccone or @melmac48, please confirm what the expected behavior should be.

samccone:
I thought this is what we wanted, but i guess the final call is @melmac48

melmac48:
If only the notebook text and button are clickable, then this is good for now, thanks.

anglee:
The goto button has been removed

melmac48:
Okay, that's fine.

PR #501, opened 2014-08-25
Update plugin manager style

### LINKED ISSUE #495: Update plugin manager styling per new UI

samccone:
fixes #495

### DIFF
--- a/core/src/main/web/app/mainapp/components/pluginmanager/pluginmanager.jst.html
+++ b/core/src/main/web/app/mainapp/components/pluginmanager/pluginmanager.jst.html
@@ -19,13 +19,15 @@
   </div>
   <div class="modal-body modal-large">
     <div>
-      <button ng-repeat="(pluginName, pluginStatus) in evalTabOp.getKnownEvaluatePlugins()"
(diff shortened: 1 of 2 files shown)


scottdraves:
missing the button around "from url..."

scottdraves:
when the buttons are clicked you get a blue outline that is partially obscured.  please remove it.

scottdraves:
buttons should not be clickable if they are already loaded or loading.

samccone:
ok i updated the style issues, per 

> buttons should not be clickable if they are already loaded or loading.

Want to open a new issue and I can resolve that since that is new functionality?

scottdraves:
thanks.  OK yes I will handle the reclick issue through another issue assigned to me.

For the "From URL" button, the bullet should be invisible but the same size as the other buttons.
That will should the 1-pixel gap above it.

screen shot 2014-08-25 at 1 19 32 pm

samccone:
ah good spot, was unsure about how we wanted it to really look. updating now

samccone:
all updated @scottdraves

Issue #353, opened 2014-06-23
if server restarts then reset local state


scottdraves:
We get problems like Issue #351 when the server is restarted and rest calls from JS with state from one server are made to another server.

When we detect the server has closed we should deactivate the page and prevent interaction (but probably leave the content up to give the user a chance to rescue it).

PR #504, opened 2014-08-25
Fix #486, handle connection status properly when recover from sleep

### LINKED ISSUE #486: putting laptop to sleep triggers disconnected dialog
Say I am running beaker and have a notebook open.
If I close my laptop, wait a bit, and then open it up again.
Then i try to edit a cell.
I get the "Disconnected" dialog, even though it is still actually connected (I can even type and evaluate cells even though the dialog is up).

anglee:
In this change, when beaker get a disconnect change (from cometd), it
will wait for 5 seconds. If within 5 seconds, there is a connect event,
it will try to ditinguish the connect event coming from a valid
reconnect(same session is still availble) from an invalid connect that
can for example come from server crash or user stopping the server
intentionally. If there is no connect event, again within 5 sec of
the disconnect event, and there is a key press event, as well as timeout
after 5 secs, beaker will treat the situation as server is permanently
gone and stop further reconnect attempt and pops up the dialog for
saving the notebook model as a client file (#353)

### DIFF
--- a/core/src/main/web/app/mainapp/mainapp.js
+++ b/core/src/main/web/app/mainapp/mainapp.js
@@ -603,7 +603,7 @@
               $timeout.cancel(reconnectTimeout);
               reconnectTimeout = undefined;
             }
-            window.removeEventListener('keypress', indicateReconnectFailed);
+            window.removeEventListener('keypress', indicateReconnectFailed, true);
           };
 
           return {
--- a/core/src/main/web/app/mainapp/mainapp.js
+++ b/core/src/main/web/app/mainapp/mainapp.js
@@ -559,16 +559,16 @@
                 "Save current notebook as a file?",
                 "Disconnected",
                 function() {
-                  // "Not now", hijack all keypress events to prompt again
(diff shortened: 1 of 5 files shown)


scottdraves:
thanks that sounds great i will look at it shortly.

scottdraves:
i am still getting the disconnection box, but it's just delayed until i type something.
stop by my desk for a demo.

scottdraves:
in fact i never see the "reconnecting" message.

scottdraves:
reproduced this at home on my external laptop, but i did see "reconnecting" flash for a second, then the dialog came up.  i dismissed and used a cell and notebook worked fine.

scottdraves:
still doesn't work :(

scottdraves:
also, i noticed if you click away the dialog (eg on the darkened background) this is interpreted as clicking "OK" and it downloads the notebook.  should mean cancel instead.

anglee:
it should be fixed now.

scottdraves:
thanks, worked fine

Issue #443, opened 2014-08-11
Show language plugin indicator on code cells in 'Advanced mode' 


anglee:
In Advanced mode, right now there is no easy way to tell which language was used for a particular code cell.

samccone:
@melmac48 @anglee @scottdraves is there a priority label for this?

scottdraves:
related: https://github.com/twosigma/beaker-notebook/issues/599
https://github.com/twosigma/beaker-notebook/issues/601

PR #516, opened 2014-08-29
Add language icons

### LINKED ISSUE #554: Distinguishing languages a bit more in regular mode
I’d consider distinguishing languages a bit more in regular mode… maybe add an icon or use a different font color for each language

anglee:
This is part of work to address #443. We still need to make the language icons appear in advanced mode.

fixes #554

### DIFF
--- a/core/src/main/web/app/app.scss
+++ b/core/src/main/web/app/app.scss
@@ -300,3 +300,17 @@ table.dataTable tbody td {
 .text-danger {
     color: #a94442;
 }
+
+bk-language-logo {
+  span {
+    width: 20px;
+    height: 19px;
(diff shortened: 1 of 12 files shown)


scottdraves:
cool.  Julia should be "jl" not "ju".  and "Ht" should be "HT" and ruby should be "rb".
also, needs space between the icon and the regular name.
actually now that i think about it, this should probably be in the icon bar on the right, not on the left.
and the vertical alignment is off between the text of the icon and the full name.
how hard would it be to add the icons to the "new code cell" menu too?=

anglee:
The letters used for short names have been updated. Please merge and @samccone please take over. Thanks guys.

Issue #514, opened 2014-08-29
update default notebook


scottdraves:
"will turn light blue when plugin loads successfully"

not anymore

how about

```
# type some python on the next line and click "Run" to evaluate it and display the result.  for example, try: 1+1
```

scottdraves:
https://github.com/twosigma/beaker-notebook/pull/527

anglee:
Thanks for pointing out the default notebook is outdated. However, we should review this as part of the default notebook plan.

It has been determined that the current default notebook serves 2 purposes now:
1) as a brief intro of how to use notebook for new users
2) as a empty point that is good for starting a research notebook.
And it was decided by @melmac48 that we will have 2 notebooks for the 2 purposes for V1.

scottdraves:
I don't see any Issue or email with this new plan and since we are in the final stages of code freeze for 1.0 we should make the smaller change which is this PR #527

PR #527, opened 2014-08-31
Fix Issue #514: improve default notebook by adding some friendly comments, making the co...

scottdraves:
...de two lines long so it is easier to click on and add to, removing the JS cell because our users are not really after JS, just add that one to the list of languages below

### DIFF
--- a/core/config/default.bkr
+++ b/core/config/default.bkr
@@ -4,31 +4,52 @@
         {
             "name": "Html",
             "plugin": "Html",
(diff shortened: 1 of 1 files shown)


anglee:
I left a comment under #514

anglee:
closing #527 and #514, opened #533. Next time a plan needs to be reviewed first and the issue has to be triaged by the team before getting worked on.

Issue #529, opened 2014-08-31
new Stdout/err panel missing resize handle


scottdraves:
The old output panel had a border that you could grab and resize it larger or smaller.
This is an important feature missing from the new one and we need to bring it back.

scottdraves:
The new panel is also missing the box around the text, so you can't tell if you print an empty line.
this makes scrolling the text especially confusing.

scottdraves:
https://github.com/twosigma/beaker-notebook/pull/530

scottdraves:
here's where this was removed, the code is from there: https://github.com/twosigma/beaker-notebook/pull/433/files

PR #530, opened 2014-08-31
Fix Issue #529 by restoring handle on top of stdout panel

scottdraves:


anglee:
Please use spaces instead of tabs for whitespace indentation.

anglee:
The color of the strip looks strange in the new UI.

Issue #531, opened 2014-08-31
open bkr notebook menu item broken


scottdraves:
TypeError: undefined is not a function
    at Object.toAdd.items.action (http://127.0.0.1:8801/beaker/plugin/menu/controlpanel.js:54:38)
    at Scope.$scope.runAction (http://127.0.0.1:8801/beaker/app/utils/basic/commonui.js:187:20)
    at http://127.0.0.1:8801/beaker/vendor/bower_components/angular/angular.js:10168:21
    at http://127.0.0.1:8801/beaker/vendor/bower_components/angular/angular.js:17823:17
    at Scope.$eval (http://127.0.0.1:8801/beaker/vendor/bower_components/angular/angular.js:11906:28)
    at Scope.$apply (http://127.0.0.1:8801/beaker/vendor/bower_components/angular/angular.js:12006:23)
    at HTMLAnchorElement.<anonymous> (http://127.0.0.1:8801/beaker/vendor/bower_components/angular/angular.js:17822:21)
    at HTMLAnchorElement.x.event.dispatch (http://127.0.0.1:8801/beaker/vendor/bower_components/jquery/jquery.min.js:5:10006)
    at HTMLAnchorElement.$event.dispatch (http://127.0.0.1:8801/beaker/vendor/jquery.event.drag/jquery.event.drag.js:374:19)
    at HTMLAnchorElement.y.handle (http://127.0.0.1:8801/beaker/vendor/bower_components/jquery/jquery.min.js:5:6796)

PR #532, opened 2014-08-31
Fix Issue #531, by fixing path to template

scottdraves:


anglee:
The change is good. Thanks.

A side note to @samccone, the plugins' templates shouldn't be bundled with the templates for core.

Issue #44, opened 2014-02-08
R plugin missing interrupt


scottdraves:
you can't stop an infinite loop

scottdraves:
fixed by https://github.com/twosigma/beaker-notebook/pull/541

scottdraves:
and on windows https://github.com/twosigma/beaker-notebook/issues/542

PR #541, opened 2014-09-03
Fix Issue #44 Make R evaluation interruptible

scottdraves:
This is not easy and it does not always work because not all R operations can be interrupted.
But presumably R users are used to this and it's the best we can do.
The way it works is, when R starts a new session we ask it for its unix PID.
Then, we interrupt it by sending a SIGINT.
The weird part is someone (Rserve?) is printing an extra empty line which must be suppressed lest our output tray pop up.

This will not work on windows, will need more work there.

### DIFF
--- a/plugin/r/src/dist/r.js
+++ b/plugin/r/src/dist/r.js
@@ -160,6 +160,16 @@ define(function(require, exports, module) {
         data: { shellID: self.settings.shellID }
       });
     },
+    cancelExecution: function() {
+      // XXX duplicate of interrupt.
+      var self = this;
+      $.ajax({
+        type: "POST",
+        datatype: "json",
+        url: serviceBase + "/rest/rsh/interrupt",
+        data: { shellID: self.settings.shellID }
+      });
+    },
     spec: {
       interrupt: {type: "action", action: "interrupt", name: "Interrupt"}
     },
--- a/plugin/r/src/dist/r.js
+++ b/plugin/r/src/dist/r.js
@@ -151,7 +151,17 @@ define(function(require, exports, module) {
         data: { shellID: self.settings.shellID }
       }).done(cb);
     },
+    interrupt: function() {
+      var self = this;
+      $.ajax({
+        type: "POST",
+        datatype: "json",
+        url: serviceBase + "/rest/rsh/interrupt",
+        data: { shellID: self.settings.shellID }
+      });
+    },
     spec: {
+      interrupt: {type: "action", action: "interrupt", name: "Interrupt"}
     },
     cometdUtil: cometdUtil
   };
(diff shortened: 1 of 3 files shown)


plugin/r/src/dist/r.js
```diff
@@ -151,7 +151,27 @@ define(function(require, exports, module) {
         data: { shellID: self.settings.shellID }
       }).done(cb);
     },
+    interrupt: function() {
+      var self = this;
+      $.ajax({
```
anglee: It would be better to use `bkHelper.httpPost` so the AngularJS knows to $digest and do any updates needed. See the `newShell` method for an example.

plugin/r/src/dist/r.js
```diff
@@ -151,7 +151,27 @@ define(function(require, exports, module) {
...
+        type: "POST",
+        datatype: "json",
+        url: serviceBase + "/rest/rsh/interrupt",
+        data: { shellID: self.settings.shellID }
+      });
+    },
+    cancelExecution: function() {
+      // XXX duplicate of interrupt.
+      var self = this;
```
anglee: If it is simply duplicating, why not have one call another?

plugin/r/src/main/java/com/twosigma/beaker/r/module/ErrorGobbler.java
```diff
@@ -38,6 +47,10 @@ public void run() {
       BufferedReader br = new BufferedReader(isr);
       String line = null;
       while ((line = br.readLine()) != null) {
+        if (expectingExtraLine && line.length() == 0) {
```
anglee, line 50: maybe change the `line.length() == 0` part to  `line.isEmpty()` or `"".equals(line)`?

plugin/r/src/main/java/com/twosigma/beaker/r/rest/RShellRest.java
```diff
@@ -393,13 +414,18 @@ private static boolean isDataFrame(REXP result, SimpleEvaluationObject obj) {
   private static class RServer {
     RConnection connection;
     ROutputHandler outputHandler;
+    ErrorGobbler errorGobbler;
```
anglee, line 417: Can this be `final`?

plugin/r/src/main/java/com/twosigma/beaker/r/module/ErrorGobbler.java
scottdraves, line 50: no

plugin/r/src/main/java/com/twosigma/beaker/r/rest/RShellRest.java
scottdraves, line 417: no

PR #571, opened 2014-09-05
Prevent modal screen overflow

### LINKED ISSUE #288: file save dialog can go off the bottom of the window
If you open a folder with a lot of subfolders or if you just make your window short, the bottom of the dialog, including the Open/Canel buttons goes off the bottom of the window and becomes inaccessible.

samccone:
Fixes #288

### DIFF
--- a/core/src/main/web/app/app.scss
+++ b/core/src/main/web/app/app.scss
@@ -26,6 +26,7 @@
 @import "styles/animations";
 @import "styles/hierarchy";
 @import "styles/stdout";
+@import "styles/modal";
 
 root {
     display: block;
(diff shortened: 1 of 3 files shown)


anglee:
the 2 vertical scroll bars seems not desirable. We should always show the header and footer
image

samccone:
hmm ok I will investigate a better fix

samccone:
ok @anglee @scottdraves resolved it with the other solution :)

samccone:
This seems to be working fine for me

samccone:
esh we should revert this for now... it actually does not cover all the cases and causes bugs as @scottdraves pointed out

PR #594, opened 2014-09-09
Handle moving markdown editor up and down

### LINKED ISSUE #350: Markdown cell contents are cleared when they are moved "up" in cell order.
This problem does not happen when you move the cells down however.

samccone:
fixes #350

Due to the way that epic editor works and the way that moving elements
via the cellmanger you need to reinit the markdown editor to reshow the
markdown content.

### DIFF
--- a/core/src/main/web/app/mainapp/components/notebook/markdowncell-directive.js
+++ b/core/src/main/web/app/mainapp/components/notebook/markdowncell-directive.js
@@ -19,6 +19,55 @@
   var module = angular.module('bk.notebook');
 
   module.directive('bkMarkdownCell', ['bkSessionManager', 'bkHelper', function(bkSessionManager, bkHelper) {
+    var editor = null;
+
+    function initializeEditor(scope, element, attrs) {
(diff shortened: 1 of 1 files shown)


samccone:
FYI this PR exacerbates the newline whitespace bug everytime you move up and down

scottdraves:
that sucks :(
what determines when the cell gets a blue outline idk if i've seen that before

anglee:
After this change, an error appears in the console when a markdown cell is inserted:
image

anglee:
I noticed `bkHelper` was injected. bkHelper should never be injected. It should be limited to used by plugins only. core code should use specific services, In this case, `bkSessionManager.isNotebookLocked` should be used.

samccone:
looks like i added this one
https://github.com/twosigma/beaker-notebook/commit/f36f4b57d7e527e7a1ca98e1a3e8103cdcd48c95

just curious as to the difference

anglee:
Just to be clear, using bkHelper has nothing to do with the console error. I just happened to notice both at the same time.

PR #604, opened 2014-09-10
Harden failing specs

samccone:


samccone:
welp it looks like this fixes two of the issues

scottdraves:
ang pointed out: this should also remove use of JS cell in the default notebook which was also removed.

samccone:
:+1: on it

samccone:
oh it passes

samccone:
:dancer: :dancer: :dancer: :dancer: :dancer: :dancer: :dancer: :dancer: :dancer:

Issue #618, opened 2014-09-11
cannot select text in cell output that's an error


scottdraves:
the first line you can highlight but it does not stay selected so that i can copy it.
the expanded msg cannot be highlighted at all.

scottdraves:
this effects all cells moving to 1.0 p1

anglee:
This was introduced by https://github.com/twosigma/beaker-notebook/pull/610

scottdraves:
sorry!

PR #625, opened 2014-09-11
p2 fixes

samccone:


scottdraves:
please do not show empty parens

samccone:
:+1:

anglee:
I can still reproduce #618 with the `z-index` change

anglee:
Now in advanced mode, the output menu hamburger icon doesn't response to mouse clicks

samccone:
I have dropped that commit for now, so we can get this in

PR #632, opened 2014-09-12
Fix #618, do not handle click in output area

### LINKED ISSUE #618: cannot select text in cell output that's an error

anglee:


samccone:
:+1:

scottdraves:
this breaks clicking on cell backgrounds.
we need to fix that fix, not undo it!

anglee:
done

scottdraves:
thanks

PR #642, opened 2014-09-13
Fix markdown go blank issue

anglee:
@samccone, I made some change on top of your change that seems to have address the scope/editor mismatching causing editor go blank or wrong focus issue. I am still seeing the whitespace growing though. Please review and see if it makes sense to merge into your branch.

### DIFF
--- a/core/src/main/web/app/mainapp/components/notebook/markdowncell-directive.js
+++ b/core/src/main/web/app/mainapp/components/notebook/markdowncell-directive.js
@@ -19,7 +19,6 @@
   var module = angular.module('bk.notebook');
 
   module.directive('bkMarkdownCell', ['bkSessionManager', 'bkHelper', '$timeout', function(bkSessionManager, bkHelper, $timeout) {
-    var editor = null;
 
     function initializeEditor(scope, element, attrs) {
       var div = element.find("div").first().get()[0];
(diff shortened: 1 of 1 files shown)


core/src/main/web/app/mainapp/components/notebook/markdowncell-directive.js
```diff
@@ -37,35 +36,47 @@
           scroll: true
         }
       };
+      if (scope.editor) {
+        scope.editor.removeListener("preview");
```
samccone, line 40: ahhhh this was it, we were not unbinding 
good catch @anglee 

I was under the impression that since we were initing a new editor we would be ok, but alas I guess not!

great catch

core/src/main/web/app/mainapp/components/notebook/markdowncell-directive.js
```diff
@@ -19,7 +19,6 @@
   var module = angular.module('bk.notebook');
 
   module.directive('bkMarkdownCell', ['bkSessionManager', 'bkHelper', '$timeout', function(bkSessionManager, bkHelper, $timeout) {
-    var editor = null;
```
anglee, line 22: I think this is another thing that worth pointing out. The directive function is called just once per app (`link` and `controller` are per for each directive instance however), making editor a variable local like this will make all the directive instances closure bind to the same editor variable. There should be an editor recorded for each scope instead.

samccone, line 22: you are right, this is not in the link, so this will cause issues.

silly me

PR #648, opened 2014-09-14
Add language logos to plugin manager 

scottdraves:
why does the logo appear outside the button?
there is something very weird happening with the styling that i don't understand.
if we can get this then cool otherwise no big deal this is cosmetic only.
screen shot 2014-09-13 at 9 11 43 pm


scottdraves:
thanks! so close!

samccone:
well the lack of logos per unloaded languages is due to the issue of not knowing the "style" spec until the evaluator is loaded.

scottdraves:
well we are only talking about a handful of values that are not very mutable.
i say just put a table mapping the labels to the logo colors and short names right here where we need it.
but i don't have time i am going to sleep :)

PR #654, opened 2014-09-15
Remove cell border when locked

samccone:


scottdraves:
@samccone sorry this got lost in the rush to 1.0.
is it still ready for merge?
will need to be reproduced on sharing server too.

samccone:
Yeah it looks like it can be merged cleanly, would want to test before tho

samccone:
:ship:

Issue #144, opened 2014-04-16
When loading a plugin from a cross-origin URL, show a dialog warning the user that it could be unsafe


jchendy:
Currently, plugins are automatically loaded. If it's coming from a cross-origin URL, this is a potential security issue. A minimal step toward address this is to present a dialog asking the user to confirm the loading of cross-origin URLs.

This is part of #142.

PR #718, opened 2014-10-02
Fix for 144: When loading a plugin from a cross-origin URL, show a dialog warning the user that it could be unsafe

ildipo:
Fix for bug [144](https://github.com/twosigma/beaker-notebook/issues/144).
In the plugin manager if the user inserts an URL as a source of the plugin, a dialog is shown to ask for confirmation.
At notebook load, if any evaluator is referencing an external URL, a dialog is shown to ask for confirmation. If the user denies the notebook is loaded without the URL-based evaluators.

### DIFF
--- a/core/src/main/web/app/mainapp/components/pluginmanager/pluginmanager-directive.js
+++ b/core/src/main/web/app/mainapp/components/pluginmanager/pluginmanager-directive.js
@@ -107,16 +107,15 @@
               bkCoreManager.getBkApp().addEvaluator(newEvaluatorObj);
             } if(fromUrl) {
                 var r = new RegExp('^(?:[a-z]+:)?//', 'i');
-                if (!r.test(plugin) || $scope.evalTabOp.forceLoad)
-                {
-                    var newEvaluatorObj = {
-                      name: "",
-                      plugin: plugin
-                    };
-                    $scope.evalTabOp.forceLoad = false;
-                    $scope.evalTabOp.newPluginNameOrUrl = "";
-                    bkSessionManager.addEvaluator(newEvaluatorObj);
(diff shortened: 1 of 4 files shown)


core/src/main/web/app/mainapp/components/pluginmanager/pluginmanager-directive.js
```diff
@@ -91,13 +98,28 @@
...
+            if (status == "known") {
               var newEvaluatorObj = {
                 name: "",
                 plugin: plugin
               };
               bkSessionManager.addEvaluator(newEvaluatorObj);
               bkCoreManager.getBkApp().addEvaluator(newEvaluatorObj);
+            } if(fromUrl) {
+                var r = new RegExp('^(?:[a-z]+:)?//', 'i');
+                if (!r.test(plugin) || $scope.evalTabOp.forceLoad)
```
scottdraves: brace on same line as condition

core/src/main/web/app/mainapp/mainapp.js
```diff
@@ -121,6 +121,54 @@
...
+            var deferred = bkUtils.newDeferred();
+            bkCoreManager.show2ButtonModal(
+                "WARNING: this notebook is asking to load the following plugins from external servers:<br/>" + urlList+ " <br/>",
+                "Do you want to continue?",
+                function() {
+                  deferred.reject();
+                },
+                function() {
+                  deferred.resolve();
+                }, "No, Cancel!", "Continue", "", "btn-danger");
```
scottdraves: if the user clicks cancel the notebooks does load, so cancel/continue isn't the right choice.
something like "clear external plugins / load as is" is more accurate.

Issue #715, opened 2014-10-02
add PP and ToU links and improve wording of About box license statement.


scottdraves:
![screen shot 2014-10-01 at 10 02 05 pm](https://cloud.githubusercontent.com/assets/963093/4486005/9351fdce-49d8-11e4-8961-effd961dc76f.png)

"the source code is available under" to "the software is available under"

PR #722, opened 2014-10-03
Fix Issue #715 by adding more links to the about box and tweaking the te...

scottdraves:
...xt.  Also, make all the links open in a new tab.


ildipo:
The test build failed but for some other reasons. Patch seems ok.

Issue #320, opened 2014-06-09
static files are cached


scottdraves:
This causes problems for users when they get an upgrade and have to clear their browser cache to prevent getting a possibly-inconsistent state.

Developers also sometimes forget to clear their cache, and it would be great o remove this step from our process.

scottdraves:
for unknown reasons, we still have cache problems.

we need to take more extreme measures.

melmac48:
@scottdraves we should add a note to the install page for users who've previously used the beta to refresh their cache.

PR #724, opened 2014-10-03
Fix for issue 320

ildipo:
This addresses the issue #320

Added git hash to each url ‘internally’ used by beaker.
Changed all urls to be relative. Used base tag in a auto-generated
index.html.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,6 @@ manifest.mf
 /core/src/main/web/app/app.css
 npm-debug.log
 core/src/main/web/app/templates.js
+core/config/hash
+
+core/src/main/web/index.html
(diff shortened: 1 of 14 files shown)


scottdraves:
Great.  Did you test this with the --public-server option, ie with the Docker container?

scottdraves:
when this is merged we should create a new issue to turn caching back, ie stop manipulating the http headers to try to disable it.

ildipo:
I did not test with docker. I'll do that ASAP.

ildipo:
Ok I tested it and I found a small issue - I fixed the rules for nginx one again and now works in both cases.

Issue #707, opened 2014-09-30
Set size of R graphics


janovergoor:
I'd like to be able to set the size of R output graphics. 

For example, I would usually shape graphics by writing to a pdf file and setting height=5, width=10.
Since the beaker screen is generally wider than taller, such aspect ratio's would be useful to set.

Right now, the only solution I can think of is using theme(aspect.ratio=0.5) from ggplot, but that creates a a bad-looking sea of whitespace.

scottdraves:
yea this has been requested several times.

scottdraves:
demo of the API (from the R tutorial):
http://sharing.beakernotebook.com/gist/anonymous/8703affdf1e53e3ac73b

remaining issue:
https://github.com/twosigma/beaker-notebook/issues/731

PR #725, opened 2014-10-03
Fix Issue #707 Set size of R graphics by defining a new function in the ...

scottdraves:
...beaker module where you can give all the options to the SVG renderer (including width and height).  This introduces a complication which is you cannot access these options until they are defined, and our normal evaluation path does SVG setup, so I had to add an option to the evaluate REST call for "init" evaluations that do not do all the wrapping and output capture that normal user evaluation does


scottdraves:
note that after this is merged we have to update the R tutorial to document it.
also the solution is not the best I think there are some R tricks that can improve it, but this is enough for 1.1

scottdraves:
https://github.com/twosigma/beaker-notebook/issues/731

Issue #695, opened 2014-09-26
python plugin fails with python3


scottdraves:
urllib2 (used by autotranslate) does not exist in python3.

HanniAli:
I would note that using 2to3 to migrate the beakey.py works fine to upgrade the code and I can successfully use most of the Ipython functionality against my Python33 installation. However it I try to use the beaker.set method I run into encoding problems even with a simple string.

Full python3 support out of the box would be appreciated.

scottdraves:
Thanks Hanniali, that helps.  This issue is next on my agenda.

scottdraves:
https://github.com/twosigma/beaker-notebook/pull/733

PR #733, opened 2014-10-05
Fix Issue #695 support python3 

scottdraves:
by making a python3 version of beaker.py and then dynamically choosing to load it or the old one, as needed.


ildipo:
tested with python 2 and 3

HanniAli:
Thanks guys, quick check though, when I migrated beaker.py I had issues running through this notebook: http://sharing.beakernotebook.com/gist/anonymous/5fcd0e4ec4e514429bb7 beaker.set would throw errors, have you addressed the encoding differences between 2 and 3 as well?

HanniAli:
Ignore me, I just pasted in the new beaker.py code to my current install and it works fine. All good.

Thanks!

PR #739, opened 2014-10-07
added helper function for sharing d3 output

ildipo:
This adds a new function to bkHelper - updateDocumentModelFromDOM -
that can be used by JavaScript plugin to update the display of an HTML
cell in order to share it.


scottdraves:
Also, please update the d3 tutorial to use this function.

PR #752, opened 2014-10-13
start the evaluators (aka plugins) asynchronously after the notebook doc...

scottdraves:
...ument model has been loaded and the session has started.  this fixes a problem with big notebooks freezing too long and crashing the load.


scottdraves:
this approach doesn't work

Issue #413, opened 2014-07-28
Save Beaker plot client state


anglee:
For example when user zoom/pan/add-tooltips we want the client states generated by interaction be kept in sessions/shared-notebooks/saved-notebook-files

scottdraves:
Yes.  The output displays should have two new methods.  One to package up their state into JSON, another to restore their state to given JSON.

PR #768, opened 2014-10-16
Ildipo/displaystate

ildipo:
addresses issue #413
implements output state saving for plot, combined plots and tables

### DIFF
--- a/core/src/main/web/app/mainapp/mainapp.js
+++ b/core/src/main/web/app/mainapp/mainapp.js
@@ -441,12 +441,6 @@
             return deferred.promise;
           };
 
-          var saveDoNotOverwrite = function(uri, uriType) {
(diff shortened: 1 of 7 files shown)


core/src/main/web/app/mainapp/mainapp.js
```diff
@@ -464,8 +458,10 @@
           var saveAlwaysOverwrite = function(uri, uriType) {
             var deferred = bkUtils.newDeferred();
             var fileSaver = bkCoreManager.getFileSaver(uriType);
-            var content = bkSessionManager.getSaveData().notebookModelAsString;
-            fileSaver.save(uri, content, true).then(function() {
+            bkSessionManager.dumpDisplayStatus();
+            $timeout(function() {
```
scottdraves, line 462: please add a comment about why the timeout is needed.

PR #770, opened 2014-10-16
use command as well as control

ildipo:
mac users will be happy


scottdraves:
need to add command-z and s.

need to update the keyboard help file.

PR #787, opened 2014-10-21
fixed python json serialization issues with DataFrame

ildipo:


scottdraves:
This PR also locks down version of Angular Datatables to protect against bug in their recent release.

PR #795, opened 2014-10-22
Kill menu from bunsen beaker

buob:
@samccone


samccone:
Ok all this needs is those few comments and we are good to goooo

Issue #769, opened 2014-10-16
Configurable gist and nbviewer URLs


rseymour:
It would be awesome to have the following two URLs (lines 59 and 74) configurable.

https://github.com/twosigma/beaker-notebook/blob/master/core/src/main/java/com/twosigma/beaker/core/rest/PublishRest.java#L59

https://github.com/twosigma/beaker-notebook/blob/master/core/src/main/java/com/twosigma/beaker/core/rest/PublishRest.java#L74

It was easy to put in our hosted github enterprise, local nbviewer and rebuild but it would be nice to have that as a config option as well.  I don't know how to do it offhand since my java skills drop sharply after hello world.  Are you using a certain config library that allows for easy access to the config.json key/values inside your compiled code?

Thanks (and thanks for electricsheep!)

scottdraves:
You are very welcome Rich.
Yes those should definitely be config options.
Should be an easy to do, we'll see if we can fit it into the next release.
Thanks.

rseymour:
Thanks!

rseymour:
There is one issue with this, sorry I didn't get a chance to test it.  The problem that I can't get around is that the github enterprise puts the /raw at the end of the URL (meaning after the gist id).  I'm thinking of a way to insert it after, but I wonder what you all think.  I had just hardcoded it.

scottdraves:
I think I know what you mean but to actually implement a fix, it would help if I could see your code.

rseymour:
I will give you the full URL string when I'm back at work tomorrow.  I went into the enterprise api side, but it still looked like the API 'raw' link appended /raw instead of prepending it like github.com does.  Full bug report forthcoming.  And since you pulled the JSON config parsing into Java I may be able to send you a proper pull request to save you some work.  Thanks.

rseymour:
My annoyance just fixed itself with our latest github enterprise upgrade.  Muchos gracias for the configurability.

PR #808, opened 2014-10-27
added GIST and sharing server config env var

ildipo:
fix for issue #769


scottdraves:
can we get this from the ~/.beaker/v1/config/beaker.pref.json file?

i know that's more work but it's long overdue that we parse that in java.  sorry i should have clarified that in advance.

PR #843, opened 2014-11-04
Remove option to select plugins bunsen doesn't support.

mforsyth:


core/src/main/web/plugin/init/addevalplugins.js
```diff
@@ -19,10 +19,10 @@
     return {
       "IPython": "./plugins/eval/ipythonPlugins/ipython/ipython.js",
       // "IRuby": "./plugins/eval/ipythonPlugins/iruby/iruby.js",
-      "Julia": "./plugins/eval/ipythonPlugins/julia/julia.js",
-      "Groovy": "./plugins/eval/groovy/groovy.js",
-      "R": "./plugins/eval/r/r.js",
+      // "Julia": "./plugins/eval/ipythonPlugins/julia/julia.js",
```
samccone, line 22: lets just :fire: these right?

mforsyth, line 22: There's another one (Iruby) that was left commented out before this commit... I don't know...I think in this case it's kind of handy to have a reminder of which plugins actually exist in the codebase (and could be turned on by just installing some more apt packages).

samccone, line 22: hmm ok, it is a pattern in the repo to comment out nonfunctional code to leave as a reminder so ok

PR #846, opened 2014-11-06
implemented new test system

ildipo:


scottdraves:
needs a new test/README.md with instructions

scottdraves:
instead of handling a missing hash by generating a random one, how about just doing the build correctly?

PR #873, opened 2014-11-15
Should work now even if homebrew is installed in non-standard location

mds2:
I am the sort of [insert negative word here] who installs his homebrew into a non-standard location.  The gradle files for beaker notebook assume that everything installed via homebrew is in the normal /usr/local location.  This patch is an attempt to fix this.  It works on my machine.

Caveat : I have never written any Groovy or Gradle before.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -70,7 +70,8 @@ task (compileCss, type: Exec) {
 
 task (copyNginx, type: Exec) {
   if (mac()) {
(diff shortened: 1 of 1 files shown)


scottdraves:
thanks mike, i like it.
if we can get a CLA from you (http://beakernotebook.com/community) i can take a PR.
otherwise i can rewrite it, but it may take longer.

scottdraves:
thank you!

PR #882, opened 2014-11-17
Ildipo/autocomplete

ildipo:


scottdraves:
need to either document how to call the parser-generator in a README, or (better) do it automatically at build time from gradle.

PR #927, opened 2014-11-25
sessions:  Allow session ID's to be controlled externally.

mforsyth:
Introduces editById as a main entrypoint (parallel to
"session" or "open"), which, given a notebook location and ID,
opens a new session to edit the notebook, or if there's already an
open session with that ID, re-uses the existing one.

This makes it so that a user can never have more than one session
open for the same notebook, which is what we want for Bunsen.

### DIFF
--- a/core/src/main/web/app/beaker.js
+++ b/core/src/main/web/app/beaker.js
@@ -119,6 +119,9 @@
           return '/session/' + newSessionId;
         };
       };
+      var makeSessionUrl = function(sessionId) {
+        return '/session/' + sessionId + "?" + window.location.href.split("?")[1];
+      };
       $routeProvider
           .when('/session/new', {
             redirectTo: makeNewProvider("new")
@@ -139,7 +142,18 @@
               sessionRouteResolve.target = function() {
                 return search;
               };
-              return '/session/' + newSessionId + "?" + window.location.href.split("?")[1];
+	      return makeSessionUrl(newSessionId);
+            }
+          })
+          .when('/edit/:notebookId', {
+            redirectTo: function(routeParams, path, search) {
+              sessionRouteResolve.editingById = function() {
+                return routeParams.notebookId;
+              };
+              sessionRouteResolve.target = function() {
+                return search;
+              };
+              return makeSessionUrl(routeParams.notebookId);
             }
           })
           .when('/control', {
--- a/core/src/main/web/app/mainapp/mainapp.js
+++ b/core/src/main/web/app/mainapp/mainapp.js
@@ -148,6 +148,18 @@
             }
             $scope.loading = false;
           };
+
+	  function loadFromRetrievedSession(session, id) {
+            var notebookUri = session.notebookUri;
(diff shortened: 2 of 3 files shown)


mforsyth:
attn @grin @samccone

core/src/main/web/app/beaker.js
```diff
@@ -142,6 +142,17 @@
...
           })
+          .when('/editById', {
+            redirectTo: function(routeParams, path, search) {
+              sessionRouteResolve.editingById = function() {
+                return true;
+              };
+              sessionRouteResolve.target = function() {
+                return search;
+              };
+              return '/session/' + search.notebookId + "?" + window.location.href.split("?")[1];
```
samccone: we for sure should use the url module and url.parse to do this work

mforsyth: this is just copied from the way that Beaker loads from url now... maybe I should make them call a common function to make that more obvious

core/src/main/web/app/beaker.js
```diff
@@ -119,6 +119,9 @@
           return '/session/' + newSessionId;
         };
       };
+      var makeSessionUrl = function(sessionId) {
+        return '/session/' + sessionId + "?" + window.location.href.split("?")[1];
```
samccone, line 123: ~~lets use the url module `parse` and `format` instead of doing this ourselves.
http://nodejs.org/api/url.html~~

nm this is the front end

core/src/main/web/app/mainapp/mainapp.js
```diff
@@ -236,6 +241,18 @@
               loadNotebookModelAndResetSession(
                   notebookUri, uriType, readOnly, format, notebookModel, false, sessionId, false);
             });
+          },
+          byNotebookId: function(target) {
+            var notebookId = target.notebookId;
+            bkSession.load(notebookId).then(function(session) {
```
samccone, line 246: do we want to handle the error case here?

core/src/main/web/app/mainapp/mainapp.js
```diff
@@ -236,6 +241,18 @@
...
+              var serializedModel = session.notebookModelJson;
+              if (serializedModel && serializedModel != "") {
+		loadFromRetrievedSession(session, notebookId);
```
samccone, line 249: missing a space

samccone:
Can we expand the commit message to describe what this is doing and how it works. As this commit stands it is kinda hard to reason about how it is working due to it being deeply nested in some existing logic paths. 

something like..

> this commit allows us to hit a new url /XXXX/<id> to connect to ....

Just to help us in the future ya know :)

mforsyth:
Ready to go again @samccone

samccone:
Seems ok to me, however I am not super intimate with this piece of code in the source, perhaps we might want a once over by @scottdraves or @anglee just to make sure we are missing anything major

scottdraves:
not super intimate either but LGTM

samccone:
:shipit:

Issue #934, opened 2014-12-02
bkHelper.evaluate over an array of cells does not return a proper promise


ildipo:
The promise is resolved immediately, not when evaluation ends

PR #936, opened 2014-12-02
Ildipo/celloptions

ildipo:
This adds a number of improvements to beaker:
- code cells have names and tags - a new dialog is used to edit these
- from JavaScript it is possible to start evaluation of named cells, tagged cells
- on the fly code can be evaluated from JavaScript 
- JavaScript cells can (SHOULD) return a deferred object if they are executing "event driven" code
- new language plugin loading mechanism
- notebook is now rendered before language plugins are loaded
- progress report for all operations (load and evaluate)
- JavaScript beaker.showProgressReport() function for partial cell output display

This addresses the issues #934 #933 #932 #916


scottdraves:
was this tested with missing backends?

ildipo:
yes

PR #949, opened 2014-12-04
bunsen:  backup session whenever an evaluator comes online.

mforsyth:
This ensures that plugin processes won't leak if the page is closed
between when the evaluator is connected and when the next scheduled
backup occurs.


buob:
:+1:

PR #950, opened 2014-12-04
[bunsen] Reconnect evaluators after Beaker client has reconnected to the server

grin:
Steps to reproduce the reconnection issue in Bunsen:
- Open a notebook and create a python cell.
- Execute the cell and the correct response comes back.
- Turn off your WiFi.
- Wait about 30 sec-1 min until "reconnecting" is displayed at the top of the notebook.
- Turn on your WiFi.
- Wait until the "reconnecting" message disappears.
- Execute the single cell.
- A result is never returned -- the cell execution counter keeps counting.

This fix is implemented on the Bunsen branch but if it turns out to be correct it should also make sense to cherry-pick it to master.

The issue was that upon client disconnect iPython evaluators were being disconnected from the shell/stdin channels due to:
- https://github.com/twosigma/beaker-notebook/blob/bunsen/plugin/ipythonPlugins/src/dist/vendor/ipython2/kernel.js#L186
- https://github.com/twosigma/beaker-notebook/blob/bunsen/plugin/ipythonPlugins/src/dist/vendor/ipython2/kernel.js#L192 

but never reconnected back.
Note, I haven't tested other evaluators yet.

@scottdraves @anglee Could you please have a look at the fix? Thanks.

### DIFF
--- a/core/src/main/web/app/mainapp/mainapp.js
+++ b/core/src/main/web/app/mainapp/mainapp.js
@@ -613,6 +613,7 @@
               disconnected = true;
             },
             onReconnected: function() {
+              bkSessionManager.reconnectEvaluators();
               disconnected = false;
             },
             getStatusMessage: function() {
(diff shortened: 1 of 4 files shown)


grin:
@anglee @scottdraves any comments/feedback?

scottdraves:
will look at this later today

scottdraves:
i think this is a good idea, will apply to master

Issue #951, opened 2014-12-04
in advanced mode the cell menu is hard to click on


scottdraves:
it is mostly under codemirror's pre element.

scottdraves:
i just noticed that going back to 1.1.2 removes the problem so it was recently introduced so there may well be a better way of fixing it we should look at what has changed since.

PR #953, opened 2014-12-05
add opacity 0.999 to the code mirror editor, which has no practical effe...

scottdraves:
...ct on its appearance, but it does change the stacking group so it doesnot block the cell menu, hence fixing Issue #951


scottdraves:
here's a discussion of stacking groups: http://philipwalton.com/articles/what-no-one-told-you-about-z-index/

note the obvious fix of increasing the z-index (and adding positioning) to the menu in question does not work, hence the above extreme measure.

PR #965, opened 2014-12-10
beaker: Force close notebook when initated from bunsen UI.

mforsyth:
We don't want beaker to show its own confirm dialog; by that point
Bunsen has already shown a dialog in the outer frame.

### DIFF
--- a/core/src/main/web/app/helpers/window_message_service.js
+++ b/core/src/main/web/app/helpers/window_message_service.js
@@ -40,7 +40,7 @@
             bkBunsenHelper.saveNotebook(e.data.name);
             break;
           case 'close':
-            bkHelper.closeNotebook();
+            bkHelper.closeNotebook(true);
             break;
           case 'showStdoutStderr':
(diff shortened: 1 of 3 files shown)


core/src/main/web/app/helpers/window_message_service.js
```diff
@@ -40,7 +40,7 @@
             bkBunsenHelper.saveNotebook(e.data.name);
             break;
           case 'close':
-            bkHelper.closeNotebook();
+            bkHelper.closeNotebook(true);
```
buob, line 43: Would it be unnecessary to send a force flag with the window message?  I guess we'll always want to force from bunsen, huh?

mforsyth, line 43: THat's right.  The close notebook UI for us is in Bunsen.  If bunsen sends the message, there's already been a confirm dialog

buob, line 43: Sounds good. :+1:

buob:
:+1:

PR #966, opened 2014-12-10
Spot/reconnect

scottdraves:
copied from https://github.com/twosigma/beaker-notebook/pull/950

i found that after being disconnected i couldn't get the notebook back to its original state, it still said i was offline, but i could evaluate cells, though sometimes i had to cancel an evaluation and then retry.

so needs more work IMO but a step in the right direction.


grin:
@scottdraves I'm trying to fix an issue in Bunsen related to this PR and #950 where the IPython kernel keeps restarting itself every 30 sec. There seems to be a race condition where the kernel isn't ready before hitting the timeout so it disconnects and then reconnects again...
The easiest fix would be to increase the cometd `timeout` value (30 sec by default) as defined here: http://docs.cometd.org/reference/java_server.html#java_server_configuration

...do you maybe know how to change this value in the java backend?

grin:
never mind, figured how to increase the timeout in the backend

PR #981, opened 2014-12-19
Fix Datatable glitch

buob:
So this fixes the bug, but I had to make a pretty big architectural change (instantiating the output `childScope`s off of the current `scope` instead of the `$rootscope`), and I'm not sure how that might affect some of our edge cases.  Even so, the solution is basically to add a random attribute to our `dtOptions` object to trigger the angular-datatable's watcher in order to rerender, which is slightly janky.  I would have expected there to be a public method to rerender the table with the same approach that the directive controller takes, but it seems like there are only factory methods that expose services that re-render the components of the datatable itself.

I've opened an issue on the library, we might want to table this until it's either addressed or perhaps switch to a library that's less complex, more angular-friendly, and has better public interaction.  @samccone has had success using a library in the publications side of bunsen that might be worth considering.

### DIFF
--- a/core/src/main/web/app/mainapp/components/notebook/codecelloutput-directive.js
+++ b/core/src/main/web/app/mainapp/components/notebook/codecelloutput-directive.js
@@ -102,6 +102,7 @@
         $scope.toggleExpansion = function() {
           if ($scope.$parent.cellmodel.output.hidden) {
             delete $scope.$parent.cellmodel.output.hidden;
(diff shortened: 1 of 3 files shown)


scottdraves:
thanks jake.  can you please link this PR to the issue you filed?

scottdraves:
https://github.com/l-lin/angular-datatables/issues/157

PR #991, opened 2014-12-31
Add `new folder` button to save as menu

buob:
new-folder mov

This is a minimum UI, and probably not ideal.  Might be best to have @bradicdesign take a look at this feature as I didn't want to make too many design decisions.


scottdraves:
thanks great.

yes for ux review, but let's not let that hold up getting the functionality in.

PR #1003, opened 2015-01-07
Fixed dialog regression

ildipo:


ildipo:
The dialog looks fine but the scope is not updated when the user changes value, so any change is not saved.

PR #1006, opened 2015-01-09
[bunsen] Detect server crash and show an error message

grin:
The window prompt logic is partially based on the logic reverted in:
https://github.com/twosigma/beaker-notebook/commit/fc9ac9d307ef64151a45652f0d03f3fb226da4ef

### DIFF
--- a/core/src/main/web/app/mainapp/mainapp.js
+++ b/core/src/main/web/app/mainapp/mainapp.js
@@ -605,6 +605,31 @@
           }
         });
 
+        $scope.promptToRefresh = (function() {
+          var prompted = false;
+          return function() {
+            if (prompted) { // prevent prompting multiple at the same time
+              return;
+            }
+            prompted = true;
+            bkCoreManager.show2ButtonModal(
+                "Beaker server crashed. All running cells were interrupted.<br>" +
+                "Please refresh the page to continue.",
+                "Error",
+                function() {
+                  window.top.location.reload();
+                },
+                function() {
+                  // "Not now", hijack all keypress events to prompt again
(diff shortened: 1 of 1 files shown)


core/src/main/web/app/mainapp/mainapp.js
```diff
@@ -616,8 +641,18 @@
               disconnected = true;
             },
             onReconnected: function() {
-              bkSessionManager.reconnectEvaluators();
-              disconnected = false;
+              bkSessionManager.isSessionValid()
```
mforsyth, line 644: So, what well isSessionValid() resolve to if the call to ../beaker/rest/session-backup/getExistingSessions returns a 503?

grin, line 644: Why would it return 503? This code is evaluated after the client was successfully reconnected to the server...specifically, to the new server instance that was started by marathon after the original instance crashed.

mforsyth, line 644: Oh, I see.  So when a beaker instance crashes, the client is already successfully reconnecting to a new instance?  (it's just that session is not valid anymore)?

grin, line 644: Yes, the session becomes invalid. To be more precise: there is a short period in between the server crash and the launch of a new instance where the client is disconnected and during that time it polls for a connection. When the connection is finally made the client validates the session.

mforsyth, line 644: Ah, I see.  Thanks for explaining :+1:

buob:
:+1:

PR #1019, opened 2015-01-12
Add sort capability to file system

buob:


scottdraves:
sort should be most recent on top!

buob:
ha, this is true.  Updating...

buob:
Updated, and tests are passing locally.

buob:
Hold up on this...

Issue #1026, opened 2015-01-13
Alt-UP to move between code cells does not reposition window correctly


ildipo:
It seems that the window is still UP by the size of the top bar.

PR #1036, opened 2015-01-13
Fix for #1026: when moving to previous code cell scroll the window a

ildipo:
little down so that more of cell editor appears.


scottdraves:
should not scroll if not necessary

PR #1056, opened 2015-01-16
Make `dropdown-promoted` directive

buob:


buob:
This passes locally and on travis (but Travis be buggin').  I just reran the build again to get the little green check mark.

PR #1098, opened 2015-01-30
fix name of pandas for pip

scottdraves:
just need to remember this


scottdraves:
apt-get results in v0.13.1-2ubuntu2

scottdraves:
pip results in v0.15.2

scottdraves:
ready to merge now

PR #1103, opened 2015-01-30
WIP File Import

buob:


core/src/main/web/app/mainapp/mainapp.js
```diff
@@ -355,6 +356,16 @@
...
             },
+            fromImport: function(sessionId) {
+              var notebook = $sessionStorage.importedNotebook;
+              var notebookUri = null;
+              var uriType = null;
+              var readOnly = true;
+              var format = null;
+              var notebookModel = bkNotebookVersionManager.open(notebook);
+              loadNotebookModelAndResetSession(
+                  notebookUri, uriType, readOnly, format, notebookModel, false, sessionId, false);
```
buob, line 369: I'm just loading the notebook here right from the JSON received from the server which is resulting in inputs as arrays, yielding empty input fields.

scottdraves:
The problem comes from not using the standard importer like this: https://github.com/twosigma/beaker-notebook/blob/master/core/src/main/web/app/mainapp/mainapp.js#L328

looks like that wants the document as a string instead of already parsed into json so maybe need to change the servlet to return text/plain?

also, i am concerned by the use of ngstorage.  i don't know what this is and why it's needed, the other import and file IO functions pass the data through callbacks instead of storing it in this relatively global location.

buob:
Yeah, I don't love it either.  Any thoughts on a better technique given our strategy of using the redirect in order to load a notebook?

scottdraves:
You could skip doing the redirect and instead directly connect the menu item to openNotebook with a newly defined uriType "fileupload" that opens the native dialog with a callback that has the text.

scottdraves:
but we can always leave that as an improvement for the future, for now it's fine to just get it working by getting the document as a string instead of already parsed as json.

Issue #1163, opened 2015-02-07
remove references to 'plugin manager'


dmitri-at-twosigma:
related to this:
https://github.com/twosigma/beaker-notebook/commit/bcead3615fbe3fe87f191cfe525afba31e689bbe

and:
https://github.com/twosigma/beaker-notebook/issues/854

there's still 35 references to 'plugin' including in user-facing documentation.
https://github.com/twosigma/beaker-notebook/search?utf8=%E2%9C%93&q=%27plugin+manager%27

scottdraves:
The only use I could find in external doc was

```
Plugins
All of the languages available in Beaker are added via plugins, so if we've missed your favorite 
language, you can create a plugin, and it will be treated as a first-class citizen. Plugins can also
easily extend the menus and output media types.
```

and that seems fine.  am i missing any?

okhobb:
this is the one that seemed user-facing:

```
To add another language to this notebook, select <b>Notebook</b>&nbsp;<span style=\"color: rgb(51, 51, 51); font-family: 'Helvetica Neue', Helvetica, 'Segoe UI', Arial, freesans, sans-serif; font-size: 14px; line-height: 22.3999996185303px;\">→</span>&nbsp;<b>Plugin manager</b> from the menu on top of the screen.</p>
```

scottdraves:
thanks that should do it.

PR #1172, opened 2015-02-08
update default notebook

scottdraves:
change plugin manager to language manager and convert text to markdown.


scottdraves:
closes https://github.com/twosigma/beaker-notebook/issues/1163

PR #1177, opened 2015-02-09
Add MathJax capabilities to Markdown Cells

buob:


scottdraves:
need to use ptsans while editing the markdown too not just displaying it.

scottdraves:
also this introduces a bug where if you keep clicking on the editor it just keeps getting bigger and bigger.

buob:
markdown mov

I'm not seeing the issue you're describing, could you give me a little more info?

(this is passing locally)

buob:
Tabling until after v1.2; the epiceditor dependency isn't as flexible as we need it to be

PR #1191, opened 2015-02-18
Fixed error display due to python exception.

ildipo:


scottdraves:
needs more explanation? how is exn generated?

ildipo:
It seems that python delivers output and reply messages in a 'pseudo random' order when there's an exception in the code.
The logic added is that:
1) we refuse any output message if the cell if the cell has finished evaluating (reply message)
2) if we receive an error output we display the error and refuse to display anything that comes after it

scottdraves:
here's the cell that produces the problem:

```
import matplotlib.pyplot as plt
plt.subplots()
"Hello" + 42
```

Issue #1084, opened 2015-01-23
page unload should checkpoint


scottdraves:
this is a regression.  we used to checkpoint on page unload (eg close the tab or reload).

scottdraves:
regressed with https://github.com/twosigma/beaker-notebook/commit/b93e7a61f257b91c5c6dcf36c11041870f9bee12
but strangely fixing the typo window.unload to window.onunload does not fix.

scottdraves:
http://stackoverflow.com/questions/4945932/window-onbeforeunload-ajax-request-problem-with-chrome

scottdraves:
both reload and close work back at 2b5df57 on July 24

PR #1192, opened 2015-02-19
fix session backup/checkpoint when reloading page

scottdraves:
move onDestroy back to onbeforeunload, unclear why but initiating the session backup from onunload does not work.  also fix the typo and window.unload to window.onunload though so bkEvaluateJobManager.cancel gets called.  this closes Issue #1084


scottdraves:
hm, this does not work if the page is closed (command-w).  it does work if the page is reloaded.
let me double check if regression applies to close.

scottdraves:
it does.

PR #1214, opened 2015-02-26
Ildipo/newbuild

ildipo:


scottdraves:
I tested this and it works and the basic technique looks good.
To merge though it needs two ways of building one for development and one for distribution.
For distribution it can gulp the files like it does now, and even minimize them.
But for development it should keep the JS files separate and untouched.
That means two different index_template.html versions, one with the single gulped JS file, and one with all the files.
The version with all the files should be generated from the gulp config, so we only have one list of all our JS source.
ditto for plugin/init/addoutputdisplays.js

in fact we should do even more of this, especially for all the python code in launcher/windows and the JREs in launcher/windows and launcher/mac.

ildipo:
We now have two grade 'modes':
grade run  (and 'grade build') build and runs the optimized version (which is also what is run by 'grade check')
gradle debug (and 'gradle builddebug') build and run the 'old version with all the individual files for easy debugging'

Issue #1180, opened 2015-02-10
better error reporting and recovery from missing or bad backend


scottdraves:
currently if a backend fails you just get an error on the JS console.

Instead, you should get a dialog that says it failed, shows any recovered error text, and
recommends that you install it if you haven't (with a link to our web page for further instructions).
it should also suggest changing the prefs, with a link that opens that window on top of this one.

And the error dialog has an OK button which just returns to the Language Manager.

The button for that language, if it failed to start, would still be gray.  If you edit the prefs, or go and download it, then you can click it again, and try to start it again.

scottdraves:
We cannot solve this properly until we solve https://github.com/twosigma/beaker-notebook/issues/1236
In the meantime the error message will just have to ask the user to restart Beaker.

PR #1247, opened 2015-03-05
Error dialog when a evaluator plugin fails to load

scottdraves:
close Issue #1180

### DIFF
--- a/core/src/main/web/app/helpers/evaluatepluginmanager.js
+++ b/core/src/main/web/app/helpers/evaluatepluginmanager.js
@@ -131,13 +131,10 @@
                       return factory.create(evaluatorSettings).then(function(ev) { deferred.resolve(ev); });
                     } else {
 		      // or should someone pass in a scope to use, like with a method on the evaluatorLoadQueue?
-		      $rootScope.pluginId = name;
-		      $rootScope.closeErrorDialog = function () {
-			  $rootScope.modal.close();
-		      }
-		      $rootScope.modal = $modal.open({backdrop: true,
-						      backdropClick: true,
-						      template: JST['helpers/plugin-load-error']()});
+                      $rootScope.pluginId = name;
+		      var modal = $modal.open({backdrop: true,
+					       backdropClick: true,
+					       template: JST['helpers/plugin-load-error']()});
 		      deferred.reject("no factory for evaluator plugin");
 		    }
                   }, function(err) {
--- a/core/src/main/web/app/helpers/evaluatepluginmanager.js
+++ b/core/src/main/web/app/helpers/evaluatepluginmanager.js
@@ -19,7 +19,7 @@
 (function() {
   'use strict';
   var module = angular.module('bk.evaluatePluginManager', ['bk.utils']);
-  module.factory('bkEvaluatePluginManager', function(bkUtils, $rootScope, $modal) {
+  module.factory('bkEvaluatePluginManager', function(bkUtils, $modal) {
     var nameToUrlMap = {};
     var nameToVisualParams = {};
     var plugins = {};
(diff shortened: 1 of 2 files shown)


core/src/main/web/app/helpers/evaluatepluginmanager.js
```diff
@@ -129,9 +129,18 @@
                   .then(function(factory) {
                     if (factory !== undefined && factory.create !== undefined) {
                       return factory.create(evaluatorSettings).then(function(ev) { deferred.resolve(ev); });
-                    } else
-                      deferred.reject("no factory for evaluator plugin");
+                    } else {
+		      // or should someone pass in a scope to use, like with a method on the evaluatorLoadQueue?
+                      $rootScope.pluginId = name;
```
buob: did `$rootScope.$new({ pluginId: name })` etc. etc. not work?

scottdraves: i tried a few things like that but did not fly.  would be happy to see it though.

core/src/main/web/app/helpers/evaluatepluginmanager.js
```diff
@@ -129,9 +129,18 @@
...
+		      var modal = $modal.open({backdrop: true,
+					       backdropClick: true,
+					       template: JST['helpers/plugin-load-error']()});
```
buob: You could also pass in a variable directly to the template invocation (especially considering it's not going to change at all after it's displayed), trying to find an example...

buob: here's a [js file](https://github.com/twosigma/beaker-notebook/blob/2b354551cca27d8c42663ae0f2234eecb77abf12/core/src/main/web/plugin/menu/file_filesys.js#L39) passing in a variable to [template](https://github.com/twosigma/beaker-notebook/blob/2b354551cca27d8c42663ae0f2234eecb77abf12/core/src/main/web/app/template/opennotebook.jst.html#L33)

buob: Might have to use [this syntax](https://github.com/twosigma/beaker-notebook/blob/2b354551cca27d8c42663ae0f2234eecb77abf12/core/src/main/web/app/template/opennotebook.jst.html#L40) (like extension) -- I was having trouble using the other in our environment

core/src/main/web/app/helpers/evaluatepluginmanager.js
```diff
@@ -129,9 +129,18 @@
                   .then(function(factory) {
                     if (factory !== undefined && factory.create !== undefined) {
                       return factory.create(evaluatorSettings).then(function(ev) { deferred.resolve(ev); });
-                    } else
-                      deferred.reject("no factory for evaluator plugin");
+                    } else {
+		      // or should someone pass in a scope to use, like with a method on the evaluatorLoadQueue?
+                      $rootScope.pluginId = name;
+		      var modal = $modal.open({backdrop: true,
```
buob: you probably don't need this `modal` var stored, do you?

core/src/main/web/app/helpers/evaluatepluginmanager.js
scottdraves: xlnt thank you, indeed the ${} syntax is required

core/src/main/web/app/helpers/evaluatepluginmanager.js
scottdraves: right, done

core/src/main/web/app/helpers/evaluatepluginmanager.js
```diff
@@ -19,7 +19,7 @@
 (function() {
   'use strict';
   var module = angular.module('bk.evaluatePluginManager', ['bk.utils']);
-  module.factory('bkEvaluatePluginManager', function(bkUtils) {
+  module.factory('bkEvaluatePluginManager', function(bkUtils, $rootScope, $modal) {
```
buob: You probably don't need this dep on `$rootScope` anymore, right?

buob:
Looks good other than small comment :+1:

core/src/main/web/app/helpers/evaluatepluginmanager.js
scottdraves: thanks, removed

PR #1280, opened 2015-03-10
remove font face

samccone:
The font tag is no longer supported in html5, instead we should be using css to style our page.


buob:
:+1:

samccone:
thanks @scottdraves

PR #1281, opened 2015-03-10
fix for issue #937 when the user does not output a \n

### LINKED ISSUE #937: too fast stdout or stderr output kills beaker client
execute:

while(true) { System.out.println("FOOBAR"); }

in groovy or java

ildipo:
Fixes #937

in Groovy:

for (i=0; i<1000; i++)
  System.out.print(i+"
01234567890123456789012345678901234567890123456789 ")

  force a line break every 240 characters.


scottdraves:
what about `System.out.print("nonewline")`?
that output is lost.

PR #1282, opened 2015-03-11
Flush remaining output when evaluation finishes

ildipo:


plugin/jvm/src/main/java/com/twosigma/beaker/jvm/object/SimpleEvaluationObject.java
```diff
@@ -467,6 +469,29 @@ private synchronized void doAppendError(String s) {
     }
   }
 
+  private synchronized void doFlush() {
```
buob, line 472: Why `doFlush` instead of just `flush` or `flushOutput`?

ildipo, line 472: doFlush -> it does something and what it does is a flush.
this mimics other doXXX functions in the same file

buob, line 472: for the sake of argument, don't all functions do something? :smile:

ildipo, line 472: well, philosophically speaking functions compute (without side effects) and other do something with side effects - this has side effects. :-)

PR #1285, opened 2015-03-11
Move new cells into viewport after adding

buob:


scottdraves:
could not reproduce the fix.  tabling this in favor of other issues.

Issue #408, opened 2014-07-25
Mac OS X 10.7.5 - Beaker 0.55 missing nib file?


randompi:
The application won't launch for me.  When run from the command line:

localhost:MacOS user$ ./Beaker 
2014-07-25 14:04:26.601 Beaker[12959:707] Unable to load nib file: MainMenu, exiting

scottdraves:
Thanks for the bug report.  We didn't have a 10.7 machine to test on, just 10.9 and 10.8.
Sorry it didn't work for you.
I did use what I thought were XCode settings to work with 10.6 and above but there must be more to it, I will look into it.

This article http://apple.slashdot.org/story/14/02/27/1513248/apple-drops-snow-leopard-security-updates-doesnt-tell-anyone
says that 20% of general users still have 10.6 so I would guess even more have 10.7.

melmac48:
currently don't support back beyond 10.8. Will look into expanding text matrix and support in v2

eubrandt:
I have a 10.6 machine if you'd like reports of trial runs there.

(I can take a crack at it myself, but my guess is Xcode 3 may not cut it either.)

doctorbrassica:
Can confirm this error in 10.7.4, late 2011 machine.  Also confirmed there are plenty of 10.7 machines out there.

scottdraves:
Thx Jonathan.  This is definitely moving up our agenda.

scottdraves:
eubrandt, is that you eli?

powderly:
Hey Scott,

just to thricely confirm

MacOS User$ ./Beaker
2014-12-28 03:29:02.558 Beaker[77935:707] Unable to load nib file: MainMenu, exiting

on 10.7.5
2012 macbook pro

scottdraves:
thanks james will def get this one in the next release.

scottdraves:
turn off Base Internationalization?
http://stackoverflow.com/questions/18917278/base-lproj-mainmenu-xib-is-not-available-when-compiling-for-targets-before-mac-o

scottdraves:
well that doesn't work, it just leaves out the nib files completely, so the app fails to run even on the machine where it was built.  i can probably figure out how to copy them into the right place though, one way or another.  the right place is Contents/Resources (not in Base.lproj subdir where Xcode was putting them).

As an experiment I made this change on the finished app file on a 10.7 machine and it does in fact start (yea!) but then nginx fails to start with this error:

```
[...]

PR #1286, opened 2015-03-11
Fix for old Mac OS X

scottdraves:
close Issue #408 by copying the nib files to where pre-10.8 Mac OS X can find them, and also by including nginx that has been built on a 10.7 Mac.


ildipo:
looks good. are we including the 'old' nginx also on newer macs?

scottdraves:
yes fortunately they do have forward compatibility.

scottdraves:
the version of nginx is the same (1.6.2) it's just linked with an old SDK.

core/build.gradle
```diff
@@ -130,7 +130,8 @@ task (compileCssDebug, type: Exec) {
 
 task (copyNginx, type: Exec) {
   if (mac()) {
-    def brewLocation = 'which brew'.execute().text[0..-10]
+//  def brewLocation = 'which brew'.execute().text[0..-10] // use nginx from current machine
```
buob: We can probably kill this comment, huh?  Or are we going to go back to it in the near future?  Or does it add value to the new code?  If it's just to keep a record, we could always come back in source control if we need it, right?

PR #1309, opened 2015-03-23
Mf/namespace css task

mforsyth:
Hi @ildipo ... this limits the "namespacing of CSS" behavior that I have mentioned to only the embedded build (which is the only place where it will be relevant).

It also publishes certain static assets which are meant to be used externally (the js and css assets needed for embeddable beaker).


ildipo:
The output stuff should go in a different directory (e.g. output) and it should not be added to the repo

ildipo:
I tried gradle check and it works - seems to fail on travis....

ildipo:
Actually technically speaking data is NOT build - it is used to build. You can re-generate its content through a gradle target but its purpose is to provide input to build stuff.

mforsyth:
@ildipo ready

PR #1310, opened 2015-03-23
Ildipo/moreserialization

ildipo:
This PR is a refactoring of serialization and deserialization code used for autotranslation in JVM-based, Python2, Python3 and R language backend. It improves auto-translation, adds type auto-detection and improves display.


scottdraves:
in the d3 tutorial the output of beaker.graph is no longer pretty printed
screen shot 2015-03-23 at 1 26 13 pm

more importantly, when i run the javascript, it doesn't work

ildipo:
This PR has been updated fixing the above bug (and adding even more functionalities such as proper serialization-deserialization-decoupling) to JavaScript

scottdraves:
"Beaker takes care of the situation by silently transforming vectors of length 1 to scalars when these are returned as results of an evaluation."

i thought we discussed this -- R experts say (and I agree) that we shouldn't make this automatic conversion

scottdraves:
otherwise this PR is looking really good!

ildipo:
I had a long chat with Peter Berlin last week (our TwoSigma R top user and package maintainer) and he thinks it is actually a good idea provided that inside R everything stays as is. He thinks that it will simplify modelers life since they are used at having to address this on their own.

PR #1313, opened 2015-03-24
Enable MathJax in Markdown Cells

buob:


scottdraves:
not working at all for me, neither $\pi$ in markdown and regular latex cells

PR #1321, opened 2015-03-25
Added dependencies to speed up re-build

ildipo:
Added dependencies to speedup re-build
Added 'realclean' target that really cleans the project

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -46,6 +46,11 @@ task clean (type: GradleBuild) {
   tasks = ['clean']
 }
 
+task realclean (type: GradleBuild) {
+  buildFile = mainGradle
+  tasks = ['realclean']
+}
+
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -46,6 +46,11 @@ task clean (type: GradleBuild) {
   tasks = ['clean']
 }
 
+task realclean (type: GradleBuild) {
```
buob, line 49: Hmm... I wonder if there's a better name for this?

Or... is there a reason one might want to `clean` without doing a `realclean`?  If not, maybe we could just add it to the normal clean?

ildipo, line 49: Usually 'clean' deletes the build artifacts while 'realclean' deletes everything that is not in the distributed software.

PR #1333, opened 2015-03-26
Spot/1025

ildipo:


buob:
Could we convert the tabs and four-space tabs to two-space spaces?

PR #1353, opened 2015-03-29
kdb support for Beaker

nyscotsman:
Here's a first cut at a kdb plugin for Beaker. I've added some support for tables and dicts (at least at the top level) and auto translation is supported with the `.bk.g` (get) and `.bk.s` (set) functions. There's autocomplete too for names in the default namespace as well as standard q functions.

I'm including the `c.java` kdb client from code.kx.com; I think this is OK given their licensing.

Really appreciate any feedback you guys have, especially if there's a better way to handle the json serialization (right now it doesn't support nested Flips or Dicts)


scottdraves:
wow thanks this sounds great.

what is this c.java?  is this the license? http://code.kx.com/wsvn/code/kx/kdb%2B/license.txt

in order to take a PR I would need you to sign and return the CLA: http://beakernotebook.com/cla.zip

nyscotsman:
c.java is the java kdb client that kx distributes - that's the right link to the license. Let me look over the CLA and get back to you.

PR #1359, opened 2015-03-30
style: Fix jscs formatting issues on core.js

samccone:


scottdraves:
whatever tool you used (jscs?) broke the code.
please do not use that tool on beaker.

samccone:
@scottdraves what part of the code broke?

The PR's goal was to update the syntax to follow javascript best practices (JSHint) and style (as per the google style guide).

It should have ideally not impacted the functionality, however since it seems like it did can you point out what broke and I can investigate. We can reasonably deduce that there is some other code that is relying on some global state mutation or "undefined" return that I can most likely fix. 

Thanks

scottdraves:
if you edit a notebook and then close it, a dialog comes up to ask you if you want to save.
the dialog was broken because the quotes around some html were mangled.

samccone:
k thanks I will update

Issue #1362, opened 2015-03-31
markdown cells should use code mirror


scottdraves:
then we get  https://github.com/twosigma/beaker-notebook/issues/704 and https://github.com/twosigma/beaker-notebook/issues/1345

buob:
Are we okay with them behaving like normal code cells?  Input and output?

scottdraves:
nothing should change except editing would be enabled.

scottdraves:
related https://github.com/twosigma/beaker-notebook/issues/1345

PR #1375, opened 2015-04-01
Fix html in markdown

aaddaamm:


scottdraves:
this doesn't work if you have a < or &, they get converted to &amp;lt; and &amp;amp;

i think the basic approach of altering the cellmodel.body is dangerous at best, but perhaps unavoidable with contenteditable attribute.

maybe best to combine this fix with https://github.com/twosigma/beaker-notebook/issues/1362

PR #1380, opened 2015-04-01
Focus a new cell on insert instead of initialize

buob:


scottdraves:
every time i click in the middle of a markdown cell to try to edit it, the cursor moves to the end.

buob:
Oh, yeah this is a bug that isn't introduced by this PR -- but it should be fixed in the codemirror fixups

PR #1383, opened 2015-04-02
Don't focus an already focused markdown cell

buob:


scottdraves:
thanks for the quick fix

Issue #1387, opened 2015-04-02
when you insert a cell it should get the focus with cursor


scottdraves:

buob:
Is this broken on all cells?  It's working for me on code cells but not markdown.

ildipo:
This now works on every cell (text, markdown, code, sections)

PR #1401, opened 2015-04-06
Revert "Adding Markdown cells now default in edit mode"

ildipo:
Reverts twosigma/beaker-notebook#1340

this introduces a 'jumping notebook' bug


buob:
This is not a clean reversion of that commit.  The line of code that actually caused the "jumping notebook" bug was already removed here: https://github.com/twosigma/beaker-notebook/commit/f3a190fb8d31baa487139e352bae5d1534295804

scottdraves:
what's not clean?

the fix you cite introduces even worse bugs https://github.com/twosigma/beaker-notebook/issues/1387

we need to get master stabilized but more important than that we need to make an internal release.

PR #1406, opened 2015-04-07
Jb/js stdout

buob:


core/src/main/web/plugin/evaluator/javaScript.js
```diff
@@ -747,6 +748,13 @@ define(function(require, exports, module) {
             beakerObj.setupBeakerObject(modelOutput);
             beakerObj.notebookToBeakerObject();
             var beaker = beakerObj.beakerObj;
+            var evaluation = { status: 'FINISHED', outputdata: [] };
+            beaker.print = function(input) {
```
scottdraves: this should go in setupBeakerObject()

scottdraves:
needs a way to print error output as well.

core/src/main/web/plugin/evaluator/javaScript.js
```diff
@@ -747,6 +748,13 @@ define(function(require, exports, module) {
...
+              evaluation.outputdata.push({
```
scottdraves: this will only show the output when evaluation finishes.  but each line of output should appear as soon as it is done.
receiveEvaluationUpdate should be called right away.

PR #1414, opened 2015-04-08
Small bug in QHOME override change; fixed up tutorial description

nyscotsman:


scottdraves:
i am confused by the new overview. does this plugin give you K or Q?

nyscotsman:
The plugin provides access to a kdb instance, which is the database program. kdb's inbuilt languages are k and q; q is the more readable extension to k. So the plugin lets you execute both k and q code on the connected kdb instance; but the bias is towards q (for the autocomplete and editor support). The majority of people will write code in q when using kdb.

Issue #1420, opened 2015-04-08
cell movement skips over section


scottdraves:
if i click the "move down" button on the first HTML cell in this notebook, it goes to the bottom, below B.  it should be above B.  if the section were closed then this would be right.

before:
![screen shot 2015-04-08 at 12 01 53 pm](https://cloud.githubusercontent.com/assets/963093/7049414/43471b0e-dde7-11e4-8e41-b36fc9baa85b.png)

after:
![screen shot 2015-04-08 at 12 02 07 pm](https://cloud.githubusercontent.com/assets/963093/7049418/4fd0e9fe-dde7-11e4-82d8-da87ccb3ecaf.png)

Issue #1421, opened 2015-04-08
cannot move a cell out of a section


scottdraves:
the up (resp down) button is grayed out if the cell is the first (resp last) in a section.

PR #1431, opened 2015-04-09
Use separate move methods for sections and code cells

buob:


scottdraves:
this is supposed to fix https://github.com/twosigma/beaker-notebook/issues/1420

there are two problems:

1) the buttons do not highlight correctly (they show inactive even when they now work)

2) if the section is closed then the old behavior is correct

buob:
I added a commit that fixed the first problem -- that was actually the "cause" (or manifestation) of #1421.  The second problem I'm not sure I understand.  If a section is collapsed and you move a cell down into the section it should go to the bottom of that section?  That's the functionality on master now.

scottdraves:
:+1: 

i think the section cell needs to open up when it receives a cell to avoid the vanishing problem.
this happens in some other circumstances, i'll file another issue for it.

PR #1465, opened 2015-04-14
CodeMirror in Markdown & Markdown in Section Cells

buob:


scottdraves:
when editable, click is being caught by the handler that puts the cursor at the beginning or the end.

scottdraves:
highlight mode doesn't seem to be set to markdown.

scottdraves:
thanks, that's a big improvement!

when the markdown formats to display, it shows the old content before updating it.
it should just show the new content.

actually that's a pre-existing bug so it does not need to be fixed in this PR.

scottdraves:
:dolphin:

Issue #1475, opened 2015-04-15
markdown cells flash white during format


scottdraves:
when you press command-enter to format the markdown cell, they vanish and you get white space until they reappear with the formatting.  should say up and then transition seamlessly.

PR #1479, opened 2015-04-15
Bug Fixes

buob:


scottdraves:
the link clicking fix looks good, but the min-height restriction seems to cause a problem where now every time i render the section, the cells below move up and down briefly.  they go back to the same place at least, but having them jump around is very annoying.  do we need https://github.com/twosigma/beaker-notebook/issues/1475 in order for this to work right?

buob:
Yeah, that fix would also fix the jumping issue, but I think the fix is going to be a lot more complex -- so I just lowered it to 1em instead of what it was before, so the jumping should be gone now until that issue is fixed.

scottdraves:
can the minimum just be the same as the header during edit, then it would not jump at all right?

buob:
Yup, that's what the update I just pushed reset it to.

PR #1487, opened 2015-04-16
Remove flash from markdown-editable edit/preview

buob:


scottdraves:
i can't edit contents of sections at all with this.

buob:
So before it would let you edit but it wouldn't show the cursor, which I'm not sure about.  This fix isn't that clean, it just sets the content twice, once before and once after render, but it shows the cursor again without the flash.

scottdraves:
editing still broken.  after about 20 seconds of clicking and editing the text stops changing but i can still see the cursor moving around.

PR #1495, opened 2015-04-17
Bug Fixes

buob:


scottdraves:
markdown cells are too short now, and change size when i click on them to edit.

buob:
Whoops, didn't test without content.  That's all set.  Also added a couple more commits.

PR #1496, opened 2015-04-17
Makeover the advanced mode insert cell button

buob:
screen shot 2015-04-17 at 11 45 19 am

Here's a way to have a quick-insert button as well as more options (with an added line for context)

### DIFF
--- a/core/src/main/web/app/mainapp/components/notebook/newcellmenu-directive.js
+++ b/core/src/main/web/app/mainapp/components/notebook/newcellmenu-directive.js
@@ -99,6 +99,17 @@
         $scope.$on('cellMapRecreated', function() {
           recentlyAddedLanguage = null;
         });
+      },
+      link: function(scope, element) {
+        element.on('click', function(e) {
+          if (e.target !== element[0]) return;
+
+          scope.newCodeCell(scope.defaultEvaluator());
+        });
+
+        scope.$on('destroy', function() {
+          element.off('click');
+        });
       }
     };
   });
--- a/core/src/main/web/app/mainapp/components/notebook/newcellmenu.jst.html
+++ b/core/src/main/web/app/mainapp/components/notebook/newcellmenu.jst.html
@@ -15,11 +15,13 @@
 -->
 <div class="btn-group new-cell" ng-class="last && 'last'">
   <button ng-click="newCodeCell(defaultEvaluator())" class="btn btn-primary insert-cell" ng-class="size && 'btn-'+size">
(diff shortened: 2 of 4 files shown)


scottdraves:
ok cool that's progress.  it has the functionality we need but it doesn't look right to me.

can you please try this:   move it to the right, out of the margin and just between the cells.  make it half as high as it is.  replace the language badge with a "+" with blue background.  remove the long blue line.

core/src/main/web/app/mainapp/components/notebook/newcellmenu.jst.html
```diff
@@ -17,9 +17,16 @@
   <button ng-click="newCodeCell(defaultEvaluator())" class="btn btn-primary insert-cell" ng-class="size && 'btn-'+size">
     Insert {{defaultEvaluator()}} Cell
   </button>
+  <bk-language-logo ng-if="!last"
+                    class="advanced-only"
+                    ng-click="newCodeCell(defaultEvaluator())"
+                    bg-color="{{getEvaluatorDetails(defaultEvaluator()).bgColor}}"
```
samccone: Seems suspicious to need {{ }} here since angular directive attributes are evaluated on the scope during the compile/link phase right?

buob:
Updated

screen shot 2015-04-29 at 11 49 00 am

scottdraves:
cool, that's even better.  what it still needs is:
1) move to the right.  it is sticking out by about one pixel.
2) half the vertical size please.  the "+" will need to be reduced.  might have to eliminate it completely.
3) double the width of the default insert button.
4) remove the blue, right-pointing triangle on the right.
5) put more horiz space between the two buttons, same as in non-advanced mode.
6) make it (default insert) clickable across the entire width of the cell.

buob:
Updated

screen shot 2015-04-29 at 3 12 06 pm

core/src/main/web/app/mainapp/components/notebook/newcellmenu-directive.js
```diff
@@ -99,6 +99,13 @@
         $scope.$on('cellMapRecreated', function() {
           recentlyAddedLanguage = null;
         });
+      },
+      link: function(scope, element) {
+        element.on('click', function(e) {
```
samccone, line 104: you will need to unbind this on destroy otherwise this is going to leak

buob, line 104: updated \/

PR #1518, opened 2015-04-24
When loading a notebook by http from same host, don't proxy.

mforsyth:
NOTE: Travis currently failing, but with same errors as master.

### DIFF
--- a/core/src/main/web/app/utils/utils.js
+++ b/core/src/main/web/app/utils/utils.js
@@ -160,9 +160,20 @@
             .error(deferred.reject);
         return deferred.promise;
       },
-      loadHttp: function(url) {
+      loadHttp: function(logicalUri) {
+        var logicalUrl = new URL(logicalUri);
+        if (logicalUrl.hostname == window.location.hostname &&
+            logicalUrl.port == window.location.port) {
+          var loadingUri = logicalUri;
+          var queryParams = {};
+        }
+        else {
+          var loadingUri = "../beaker/rest/http-proxy/load";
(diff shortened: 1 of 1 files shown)


scottdraves:
hi what's the motivation for this PR?

mforsyth:
In an external system, when a user opens a notebook, it loads the notebook contents from a url that is from the same host & port as the main document containing the Beaker interface.  For this reason, there's no need to proxy the request through the Beaker server, since CORS will never block the AJax request (it's for the same host and port as the main window).

Also, if we DID try to proxy the request, it wouldn't work, because the Beaker server (being a container running inside external network) won't be able to load the notebook contents - it woudn't have the right cookies to gain access, and actually DNS resolution would be different from inside the container & the request wouldn't go to the right place.

scottdraves:
shouldn't external system have its own location type instead of using location_http?
what happens when you try to save?

mforsyth:
@scottdraves I'd be open to making a separate location type.  I just figured it actually still was using http, so it would fall under http....  

We've always been using a separate save function to save notebook contents... I haven't actually gotten to the point of calling a builtin Beaker function to do the saving yet.

Does this existing http location type support saving back to the original url of the file?  Via a POST?

scottdraves:
it just crashes if you try that :) https://github.com/twosigma/beaker-notebook/issues/1523
thanks for prompting me to test it.

yes please make a location type, this is exactly what they are for.

mforsyth:
Sounds good @scottdraves

PR #1529, opened 2015-04-28
Convert from MathJax to Katex

buob:


scottdraves:
double dollar isn't being handled.  $$math$$ should be interpreted as a display equation, centered on its own line.  instead $$ is being interpreted as an empty equation and "math" is being interpreted as markdown.

scottdraves:
The performance is great though :)

scottdraves:
this is a huge relief :+1:

core/src/main/web/outputdisplay/bko-latex.js
```diff
@@ -24,8 +24,13 @@
     return {
       link: function(scope, element, attrs) {
         scope.$watch('model.getCellModel()', function(newValue) {
-          element.html("MATH_JAX_INLINE_BEGIN" + newValue + "MATH_JAX_INLINE_END");
-          MathJax.Hub.Queue(["Typeset", MathJax.Hub, element[0]]);
+          element.html("$" + newValue + "$");
```
sophiebits, line 27: By the way, this can just be

```
katex.render(newValue, element[0]);
```

since the math is already parsed out.

buob, line 27: That's true :+1:  Thanks!  Much nicer.

Issue #85, opened 2014-03-12
Support IPython Latex display


jchendy:
The following should all work in Python. None of them currently do.

```
from IPython.display import Latex
Latex("$x_i$")
```

```
from IPython.display import Math
Math(r'x_i')
```

```
%%latex
$x_i$
```

PR #1534, opened 2015-04-29
Katex Cleanups

buob:


buob:
This doesn't include embeddable_assets, launcher, or [iPython3 outputarea](https://github.com/twosigma/beaker-notebook/blob/master/plugin/ipythonPlugins/src/dist/vendor/ipython3/outputarea.js#L590-L602)

scottdraves:
embeddable_assets and launcher we'll deal with separately.

should include IPython v3 though. however, looks like IPy v2 is not working:

screen shot 2015-04-29 at 5 52 24 pm

ildipo:
Scott, are you sure this worked before? I think we have a github issue for 'support Latex output from python'

scottdraves:
you are right https://github.com/twosigma/beaker-notebook/issues/85

PR #1539, opened 2015-04-30
Bug Fixes

buob:


scottdraves:
what is the change to gulpfile?

buob:
It's just an added utility so devs can run `gulp` and it compiles and watches.

@samccone no, partial revert of larger feature.

scottdraves:
sorry i don't understand, gulp already runs to compile the files....
what does this change do?

buob:
It's just a shorthand.  Before this commit, the dev process was `gulp compile --debug` then `gulp watch`.  This just adds a "default task", which means a developer can only run `gulp` and it will do those things, because that's more or less the default manual task.

scottdraves:
thanks.  that's fine.

can you also fix the pointer changing to a hand?  i think it should only be a hand if you are over an element you can click on.

buob:
Updated, and added a couple other small fixes

buob:
Added one more commit for opening markdown links in new windows

PR #1541, opened 2015-05-01
Bug Fixes and UI Improvements


buob:


scottdraves:
i am not seeing an insert button before the 1st cell.

buob:
I changed it so that it's only visible on hover like the rest of the buttons (except when the notebook is empty).  I thought it looked a little cluttered or messy with the button always visible at the top, but I can turn it back on.  You can check it out in the chrome debugger by inspecting that first `bk-new-cell-menu` and putting `display: block` on it.

scottdraves:
i am hovering, but there's nothing there.

scottdraves:
my tree was goofed up, sorry.  looks good.

but i liked the big button for the end or when empty.

PR #1557, opened 2015-05-04
ci: cache installs

samccone:
Lets see how CI likes this :)

---

Caches: 
- Pip
- R
- npm

http://docs.travis-ci.com/user/caching/


samccone:
punting on this until I get a response from travis as to if they **do** support caching on OS projects.

PR #1563, opened 2015-05-06
style: Fix markdown output overlap with menu.

### LINKED ISSUE #1532: formatted markdown overlaps cell menu
When a markdown cell is formatted, if the text is long enough it overlaps with the cell menu buttons:
screen shot 2015-04-29 at 11 34 29 am

Would it be possible to flow the text around, ie break that first line early?  Is there a better way of solving this?

samccone:
The issue, was the fact that we have horizontal menu of cell actions
that are absolutely positioned on top of the output cell. Due to the
positioning the menu, the markdown cell output did not flow around the menu
element, thus causing a visual overlap. The CSS2 spec did introduce the
concept of a `::first-line` pseudo selector, however the rules that can be
applied are limited to a small subset of css rules [1]. None of which
can limit the width of the first line. The solution that I opted for
instead was to take advantage of a CSS3 pseudo selector `:before` to
insert a "fake" element that takes up the same size as the menu that is
out of the flow. into the output element to cause the output text
to reflow around

[1] https://developer.mozilla.org/en-US/docs/Web/CSS/::first-line

Fixes #1532

---
#### Visualizing the pseudo element

screen shot 2015-05-06 at 9 34 52 am
#### Applied

screen shot 2015-05-06 at 9 53 12 am

### DIFF
--- a/core/src/main/web/app/styles/cell-outputs.scss
+++ b/core/src/main/web/app/styles/cell-outputs.scss
@@ -56,6 +56,14 @@ bk-code-cell-output {
   .bk-section-title & {
     min-height: 1em;
   }
+
+  p:before {
+    content: ' ';
+    display: block;
+    float: right;
+    height: 10px;
+    width: 79px;
+  }
 }
 
 .modal.output-table-options {


scottdraves:
could be done by adding a blank element with float right to the DOM but :before is fine with me.

PR #1567, opened 2015-05-06
Allow users to enter an escaped $

### LINKED ISSUE #1428: need a way to put a dollar into markdown
$ is triggering LaTeX mode.  We need a way for users to mean it literally.

samccone:
Before this fix `\$` was shown as `\$` in the markdown output.
As part of the parsing step for marked we are now replacing instances of
`\$` with `$`. This step runs after the katex interpolation so there is
no need to be concerned with the `$` being interpolated into a new LateX
section.

Fixes #1428

### DIFF
--- a/test/tests/notebook.js
+++ b/test/tests/notebook.js
@@ -75,4 +75,18 @@ describe('notebook', function () {
     done();
   });
 
+  it('can handle escaping $ in markdown', function(done) {
+    beakerPO.newEmptyNotebook.click()
+    .then(function() {
+      return beakerPO.createMarkdownCell('hello world \\$');
+    })
+    .then(function() {
+      return beakerPO.readMarkdownCell();
+    }.bind(this))
+    .then(function(txt) {
+      expect(txt).toEqual('hello world $');
+      done();
+    });
+  });
+
 });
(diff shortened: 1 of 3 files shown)


test/tests/notebook.js
```diff
@@ -75,4 +75,18 @@ describe('notebook', function () {
...
+  it('can handle escaping $ in markdown', function(done) {
+    beakerPO.newEmptyNotebook.click()
+    .then(function() {
+      return beakerPO.createMarkdownCell('hello world \\$');
+    })
+    .then(function() {
+      return beakerPO.readMarkdownCell();
+    }.bind(this))
+    .then(function(txt) {
+      expect(txt).toMatch('hello world');
```
samccone: odd this should have failed...

samccone: toMatch hahah should be toEqual :)

scottdraves:
screen shot 2015-05-06 at 5 59 42 pm

produces

ParseError: KaTeX parse error: Expected 'EOF', got '\' at position 12: notmath\$
on the js console

samccone:
Looks like this is an issue with `\\$` specifically, ~~even on master `\\$` will break the current app~~ never mind it was me :) will fix and put a test in place

scottdraves:
yes pls

samccone:
moved to #1570

Issue #1568, opened 2015-05-07
copy from table to clipboard should use tab for separator


scottdraves:
not csv

PR #1572, opened 2015-05-07
Fix Issue #1568 by using tabs for clipboard

scottdraves:


ildipo:
What happens if you have a tab inside a cell?

PR #1575, opened 2015-05-07
Update publishAssets command dependency

buob:
And tweak semantics


scottdraves:
what's this PR about?

buob:
The publishAssets command to make a compiled version of beaker requires there to be an existing copy of our assets, so I just added a dependency to compile those assets before publishing.

Issue #1360, opened 2015-03-31
Remove "cell output"


scottdraves:
and move up the content of the output box.

buob:
Right now it shows "Error" if there's an error, otherwise "Cell Output".  It also shows the cell run time next to it.  Should we kill all of those?  The run time in small print might look weird by itself.

scottdraves:
Keep "Error" if there is one.
remove the run time.  i actually don't see that anymore IDK why but I know it used to be there.
any any case, it's in the cell menu so not important to keep here.

scottdraves:
"Output Hidden" should also be removed.

buob:
So should we just show an empty cell output if it's hidden?

scottdraves:
yup, just the controls on the right visible.

PR #1577, opened 2015-05-07
Ensure initial input output cells are = height

### LINKED ISSUE #1565: one line output is taller than one line input
screen shot 2015-05-06 at 11 52 17 am

i am not 100% sure this is a problem but i don't understand why a one-line output cell is taller than the one-line input side of the same cell.  above is what i get and if you copy and paste, i see a six pixel difference:

screen shot 2015-05-06 at 11 52 17 am

samccone:
#### Before

screen shot 2015-05-07 at 2 38 14 pm
screen shot 2015-05-07 at 2 38 18 pm
#### After

screen shot 2015-05-07 at 3 47 08 pm

screen shot 2015-05-07 at 3 47 13 pm

The root of this cause was a base line height rule being applied to
single outputs and to the entire body. Epic editor however overrides
line-height with a more specific CSS rule, thus the base body line-height
no longer applies, thus causing a difference in empty inline element
heights.

line-height initially changed in a50414819dbba547b4f8ef7ba639f95bc2bcb90d
For: https://github.com/twosigma/beaker-notebook/issues/1360

Fixes #1565

### DIFF
--- a/core/src/main/web/app/styles/cells.scss
+++ b/core/src/main/web/app/styles/cells.scss
@@ -115,7 +115,6 @@ bk-section-cell {
         padding: $grid-row-height $grid-gutter-width;
         position: relative;
         border-top: solid 1px #CCC;
-        min-height: #{$line-height-base}em;
         box-sizing: content-box;
 
         .out_error {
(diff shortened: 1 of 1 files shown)


scottdraves:
:zap:

Issue #1571, opened 2015-05-07
save as csv problems


scottdraves:
the dialog has an "open" button instead of save.
if the file exists, it silently fails to overwrite.

PR #1580, opened 2015-05-08
correctly label "save" button in file select dialog for "save as csv"…

scottdraves:
… feature in tables, and allow overwrite of an existing file.  Fixes Issue #1571.

### DIFF
--- a/core/src/main/web/outputdisplay/bko-tabledisplay/bko-tabledisplay.js
+++ b/core/src/main/web/outputdisplay/bko-tabledisplay/bko-tabledisplay.js
@@ -191,9 +191,9 @@
               return;
             if (n.indexOf(suffix,n.length-suffix.length) === -1)
               n = n + suffix;
-            // TODO check for error
-            return bkHelper.saveFile(n,out);
-          } , "Select name for CSV file to save", "csv");
+            // TODO check for error, prompt for overwrite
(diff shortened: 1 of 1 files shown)


core/src/main/web/outputdisplay/bko-tabledisplay/bko-tabledisplay.js
```diff
@@ -191,9 +191,9 @@
               return;
             if (n.indexOf(suffix,n.length-suffix.length) === -1)
               n = n + suffix;
-            // TODO check for error
-            return bkHelper.saveFile(n,out);
-          } , "Select name for CSV file to save", "csv");
+            // TODO check for error, prompt for overwrite
+            return bkHelper.saveFile(n, out, true);
```
samccone, line 195: Just curious as to why we default to overwrite now? I think this is the correct behavior however :+1:

scottdraves, line 195: best is to ask but that's more work than i have time for right now.
given the choice of silently overwrite or do nothing, overwrite is better.

samccone, line 195: I agree :+1:

PR #1583, opened 2015-05-08
Prepend promoted dropdowns to notebook instead of body

buob:
Promoted dropdowns are a component of notebooks,
so should be added and styled in relation to the notebook,
not the body.

### DIFF
--- a/core/src/main/web/app/styles/notebook.scss
+++ b/core/src/main/web/app/styles/notebook.scss
@@ -32,6 +32,10 @@
   }
 }
 
+bk-notebook {
+  position: relative;
+}
+
 bk-new-cell-menu {
   position: relative;
 
--- a/core/src/main/web/app/utils/basic/commonui.js
+++ b/core/src/main/web/app/utils/basic/commonui.js
@@ -73,7 +73,7 @@
   });
   module.directive('dropdownPromoted', function() {
     // Is your dropdown being covered by its ancestors siblings?
-    // Promote that shiz, and prepend it to the body so it doesn't
+    // Promote that shiz, and prepend it to the notebook so it doesn't
     // ever get bullied again.
     return {
       restrict: 'C',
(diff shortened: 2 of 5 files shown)


core/src/main/web/app/utils/basic/commonui.js
```diff
@@ -83,13 +83,15 @@
 
         var showDropdown = function() {
           var togglePosition = toggle.offset();
+          var $notebook = $('bk-notebook');
```
samccone: is it possible we could have multiple notebooks on a page?

buob: Yeah, I could find the `closest` bk-notebook from the element.  Currently we don't though, and it would cost more to do that.

core/src/main/web/app/utils/basic/commonui.js
```diff
@@ -83,13 +83,15 @@
...
+          var notebookPosition = $notebook.offset();
```
samccone: this method is now going to cause two layout calcs .. causing some perf hits :| perhaps we could use RAQ and do the majority of the reflow work on the next available paint frame (this change would require some refactoring)

samccone:
on an unrelated note this directive can possibly leak
https://github.com/twosigma/beaker-notebook/pull/1583/files#diff-c59bab0b64fc0079d2dee2e7f8d9f92eR98

if you are in this file already mind stacking another commit to resolve this?

core/src/main/web/app/styles/notebook.scss
```diff
@@ -32,6 +32,10 @@
   }
 }
 
+bk-notebook {
+  position: relative;
```
samccone, line 36: what happens without position relative?

core/src/main/web/app/utils/basic/commonui.js
buob: Yeah, we could do that.  I'm not sure if that'd be worth the complexity in this case though, as it's on user interaction.

buob:
Yeah I could fix that

core/src/main/web/app/utils/basic/commonui.js
samccone: I guess, global non scoped lookups for directives are non idea, I wonder if there is a better solution than just grabbing the global bk-notebook DOM node?

core/src/main/web/app/styles/notebook.scss
buob, line 36: the dropdown isn't positioned to the bk-notebook, it's positioned to the closest positioned parent

core/src/main/web/app/utils/basic/commonui.js
buob: That's fair.  I wouldn't mind looking for the closest one, as again, it's only on user interaction.

core/src/main/web/app/styles/notebook.scss
samccone, line 36: 👍

core/src/main/web/app/utils/basic/commonui.js
samccone: you could walk the parent scope up until you find the notebook directive and get the element that way... that might be nice

samccone: or... pass it in :) via an isolated scope

core/src/main/web/app/utils/basic/commonui.js
samccone: Yup on user interaction which means other things might also be happening, sometimes in this app as you have seen it will randomly feel sluggish on user interactions. Ideally we would guard against these perf pitfalls by coding in these solutions to our directives. 

But yeah it is only "2" recalcs of the entire document layout, which in isolation with nothing else happening would most likely not hurt FPS in a noticeable way. Given how complex this app is, I would not bet on that fact.

samccone:
this looks good to go @scottdraves tested locally

PR #1584, opened 2015-05-08
Fix vertical jump on single line headlines.

samccone:
By enforcing that the content is aligned to the top of the inline-block
container there is no shift between the output mode and edit mode for
single lines of content.


scottdraves:
what problem does this fix?

samccone:
Let me make a gif for you to see

samccone:
Here is the vertical jump.. this code makes it not happen for the reasons stated in the commit message 

out

scottdraves:
:+1:

PR #1589, opened 2015-05-11
Initial code-mirror update

samccone:
This is the initial work to get us inline with the latest release of code mirror. 

It seems like everything is works just fine. :tea: 

---

@scottdraves If you would not mind giving the code editor this a run through I tested the functionality that I know of, but I am sure there are more things that I do not know of that you might be able to catch.

### DIFF
--- a/core/src/main/web/app/styles/vendor_overrides.scss
+++ b/core/src/main/web/app/styles/vendor_overrides.scss
@@ -17,7 +17,6 @@
 .CodeMirror {
   clear: both;
   height: auto;
-  overflow: visible;
 }
 
 .CodeMirror-gutters {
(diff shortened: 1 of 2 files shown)


core/src/main/web/app/styles/vendor_overrides.scss
```diff
@@ -17,7 +17,6 @@
 .CodeMirror {
   clear: both;
   height: auto;
-  overflow: visible;
```
scottdraves, line 20: did you test this on windows?

samccone:
screen shot 2015-05-11 at 1 57 26 pm

scottdraves:
thanks

PR #1593, opened 2015-05-11
Make buildembed run copyNginx

mforsyth:
(build and builddebug already do).


samccone:
what happened when this was not run?

PR #1595, opened 2015-05-12
style fixups

samccone:


core/src/main/web/app/utils/basic/commonui.js
```diff
@@ -1,14 +1,14 @@
 /*
  *  Copyright 2014 TWO SIGMA OPEN SOURCE, LLC
  *
- *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  Licensed under the Apache License, Version 2.0 (the 'License');
```
scottdraves: please do not change the license banner

samccone: whoops 1 moment :)

PR #1603, opened 2015-05-13
Allow for multi-user environment instance id's to be UUID's.

mforsyth:
(Previous version supported numeric id's instead).


samccone:
:+1:

Issue #1546, opened 2015-05-04
Performance Investigation Tickets


samccone:
Here are the current tickets that we have filed:
#178 -- initial notebook load time
#838 -- write performance tests
#1530 -- watchers

---

An initial step might be to manually generate some performance numbers so that we have a base to go from.
What do you think @scottdraves ?

scottdraves:
i believe we already have a test for 178, though i am not sure it is right.
we need to have some large notebooks for 838, i can provide you with links.
adding a protractor tests that do these and just report the number somewhere on the console would be great.

scottdraves:
http://nbviewer.ipython.org/url/norvig.com/ipython/TSPv3.ipynb
is a good big one.  also very instructive to see how fast it loads in ipython vs beaker.
it took me about 20s to load in beaker vs 5s in ipython, timed informally.
plus, it opens almost instantly on their sharing server.

samccone:
# Performance Investigation into slow initial boot. 

_(5/5/15)_

The first several seconds when visiting a notebook from a cold start seems to take a very long time. I broke the "slowness" into two discrete categories of issues. One is more the "why" someone things a page loads slowly and the other is the actual performance issues that cause certain aspects of the page to load sluggishly. 

## Perceived slowness

When A user visits a new notebook they are presented with a screen that looks like an empty notebook 

![screen shot 2015-05-05 at 3 03 51 pm](https://cloud.githubusercontent.com/assets/883126/7480904/0ab5f81a-f33a-11e4-83c1-6257bd79f914.png)

The screen then sits in that state until the entire notebook populates in (on large notebooks this can take multiple seconds) – most of this time is actually the load time of the notebook from the sever .. not the render time.

#### Solution

This is more a a UX problem, and should be quite easy to solve with a loading indicator and or spinner, to give the user some feedback that something is happening. 

## Perf Bottlenecks

On large notebooks the perf issues become quite obvious, due to the sheer number of elements and directives that are rendered into the DOM. (on larger notebooks this can be upwards of 100+ code editors)

### CodeMirror

The biggest offender by far is the code editor directives and more specifically codemirror performing multiple expensive root layout reflows when generating a single codemirror instance 

![screen shot 2015-05-05 at 10 35 40 am](https://cloud.githubusercontent.com/assets/883126/7481546/a874c7da-f33e-11e4-93b9-89717c621fc4.png)

These four calls are taking up around 20% of the total JS execution time for rendering one large notebook **yikes**

[...]

Issue #178, opened 2014-04-23
Improve notebook loading speed


jchendy:
Try a large notebook such as https://raw.githubusercontent.com/gumption/Python_for_Data_Science/master/Python_for_Data_Science_all.ipynb

In Beaker, it takes ~45 seconds before anything appears and another 45 seconds before  the loading is fully complete.

For comparison, IPython loads the whole thing in ~12 seconds.

scottdraves:
not there yet

samccone:
@jchendy Would love to to have some feedback on how this "feels" now with the new fix that just landed

scottdraves:
we made some improvements on this front, need to keep working on it.

scottdraves:
dup of https://github.com/twosigma/beaker-notebook/issues/2336

PR #1615, opened 2015-05-14
Lazily instantiate codemirror instances

samccone:
By taking advantage of the fact that a user can not interact
with a codemirror instance until the editor is in the viewport, we can be
smart and only set the codemirror editors up once a user scrolls them into view.

For small notebooks this change has a limited impact, however for large
notebooks with 100s of code cells this change results in a significant
performance boost to the application, especially in the initial time to
render metric for notebooks.

---

**Tested With http://nbviewer.ipython.org/url/norvig.com/ipython/TSPv3.ipynb**
##### Initial setup work
- Download the ipython notebook
- Convert to a beaker notebook
- Save to filesystem
#### Testing Process
- Start the profiler
- Open the bkr notebook 
- Once the initial map paints, stop the profiler

screen shot 2015-05-14 at 2 31 40 pm

screen shot 2015-05-14 at 2 31 57 pm

_[Source analysis](https://docs.google.com/spreadsheets/d/1QQ9uWVikhvgqPdwvO03B5WnJkac3507uiBWPAwBDpl4/edit#gid=0)_

---

**Ref** #1546
**Fixes** #178

### DIFF
--- a/core/bower.json
+++ b/core/bower.json
@@ -23,6 +23,7 @@
     "marked": "~0.3.3",
     "acorn": "0.3.1",
     "katex-build": "~0.3.0",
-    "scrollin": "~0.1.4"
+    "scrollin": "~0.1.4",
+    "rAF": "https://gist.githubusercontent.com/paulirish/1579671/raw/3d42a3a76ed09890434a07be2212f376959e616f/rAF.js"
   }
 }
--- a/core/bower.json
+++ b/core/bower.json
@@ -22,6 +22,7 @@
     "big.js": "~2.5.1",
     "marked": "~0.3.3",
     "acorn": "0.3.1",
-    "katex-build": "~0.3.0"
+    "katex-build": "~0.3.0",
+    "scrollin": "~0.1.4"
   }
 }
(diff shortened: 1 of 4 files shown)


ildipo:
The power of JS libraries....
It looks nice

scottdraves:
i found i could get a code cell that had not been instantiated on screen
screen shot 2015-05-14 at 3 31 30 pm

samccone:
I did some follow up time measurements on my computer (these numbers will be highly variable depending on the system)

```
New: (lazy codecell instantiation)
16.42 seconds

Old: (eager codecell instantiation) 
26.5 seconds
```

Tested With http://nbviewer.ipython.org/url/norvig.com/ipython/TSPv3.ipynb

Time from clicking on the recent notebook item on the dashboard until being able to interact with the first codecell

scottdraves:
Thanks, that's a fantastic improvement in performance.

However, I can see the code cells instantiating as I scroll, which is pretty annoying.

Instead of starting them out as unstyled text, can you initialize them to the right font size
with the box around them, but otherwise static?  That should be really fast, and should look
exactly like they do after they become real CodeMirrors, except for the text changing color for
the highlighting.

samccone:
updated

scottdraves:
:dolphin: :dolphin:

samccone:
:dancers: :dancers: :dancers: :dancers:

PR #1628, opened 2015-05-18
Fix for issue 1622 and refactoring R serialization to reduce overhead

ildipo:


samccone:
@ildipo was there a decent test that we could put in place for this?

scottdraves:
confirm that the code on the original issue produced a plot output type.

PR #1629, opened 2015-05-18
Add public getStatus method to bkHelper

buob:
Just adds a little utility method to have access to the status message outside of the app.

### DIFF
--- a/core/src/main/web/app/helpers/helper.js
+++ b/core/src/main/web/app/helpers/helper.js
@@ -199,6 +199,13 @@
           console.error("Current app doesn't support updateStatus");
         }
       },
+      getStatus: function() {
+        if (getCurrentApp().getStatus) {
+          return getCurrentApp().getStatus();
+        } else {
+          console.error("Current app doesn't support getStatus");
+        }
+      },
       clearStatus: function(message, nodigest) {
         if (getCurrentApp().clearStatus) {
           return getCurrentApp().clearStatus(message, nodigest);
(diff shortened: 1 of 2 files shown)


core/src/main/web/app/helpers/helper.js
```diff
@@ -199,6 +199,13 @@
           console.error("Current app doesn't support updateStatus");
         }
       },
+      getStatus: function() {
+        if (getCurrentApp().getStatus) {
```
samccone, line 203: `_.isFunction` I think would be better here right?

core/src/main/web/app/mainapp/mainapp.js
```diff
@@ -70,6 +70,9 @@
           if (!($scope.$$phase || $scope.$root.$$phase))
             $scope.$digest();
         };
+        var getLoadingStatusMessage = function() {
```
samccone, line 73: why are you using var vs function getLoadingStatusMessage ?

core/src/main/web/app/helpers/helper.js
buob, line 203: It would certainly be more specific, but in my (obviously humble) opinion it is unnecessary.  It would be useful if there were some case where `getStatus` was an existing attribute but wasn't a function (which would be a terrible thing).  If that isn't the case, isn't it easier to just test the falsiness of the attribute?

core/src/main/web/app/mainapp/mainapp.js
buob, line 73: It's in our style guide - http://google-styleguide.googlecode.com/svn/trunk/javascriptguide.xml?showone=Function_Declarations_Within_Blocks#Function_Declarations_Within_Blocks

samccone, line 73: Inside of blocks yep, inside functions should be be fine though

PR #1654, opened 2015-05-20
Remove unneeded extra right padding on cells.

### LINKED ISSUE #1643: markdown button flow-around triggers too often
The way markdown formats it should flow around the cell menu in the upper-right.  It does, but it looks like more than just the 1st line is truncated.  It looks like it generates an extra block of negative space for every paragraph in fact.  Just looking at the first cell of the main tutorial you can see the 2nd paragraph 1st line wraps too soon, after "variety of" even though "languages" would fit:
screen shot 2015-05-19 at 5 42 34 pm
If you select the text you can see unwanted negative space blocking it there:
screen shot 2015-05-19 at 5 44 02 pm 2

samccone:
Fixes #1643
##### Before

screen shot 2015-05-21 at 10 05 09 am
##### After

screen shot 2015-05-21 at 10 05 05 am

### DIFF
--- a/core/src/main/web/app/styles/cell-outputs.scss
+++ b/core/src/main/web/app/styles/cell-outputs.scss
@@ -51,7 +51,7 @@ bk-code-cell-output {
     display: block;
     float: right;
     height: 10px;
-    width: 79px;
+    width: 117px;
   }
 }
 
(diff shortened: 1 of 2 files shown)


samccone:
alright I think this should be ok now @scottdraves

PR #1656, opened 2015-05-21
Mf/new location type

mforsyth:
This adds a new location type, "ajax".

Ajax is different from the existing HTTP location type in that notebook contents will be loaded directly from the client using an Ajax request to the source url, rather than the request being proxied through the Beaker server.   (this also means that the ajax location must be either on the same host & port as the Beaker interface, or made available via CORS settings).

This location type also supports saving via a PUT request;  an Ajax notebook location has both a "source path" and a "destination path", for loading and saving, respectively.


ildipo:
looks ok to me

PR #1660, opened 2015-05-21
Make AJAX location type save data in JSON format.

mforsyth:
This is going to work better for servers that are strict about the
format and encoding of the request.


mforsyth:
Oops... please wait until I have pushed the new embedded assets....

mforsyth:
Ready.

PR #1677, opened 2015-05-28
Ildipo/jvm serialization extension

ildipo:


plugin/jvm/src/main/java/com/twosigma/beaker/jvm/serialization/ObjectSerializer.java
```diff
@@ -0,0 +1,11 @@
+package com.twosigma.beaker.jvm.serialization;
```
scottdraves, line 16: needs banner

plugin/jvm/src/main/java/com/twosigma/beaker/jvm/serialization/ResultsDeserializer.java
```diff
@@ -1,40 +1,22 @@
 package com.twosigma.beaker.jvm.serialization;
```
scottdraves, line 16: needs banner

plugin/scala/src/main/scala/com/twosigma/beaker/scala/Main.java
```diff
@@ -62,6 +63,8 @@ public static void main(String[] args) throws Exception {
 
     Server server = injector.getInstance(Server.class);
     server.start();
+    System.out.println("waiting");
```
scottdraves: remove

PR #1683, opened 2015-05-29
Preserve document title and append notebook name

buob:
Hey @scottdraves, this change is more for embeddable beaker, but wasn't sure if you would be interested in having it in desktop beaker as well.  Right now if you switch from beaker app to a notebook, the document title will change from `Beaker` to `New Notebook`.  This change makes it so it changes from `Beaker` to `Beaker - New Notebook`.

I can look into another approach for embeddable if this isn't a change you'd be interested in, but thought I would present it as an option.  :dancers:


buob:
Testing this in another scenario didn't work with this implementation, so closing for now, but let me know if you'd be interested in the functional change.

scottdraves:
the user can often only see the beginning of a window title (when they have many tabs open).
so i don't think this makes sense for any version.
please skip it completely.

PR #1686, opened 2015-06-01
Fix focusing of new cells.

### LINKED ISSUE #1680: when you add a new cell with the keyboard shortcut
it should get the focus so you can just start typing into it.
this is a regression.
the shortcut is control-shift-a

samccone:
This bug was introduced in 632880f4bce4af8cb14817e1fb68d632fb193525.

The reason why this happened was because since we lazily instantiate
each codemirror instance we can not focus it right away (since it is
not yet created).

By setting some intermediate state on the codecell scope, we can then
hook into it when the cell enters the viewport and is created.

---

Fixes #1680


scottdraves:
travis problem is real?

PR #1694, opened 2015-06-02
Only backup session if something has changed.

### LINKED ISSUE #1691: do not backup if there were no changes
even with no activity, backups happen
screen shot 2015-06-01 at 7 16 55 pm

samccone:
Fixes #1691

### DIFF
--- a/core/src/main/web/app/mainapp/services/sessionmanager.js
+++ b/core/src/main/web/app/mainapp/services/sessionmanager.js
@@ -415,6 +415,7 @@
     var _format = null;
     var _sessionId = null;
     var _edited = false;
+    var _needsBackup = false;
 
     var BeakerObject = function(nbmodel) {
       this.knownBeakerVars = { };
@@ -832,7 +833,7 @@
         _format = format;
         _notebookUri.set(notebookUri);
         _notebookModel.set(notebookModel);
-        _edited = !!edited;
+        this.setNotebookModelEdited(!!edited);
         _sessionId = sessionId;
 
         bkNotebookNamespaceModelManager.init(sessionId, notebookModel);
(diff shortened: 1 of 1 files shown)


scottdraves:
i don't like the idea of keeping around a whole extra copy of the notebook.
it would be better to either use a hash of it, or beaker is keeping track of the dirty state anyway, just use that.

samccone:
~~found a bug in it one moment @scottdraves~~

resolved

scottdraves:
This needs its own "edited" flag, not the one based on saving.
but that flag should be set by the same function that is called from many places in the app when the notebook is made dirty by editing.

samccone:
ah I understand, sorry misstep on my part.

core/src/main/web/app/mainapp/services/sessionmanager.js
```diff
@@ -832,7 +833,7 @@
         _format = format;
         _notebookUri.set(notebookUri);
         _notebookModel.set(notebookModel);
-        _edited = !!edited;
+        this.setNotebookModelEdited(!!edited);
```
ildipo, line 836: Semantic here is different - was this intentional?

core/src/main/web/app/mainapp/services/sessionmanager.js
```diff
@@ -892,8 +893,11 @@
         }
       },
       backup: function() {
-        if (_sessionId && !_notebookModel.isEmpty()) {
-          return bkSession.backup(_sessionId, generateBackupData());
+        if (_sessionId && !_notebookModel.isEmpty() && _needsBackup) {
+          return bkSession.backup(_sessionId, generateBackupData())
+          .then(function() {
+            _needsBackup = false;
```
ildipo: Here I think you should clear _needsBackup before executing bkSession.backup() - what happens if I'm editing the notebook and I put edits during the time it takes for the request to go out to when the promise is fulfilled?

core/src/main/web/app/mainapp/services/sessionmanager.js
samccone, line 836: can you expand on what you mean?

core/src/main/web/app/mainapp/services/sessionmanager.js
samccone: k

Issue #1689, opened 2015-06-01
command-click on logo broken


scottdraves:
it opens up a page like http://127.0.0.1:8801/2fa82e1/beaker/
with a directory listing and the old favicon.

there are 3 things to fix here:
1) remove the old favicon
2) do not serve directory listings
3) command-lick on logo should open control panel in another tab, just like the "New Beaker Window" menu command.

PR #1698, opened 2015-06-03
Fixed issue #1689: Used correct target URL for command-click on brand…

DiegoAlbertoTorres:
… logo. Turned off 'autoindex' (directory listing) on nginx at (urlhash)s/beaker. Not sure why this was necessary, since 'autoindex' was already off at the root of the config file.

### DIFF
--- a/core/nginx/nginx.conf.template
+++ b/core/nginx/nginx.conf.template
@@ -74,6 +74,7 @@ http {
       proxy_set_header Authorization "Basic %(auth)s";
     }
     location = /%(urlhash)s/beaker/ {
+      autoindex off;
       return 301 %(start_page)s;
     }
 
--- a/core/nginx/nginx.conf.template
+++ b/core/nginx/nginx.conf.template
@@ -74,7 +74,6 @@ http {
       proxy_set_header Authorization "Basic %(auth)s";
     }
     location = /%(urlhash)s/beaker/ {
-      autoindex off;
       return 301 %(start_page)s;
     }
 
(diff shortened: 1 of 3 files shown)


core/nginx/nginx.conf.template
```diff
@@ -74,6 +74,7 @@ http {
       proxy_set_header Authorization "Basic %(auth)s";
     }
     location = /%(urlhash)s/beaker/ {
+      autoindex off;
```
samccone: was this needed to fix the click issue or is this just a separate fix?

DiegoAlbertoTorres: Separate fix I guess, but Scott noticed directory listings were still accessible thanks to the click issue.

samccone: Word, sometimes it is nice to split these changes up into separate commits. Take for instance this case.

You open a PR with one commit that does 3 things. 

1 of the things causes a bug or regression.
Since it is all in one commit we can not git revert <SHA> without losing the other features.

If it is a single commit per "feature or change" it makes it easy to back out mistakes.

DiegoAlbertoTorres: You're right. I'll keep it in mind!

core/src/main/web/app/controlpanel/controlpanel-directive.js
```diff
@@ -36,7 +36,7 @@
 
         $scope.gotoControlPanel = function(event) {
           if (bkUtils.isMiddleClick(event)) {
-            window.open("./");
+            window.open("../../");
```
ildipo: Can we use the bkHelper.getServerUrl() here?

DiegoAlbertoTorres: bkHelper.serverUrl(x) gives a URL of the form 'ip:port/hash/x', but the click is supposed to redirect to the root ('ip:port/'). Is there another function that returns the base url without the hash?

scottdraves:
:+1:

PR #1699, opened 2015-06-03
Add notebook-uri and session-id directive options

buob:
This allows embedding applications to control the behavior of the beaker-notebook directive without needing to rely on the state of ngRoute.  This is especially helpful for embedding applications that don’t use ngRoute at all.


core/src/main/web/app/mainapp/mainapp.js
```diff
@@ -59,7 +59,13 @@
     return {
       restrict: 'E',
       template: JST["template/mainapp/mainapp"](),
-      scope: {},
+      scope: {
+        notebookUri: "@",
```
samccone: quote style `'` plz

PR #1722, opened 2015-06-08
Min-height avoids scrolling. Scrolling happens because cell text is t…

DiegoAlbertoTorres:
…oo short

### DIFF
--- a/core/src/main/web/app/styles/cells.scss
+++ b/core/src/main/web/app/styles/cells.scss
@@ -23,12 +23,6 @@
     padding-top: $grid-row-height;
 }
 
-.CodeMirror-scroll {
-    overflow-y: hidden !important;
-    overflow-x: auto !important;
-    height: auto;
-}
-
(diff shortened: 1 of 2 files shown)


samccone:
Can you post before and after screenshots? and how this fixes things in a non brittle way?

scottdraves:
this adds too much space below the code box.

core/src/main/web/app/styles/cells.scss
```diff
@@ -23,6 +23,12 @@
     padding-top: $grid-row-height;
 }
 
+.CodeMirror-scroll {
+    overflow-y: hidden !important;
```
samccone: can we put this into vendor overrides and also drop !important

samccone:
screen shot 2015-06-11 at 9 38 31 am
screen shot 2015-06-11 at 9 38 14 am

:boat: :+1:

PR #1732, opened 2015-06-10
Fix casting issue of scss units.

samccone:
On windows, the scss compiler fails due to it being unable to handle
converting `px/px` units to `em`.

ref sass/sass#1743


samccone:
going to close this for now since it is not the real "fix"

PR #1760, opened 2015-06-16
Update notebook name when saved as.

### LINKED ISSUE #1737: New notebooks not added to 'Open notebooks'
When creating an empty notebook, adding stuff, saving, then returning to home, the recently saved notebook (with whatever name it was saved with) should show up in 'Open notebooks' at Beaker's landing page. Instead, only an 'empty notebook' is showing up right now.

samccone:
Fixes #1737

### DIFF
--- a/core/src/main/web/app/mainapp/mainapp.js
+++ b/core/src/main/web/app/mainapp/mainapp.js
@@ -527,9 +527,20 @@
           var saveStart = function() {
             showLoadingStatusMessage("Saving");
           };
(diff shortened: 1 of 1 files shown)


scottdraves:
seems to have two problems:
1) this fix only applies to the "save as" menu item, not to the "save" item.
2) it does fix the name that shows up in the list of open notebooks, but when i click on it, i get the empty notebook, not the current content.

samccone:
ok let me address it

PR #1761, opened 2015-06-16
Remove dangerous tags from html

ildipo:


core/src/main/web/app/mainapp/components/notebook/outputdisplay/outputdisplayfactory-service.js
```diff
@@ -82,12 +82,27 @@
           });
         },
         link: function(scope, element, attrs) {
+          var tagstofilter = ['applet', 'base', 'basefont', 'body', 'frame', 'frameset', 'head', 'html',
```
samccone, line 85: Is there no lib to do this that we can use? This seems like something we do not want to roll on our own.

ildipo, line 85: I did search for one when I did it for our sharing server and found nothing simple. There are some full sanitizers but these are very complex and work by whitelist - never updated since years so, for example, no svg.

samccone, line 85: well, this just seems so simple to bypass that I am not even sure why we want it.

PR #1770, opened 2015-06-17
Inform the user about KaTeX errors and its limitations.

ildipo:


scottdraves:
needs to also work for LaTeX in markdown, not just LaTeX cells

core/src/main/web/outputdisplay/bko-latex.js
```diff
@@ -24,7 +24,11 @@
     return {
       link: function(scope, element, attrs) {
         scope.$watch('model.getCellModel()', function(newValue) {
-          katex.render(newValue, element[0]);
+          try {
```
samccone, line 27: something to keep in mind as soon as we use a try catch the entire code block gets deoptimized in v8. Since this is inside a `$watch` this might actually add up perf wise sine it is going to get hit during every digest loop.

This might be worth investigating if there is another way other than a try catch.

ildipo, line 27: this should trigger only when the value changes anyway and the block contains only an invocation to another function which is optimized, so the performance hit should not be so bad

PR #1781, opened 2015-06-18
Rework progress-update template

ildipo:


scottdraves:
this is showing me the progress % bar even when it's not used

PR #1782, opened 2015-06-18
Fix delayed spinner hide in advanced mode.

### LINKED ISSUE #1771: advanced mode spinner shows when cells are inserted
if i have a notebook with backends loaded, and i insert a cell, then i see the spinner for about 0.75s

samccone:
Angular prevents hiding an element while an animation is happening on
the element. In this case we were trying to hide an element that was
"spinning" and since it had a 2 second animation duration, the ng-if would only
act on the element after the animation cycle had completed.

By putting a containing div on the ng-if directive we work around this
limitation. In someways this is a hack.

Fixes #1771


samccone:
https://github.com/angular/angular.js/issues/8033

scottdraves:
big progress :)
but in nonadvanced mode, on markdown cells, the spinner always shows.
also, the spinners show when the cell first appears on the page as i scroll down.

PR #1838, opened 2015-06-24
Fix table refresh when inside a result set

ildipo:


ildipo:
Updated to include also tables inside OputputContainer, tables inside collapsed section and bug #1839

scottdraves:
still seeing the bugs

ildipo:
pull? clear cache? works for me

PR #1868, opened 2015-06-30
Sjs/cell viewport enter fixes

### LINKED ISSUE #1796: uninstantiated code cells visible by deleting or running cells
we lazily instantiate codemirrors as they scroll onto the page,
but there are some ways to make them visible that are not detected.
for example, you if you delete cells, then the cells that are below the screen become visible
but the CMs are not instantiated.  another possibility is if you have a cell with a large output.
when you run it, the output is deleted, and a code cell below comes on screen, but is not instantiated.

there are two things to fix here.  one is to detect when cms are visible in these situations.
the other is to make the uninstantiated version of the cm more like the instantiated one.
as you can see in this screenshot, it has a scrollbar and much of the text is not visible.

screen shot 2015-06-20 at 5 19 53 pm

### LINKED ISSUE #1850: switching from normal mode to advance mode does not instantiate CodeMirror
If switching from normal to advanced mode shows more code cells the editor is not instantiated until the user scrolls the page

samccone:
Fixes #1850
Fixes #1796

---

Fix uninstantiated codecell visuals
##### Normal mode:

screen shot 2015-06-30 at 11 03 48 am
##### Advanced mode:

screen shot 2015-06-30 at 10 47 03 am

### DIFF
--- a/core/bower.json
+++ b/core/bower.json
@@ -23,7 +23,7 @@
     "marked": "~0.3.3",
     "acorn": "0.3.1",
     "katex-build": "~0.3.0",
-    "scrollin": "~0.1.4",
+    "scrollin": "~0.2.2",
     "rAF": "https://gist.githubusercontent.com/paulirish/1579671/raw/3d42a3a76ed09890434a07be2212f376959e616f/rAF.js"
   }
 }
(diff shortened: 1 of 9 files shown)


scottdraves:
looks good but has merge conflict

samccone:
boom! so excited to have this in now :)

PR #1870, opened 2015-06-30
Fix uninstantiated codecell visuals

samccone:


scottdraves:
thanks that's better but i am only seeing 2 lines of text and a scrollbar.
if it can show all the text without scrolling then this will be good.

samccone:
I agree that would be very nice, the textarea node does not have the ability to grow in height dynamically without using javascript to calculate the height and dynamically resize it via JS.

So that being said, to implement dynamic heights, we would need some JS to be bound and run on uninstantiated codecells on load, which will come with a perf hit.

---

refs: 

http://www.jacklmoore.com/autosize/
https://github.com/javierjulio/textarea-autosize

http://stackoverflow.com/questions/15844868/can-i-make-a-textarea-grow-with-text-to-a-max-height

thoughts?

scottdraves:
how about using a div until it's instantiated and then switching it to a textarea just before instantiation?
that might be even faster since the div doesn't support editing at all, and so might be simpler and faster.

samccone:
sure thing!

samccone:
:shaved_ice: woot multiline support :white_check_mark: 

great idea @scottdraves !!

PR #1871, opened 2015-06-30
remove zip!

samccone:
This change should prevent the entire class of issues related to #1746.
From this point forward when someone adds, removes, or updates a
dependency in the repo it will show up in the diff, making it easy to
spot when the incorrect files get added or deleted.

The initial motivation for the zip file was because there was so
many files from mathjax, but after moving to katex and updating our
gitignore to not track unused vendor files, the total number of files that have been added to the repo is
644.

---

Committing the vendor files added 2094650 or 2 megs to the repo size, however unlike before any change to vendors files will not require a new 8meg blob to be added each time :)


.gitignore
```diff
@@ -24,6 +24,32 @@ manifest.mf
 npm-debug.log
 core/src/main/web/app/templates.js
 core/config/hash
-core/src/main/web/vendor/
 core/src/main/web/index.html
 /bin/
+
+core/src/main/web/vendor/bower_components/jquery/test
```
scottdraves, line 30: core/src/main/web/vendor/bower_components/

shouldn't just that one take care of all of them?

.gitignore
```diff
@@ -24,6 +24,32 @@ manifest.mf
...
+
+core/src/main/web/vendor/bower_components/font-awesome/src
+core/src/main/web/vendor/bower_components/font-awesome/less
+core/src/main/web/vendor/bower_components/font-awesome/scss
+
+core/src/main/web/vendor/bower_components/acorn/test
+
+core/src/main/web/vendor/bower_components/underscore.string/test
+
+core/src/main/web/vendor/DataTables-1.10.7/extensions/ColVis
```
scottdraves, line 46: why is DataTables-\* here?

.gitignore
samccone, line 30: the reason why I did not gitignore all of `core/src/main/web/vendor/bower_components/`
is because I want to commit (in source control) the files that we depend on. If we did ignore all of the `bower_components` then to run the app we would always have to `bower i` which would hit the network, thus negating the benefits of vendoring the files.

.gitignore
samccone, line 46: I am ignoring `extensions/ColVis` and several other folders within DataTables which we are not using anything out of.

DataTables ships with tons and tons of files, I ignore the ones that we are not using to prevent unneeded bloat to the repo.

PR #1886, opened 2015-07-02
style cleanups as I went through some files

samccone:


samccone:
investigating the failing tests

samccone:
let me know if there is anything else you want me to look at here

PR #1917, opened 2015-07-11
moment was missing

scottdraves:
how is this working on travis?


scottdraves:
ok we've got this: https://github.com/twosigma/beaker-notebook/tree/master/core/src/main/web/vendor/moment

scottdraves:
but also: https://github.com/twosigma/beaker-notebook/blob/master/core/bower.json#L20-L21
and
https://github.com/twosigma/beaker-notebook/blob/master/core/src/main/web/app/template/index_template.html#L99-L100

samccone:
I think we should be able to ignore a fair amount of these files via the gitignore

`core/src/main/web/vendor/bower_components/moment/src/`

scottdraves:
you are right thanks, will update.

scottdraves:
ready to merge

core/src/main/web/vendor/bower_components/moment-timezone/builds/moment-timezone-with-data-2010-2020.js
```diff
@@ -0,0 +1,1015 @@
+//! moment-timezone.js
```
samccone, line 1: I do not think we use this file

samccone:
I think we should try to be pretty strict about the vendor files we add in (not committing files that we do not use), just to remove the noise for these commits and the file count overhead that caused issues with mathjax

samccone:
@scottdraves per how is this working on travis, that is a good question, I think that there were 0 tests testing moment or the date functionality.

Can you think of a nice reduced test case for me to codify?

PR #1928, opened 2015-07-13
Diego/better gradle

DiegoAlbertoTorres:
Many improvements to the gradle script, particularly on Windows.


launcher/electron/README.md
```diff
@@ -54,15 +57,22 @@ If you do not have Electron's prebuilt binaries installed, simply do
 gradle installElectron
 ```
 
+If you are on Windows you will have to add the prebuilt binaries to your path (they should
+be in your global node_modules folder if you called `gradle installElectron`)
+
 To produce an executable you can run (an app bundle on OSX, a folder with an executable on Windows and Linux), do:
 ```sh
-gradle makeBundle
+gradle makeBundle -P version=0.29.2 -P arch=x64
```
scottdraves, line 65: bake arch in

scottdraves, line 65: make the current good version the default

DiegoAlbertoTorres, line 65: Done

PR #1934, opened 2015-07-14
Diego/1916

DiegoAlbertoTorres:


samccone:
if you wanted to you could use an interactive rebase to squash this into a single commit to keep the git log clean

PR #1940, opened 2015-07-15
Add sign in and publish functionality

grin:
https://www.pivotaltracker.com/story/show/98653246
https://www.pivotaltracker.com/story/show/88906262

### DIFF
--- a/core/src/main/web/app/mainapp/components/publication/publish.jst.html
+++ b/core/src/main/web/app/mainapp/components/publication/publish.jst.html
@@ -25,7 +25,7 @@
       </form>
       <div class="sign-up">
         No account?
-        <a target="_blank" href="https://pub.beakernotebook.com/#/sign_up">Create New Account</a>
+        <a target="_blank" href="http://pub.beakernotebook.com/#/sign_up">Create New Account</a>
       </div>
     </div>
     <div class="publish" ng-if="isSignedIn()">
--- a/core/src/main/web/app/mainapp/components/publication/publish.jst.html
+++ b/core/src/main/web/app/mainapp/components/publication/publish.jst.html
@@ -25,7 +25,7 @@
       </form>
       <div class="sign-up">
         No account?
-        <a target="_blank" href="http://pub.beakernotebook.com/#/sign_up">Create New Account</a>
+        <a target="_blank" href="https://pub.beakernotebook.com/#/sign_up">Create New Account</a>
       </div>
     </div>
     <div class="publish" ng-if="isSignedIn()">
--- a/core/src/main/web/app/mainapp/components/publication/publish.jst.html
+++ b/core/src/main/web/app/mainapp/components/publication/publish.jst.html
@@ -1,0 +1,61 @@
+<!--
+    Copyright 2015 TWO SIGMA OPEN SOURCE, LLC
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+           http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+-->
+<div>
+  <div class="modal-header">
+    <div class="sign-in" ng-if="!isSignedIn()">
+      <h1 class="error" ng-if="error">{{error}}</h1>
+      <h1 ng-if="!error">Sign in to Beaker Publications</h1>
+      <form ng-submit="signIn()">
+        <input ng-model="user.email" type="email" name="email" placeholder="Email" class="field">
+        <input ng-model="user.password" type="password" required name="password" placeholder="Password" class="field">
+        <input type="submit" value="Sign in" class="btn btn-primary">
+      </form>
+      <div class="sign-up">
+        No account?
+        <a target="_blank" href="https://pub.beakernotebook.com/#/sign_up">Create New Account</a>
+      </div>
+    </div>
+    <div class="publish" ng-if="isSignedIn()">
+      <h1>{{title}}</h1>
+      <form ng-submit="publishAction()">
+        <label for="category" class="full">Notebook name others will see when you publish:</label>
+        <input ng-model="model.name" required type="text" class="field">
+        <label for="category" class="full">Notebook description:</label>
+        <textarea ng-model="model.description" required rows="3" class="field"></textarea>
+        <label for="category" class="full">Category that describes your notebook:</label>
+        <select ng-model="model['category-id']" required ng-options="category['public-id'] as category.name for category in categories" name="category" ng-class="{error: error}" ng-change="refresh()" class="field">
+          <option value="">Select a category</option>
+        </select>
+        <ul ng-if="published" class="publication-details">
+          <li>
+            <div class="time-label">Published</div>
+            <div class="value publish-time">{{model['created-at'] | date:'short'}}</div>
+          </li>
+          <li>
+            <div class="time-label">Last updated</div>
+            <div class="value update-time">{{model['updated-at'] | date:'short'}}</div>
+          </li>
+        </ul>
+        <div class="actions">
+          <div class="warning" ng-if="edited">Warning, your notebook has unsaved changes.</div>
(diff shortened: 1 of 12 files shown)


grin:
One thing I wanted to do but couldn't achieve easily was to move the API base url from: https://github.com/twosigma/beaker-notebook/commit/7031b7d55fa3bd546d76a21df98b1d9c85e93fd2#diff-267c5d1e1ba95843db484b01357360f9R25 to a config file. I tried adding it to `beaker.conf` but it was a bit tricky and messy to initialize the angular API service asynchronously with the url value (obtained via a rest call) ...so I ditched the idea and left the url hardcoded :/

core/src/main/web/plugin/menu/file.js
```diff
@@ -74,13 +74,13 @@ define(function(require, exports, module) {
       items: []
     },
     {
-      name: "Share to Web",
```
scottdraves: don't remove the existing mechanism, there should be 2 submenu items.

core/src/main/web/app/mainapp/components/publication/publish.jst.html
```diff
@@ -0,0 +1,36 @@
...
+      <h1 class="error" ng-show="error">{{error}}</h1>
+      <h1 ng-hide="error">Sign in to Beaker Publications</h1>
+      <form ng-submit="signIn()">
+        <input ng-model="user.email" type="email" name="email" placeholder="Email" id="user-email">
+        <input ng-model="user.password" type="password" required name="password" placeholder="Password" id="user-password" class="">
+        <input type="submit" value="Sign in" class="btn btn-primary">
+      </form>
+      <div class="sign-up">
+        No account?
+        <a target="_blank" href="https://beaker-staging.withmojo.com/#/sign_up">Create New Account</a>
```
scottdraves: i believe davide has the real URL

core/src/main/web/plugin/menu/file.js
grin: :+1:

core/src/main/web/app/mainapp/components/publication/publish.jst.html
grin: @ildipo could you send me the real url? Thanks.

grin:
@scottdraves @ildipo I asked about this already in PT but figured maybe it's better to ask here, so let me repeat my question...

I'm looking for a way to maintain a notebook-to-publication relationship in Beaker Solo which is gonna be needed for updating publications (once a notebook is published we want it to have the id of the corresponding publication so we can update the publication it in the future)...is there a way to add some kind of  metadata containing the publication-id to a notebook? Which is saved in the notebook session/file and is easy to retrieve programmatically?

Currently I'm investigating using Notebook Namespaces for this purpose...but I'm not yet confident if it's a good choice.

scottdraves:
it should be OK to add a "metadata" dict to the notebook model.

grin:
ok, gonna try this in my next PR, thanks

grin:
Updated the PR:
- brought back the "Share" menu item
- implemented the actual publishing functionality in: a65c1ec (this also adds metadata to the notebook model)

core/src/main/web/app/styles/modal.scss
```diff
@@ -70,3 +70,55 @@ html {
...
...yYXRvcjogU2tldGNoIDMuMC40ICg4MDUzKSAtIGh0dHA6Ly93d3cuYm9oZW1pYW5jb2RpbmcuY29tL3NrZXRjaCAtLT4KICAgIDx0aXRsZT5GaWxsIDMzPC90aXRsZT4KICAgIDxkZXNjPkNyZWF0ZWQgd2l0aCBTa2V0Y2guPC9kZXNjPgogICAgPGRlZnM+PC9kZWZzPgogICAgPGcgaWQ9IlBhZ2UtMSIgc3Ryb2tlPSJub25lIiBzdHJva2Utd2lkdGg9IjEiIGZpbGw9Im5vbmUiIGZpbGwtcnVsZT0iZXZlbm9kZCIgc2tldGNoOnR5cGU9Ik1TUGFnZSI+CiAgICAgICAgPGcgaWQ9IjEuLVB1Ymxpc2hpbmctZmxvd3MiIHNrZXRjaDp0eXBlPSJNU0FydGJvYXJkR3JvdXAiIHRyYW5zZm9ybT0idHJhbnNsYXRlKC0yNjMuMDAwMDAwLCAtNjU4LjAwMDAwMCkiIGZpbGw9IiMzMzNCM0IiPgogICAgICAgICAgICA8ZyBpZD0iR3JvdXAiIHNrZXRjaDp0eXBlPSJNU0xheWVyR3JvdXAiIHRyYW5zZm9ybT0idHJhbnNsYXRlKDExOS4wMDAwMDAsIDMyNS4wMDAwMDApIj4KICAgICAgICAgICAgICAgIDxnIGlkPSJHcm91cC04LSstUHVibGlzaGluZy1vcHRpb25zOi00IiB0cmFuc2Zvcm09InRyYW5zbGF0ZSgyOC4wMDAwMDAsIDI5Ni4wMDAwMDApIiBza2V0Y2g6dHlwZT0iTVNTaGFwZUdyb3VwIj4KICAgICAgICAgICAgICAgICAgICA8ZyBpZD0iR3JvdXAtOCIgdHJhbnNmb3JtPSJ0cmFuc2xhdGUoMC4wMDAwMDAsIDIzLjAwMDAwMCkiPgogICAgICAgICAgICAgICAgICAgICAgICA8cGF0aCBkPSJNMTIxLDE5IEwxMjYsMTQgTDExNiwxNCBMMTIxLDE5IFoiIGlkPSJGaWxsLTMzIj48L3BhdGg+CiAgICAgICAgICAgICAgICAgICAgPC9nPgogICAgICAgICAgICAgICAgPC9nPgogICAgICAgICAgICA8L2c+CiAgICAgICAgPC9nPgogICAgPC9nPgo8L3N2Zz4=')) no-repeat right 8px center;
```
scottdraves: no inline image please, use core/src/main/web/app/images

ildipo:
the URL will be http://pub.beakernotebook.com/

grin:
Updated the PR:
- used the publication api url from @ildipo 
- implemented @samccone 's code review suggestions (introduced a service that stores the current user instance: https://github.com/twosigma/beaker-notebook/commit/298cf99d972776c5ca24bc684dc0dd470dea0f4f#diff-81bdedc1692c3c7f959a697ad1c99180R52 )
- removed the inline css image
- added delete publication button

grin:
Hey, I'm running into an issue where metadata I'm adding to a notebook model in https://github.com/twosigma/beaker-notebook/commit/738a3f0d2fbd20cb3e83199a9538df1d02419a5a#diff-fc1679838ea7d632d447077e8dabdff6R82 is not always saved. Based on my observations attaching the publication id as metadata works if I execute some cells in the notebook prior to publishing and it fails when there are no output cells present. Any thoughts what could be causing this behavior?

samccone:
What is `bkSessionManager.getRawNotebookModel()` before you have evaluated any code?

grin:
@samccone Ok, I wasn't too accurate with that issue description, turns out that was just an error on my end...the value of `bkSessionManager.getRawNotebookModel()` is correct, I just need to force notebook save after obtaining a publication id in a scenario where new notebook is being published. Thanks

grin:
Fixed that one issue and re-pushed.

ildipo:
This PR looks good. I only updated it to use http instead of https until we get the proper certificate for the server.

ildipo:
I'm going to merge after CI finishes

grin:
Updated.

I haven't added any explicit error handling, I only improved a scenario where if a notebook is published in Beaker and then the publication gets deleted on the Publications website, the UI in Beaker will just act as if the notebook was never published if you go to publish it agin. 

For any errors related to the API server not being available I'm assuming it's ok for now if we let users to investigate with browser tools (network inspector, js console)

ildipo:
@grin you probably forgot to push your changes. I'm merging this anyway, open a new PR with the further changes.

grin:
@ildipo I did push my changes...I force pushed them and overwrote the whole branch, the actual change was in the initPublication function: https://github.com/twosigma/beaker-notebook/commit/fc262df2a6bc74ae85d5b3633fb0fa481fec3613#diff-fc1679838ea7d632d447077e8dabdff6R45

where I'm now marking the notebook as published only when there is a corresponding publication on the Publication server (previously I only looked at the publication-id saved in the notebook metadata, but Sam pointed out we should make it more robust).

ildipo:
@grin that's ok didn't notice it. everything is good then

PR #1958, opened 2015-07-20
Data tables

### LINKED ISSUE #1905: 🚰 datatable output leaks
### Memory leak on datatable output investigation :mag:

screen shot 2015-07-08 at 12 15 43 pm

**Hypothesis**: When displaying and removing datatable outputs we are leaking listeners and DOM nodes.
##### Testing Methodology:
- create an empty notebook
- add a new JS cell
- enter `a = {1:1, 2:2};`
- evaluate cell
- remove output cell
- take a heap snapshot
- start the profiler and force a GC
- (repeat below 5 times)
  - enter `a = {1:1, 2:2};`
  - evaluate
  - remove output cell
- force a GC and stop the profiler
##### Results:

Profiler results:

screen shot 2015-07-08 at 11 46 09 am

The profiler seems to indicate that we do have a pretty significant leak of listeners and DOM nodes, to get a better idea of what exactly is leaking lets take a look at the heap snapshot comparison view.

screen shot 2015-07-08 at 12 10 14 pm

screen shot 2015-07-08 at 12 11 25 pm

Ok from here I took at look into the retaining paths for the HTML elements in the memory graph, to get a feel for who might be leaking (or where we might need to unregister)

screen shot 2015-07-08 at 11 57 15 am
screen shot 2015-07-08 at 11 56 26 am
screen shot 2015-07-08 at 11 56 10 am
screen shot 2015-07-08 at 11 55 59 am

These results seem to point jquery at the base and more specifically the actual table plugin that we are using.

---

Ideas on how to fix? 
- Look into the table directive and make sure that we are explicitly tearing down the table plugin
- Do a look into any events that we are registering for on the output cell and make sure that we are unbinding correctly.

samccone:
switch to bower for data tables

---

Fixes listener leaks in the fixed column plugin.

DataTables/FixedColumns@c0f03d8
DataTables/FixedColumns@f8669d4
DataTables/FixedColumns@a98d107

Fixes #1905

---

Ideally will be fixed by
https://github.com/DataTables/FixedColumns/issues/46


samccone:
~~working through trying to isolate the failure on CI~~

case sensitive filenames :clap:

PR #1959, opened 2015-07-20
tests: remove R setup for travis

samccone:
Currently we have no R tests in place, so this can be removed for now
since it causes a significant slowdown

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ before_install:
 before_script:
   - sudo add-apt-repository --yes ppa:nginx/stable
   - sudo add-apt-repository --yes ppa:chris-lea/zeromq
-  - sudo add-apt-repository --yes ppa:marutter/rrutter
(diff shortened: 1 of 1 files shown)


scottdraves:
can you add an R test (another 1+1 is fine) instead?

samccone:
Just a follow up @scottdraves this cuts around 15 min off of the total build time on travis, since we do not have any R tests for the time being I figured this would be ok. We can circle back and add some R evaluation tests if you would like, I still think this PR might be useful even for the short term since it will shorten the iteration cycle.

scottdraves:
with your nice evalInLanguage() function, can't you add an R test with 3 lines?
that is definitely the right way to move on this.  it sucks that it takes so long to install R but it's an important functionality for us.

samccone:
k no worries :+1: I will add a test in for R and hopefully it should just work

PR #1964, opened 2015-07-21
Allow CodeMirror hints in embeddable beaker

buob:
Before this our styles for codemirror hints were scoped to beaker sandbox but the hint dialog was appended to body, so this just makes those codemirror hint styles global.


samccone:
:+1: lgtm

PR #1969, opened 2015-07-22
some visibility changes in java classes

ildipo:


samccone:
Just curious as to why :)

PR #1974, opened 2015-07-22
🔨 Fix broken dashboard save

### LINKED ISSUE #1909: Saving on close from control panel does not save most recent changes
To reproduce:
1. Create new notebook
2. Create new cell, type something
3. Open control panel in different tab
4. Close the session of the notebook you created from the control panel tab
5. Accept to save changes, save to some file
6. Open file, see that the cell you created is missing

samccone:
When a user had multiple windows open, if they were to save from the dashboard it was possible to save an old version of the notebook.

This change makes it so a save from the dashboard will always use the latest backup.

---

Using the latest backup did raise an interesting problem however. Since we used to only update the backup on a poll we could reach a condition where we were backing up an old version. To fix this I now backup the notebook on edit state change

---

Fixes #1909

---

**Implementation notes**

As @DiegoAlbertoTorres pointed out there is a cost in calling backup eagerly for large notebooks. A solution that we can implement ontop of this PR would be to do what is outlined here #1975

### DIFF
--- a/core/src/main/web/app/controlpanel/controlpanelsessionitem-directive.js
+++ b/core/src/main/web/app/controlpanel/controlpanelsessionitem-directive.js
@@ -75,7 +75,7 @@
                             });
                           } else {
                             var fileSaver = bkCoreManager.getFileSaver(pathInfo.uriType);
-                            fileSaver.save(pathInfo.uri, notebookModelAsString).then(function () {
+                            fileSaver.save(pathInfo.uri, notebookModelAsString).then(function() {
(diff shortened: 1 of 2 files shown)


scottdraves:
backups should only trigger once per minute.  please do not confuse backups with saving.

now, about this bug.  when the user asks to save from the control panel, can't the server use comet to get a copy of the latest notebook?  i do not see that any extra large copies are required as the extra back-and-forth between client and server is small, just the trigger, not the payload.

samccone:
sure I will update this to work that way

PR #1979, opened 2015-07-24
Fix broken dashboard save

### LINKED ISSUE #1909: Saving on close from control panel does not save most recent changes

samccone:
Fixes #1909

### DIFF
--- a/core/src/main/web/app/controlpanel/controlpanelsessionitem-directive.js
+++ b/core/src/main/web/app/controlpanel/controlpanelsessionitem-directive.js
@@ -47,9 +47,8 @@
             });
 
           };
-          bkUtils.httpGet('rest/session-backup/getEdited', {
-            sessionid: session.id
-          }).then(function(response) {
+          bkSession.getSessionEditedState(session.id)
+          .then(function(response) {
             var edited = response.data.edited;
             if (!edited) {
               // close session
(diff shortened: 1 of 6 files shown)


scottdraves:
I am getting the new copy of the notebook, but the file has not been pretty printed, and multiline cells have not been broken into arrays of lines.

samccone:
:+1: will fix

core/src/main/web/app/controlpanel/controlpanelsessionitem-directive.js
```diff
@@ -25,6 +25,33 @@
...
+        .catch(deferred.reject);
+
+        $.cometd.unsubscribe('/latest-notebook-model');
+      });
+
+      $.cometd.publish('/request-latest-notebook-model', {sessionId: sessionId});
+
+      return deferred.promise;
+    }
+
```
DiegoAlbertoTorres, line 54: Why not move the save to bkNotebookManager?

samccone, line 54: I approached this problem with the general assumption that the controlpanel has responsibility for saving the notebook. The reason for this approach was because the controlpanel has the responsibility over handing errors, issuing the final close command to all open notebook sessions, and providing feedback to the user about the operation.

If we were to change where the execution (save) was happening we would be essentially creating an RPC approach across windows / tabs.

Is this what we want?

PR #1980, opened 2015-07-24
Diego/electon window management

DiegoAlbertoTorres:


core/src/main/web/app/helpers/helper.js
```diff
@@ -163,7 +163,27 @@
...
+            if (uriType) {
+              routeParams.type = uriType;
+            }
+            if (readOnly) {
+              routeParams.readOnly = true;
+            }
+            if (format) {
+              routeParams.format = format;
+            }
+            var serialize = function(obj) {
```
DiegoAlbertoTorres: I am almost sure we already have something like this function. Anyone know? Is there a cleaner alternative? Or should I just throw this in bkUtils?

core/src/main/web/app/helpers/helper.js
```diff
@@ -163,7 +163,27 @@
...
+              var str = [];
+              for(var p in obj)
+                if (obj.hasOwnProperty(p)) {
+                  str.push(encodeURIComponent(p) + "=" + encodeURIComponent(obj[p]));
+                }
+              return str.join("&");
+            }
```
DiegoAlbertoTorres: I am almost sure we already have something like this function. Anyone know? Is there a cleaner alternative? Or should I just throw this in bkUtils?

samccone: http://api.jquery.com/jquery.param/ :)

Issue #1618, opened 2015-05-15
404 from plots


scottdraves:
GET http://127.0.0.1:8801/3e4a523/beaker/app/dist/images/ui-icons_222222_256x240.png 404 (Not Found)curCSS @ beakerVendor.js:19560jQuery.extend.css @ beakerVendor.js:19537isHidden @ beakerVendor.js:19315showHide @ beakerVendor.js:19348jQuery.fn.extend.show @ beakerVendor.js:19402jQuery.each.jQuery.fn.(anonymous function) @ beakerVendor.js:21536t.fn.extend.show @ beakerVendor.js:95907e.widget._create._renderAxis @ beakerVendor.js:95906e.widget._create @ beakerVendor.js:95906(anonymous function) @ beakerVendor.js:95906e.Widget._createWidget @ beakerVendor.js:95906e.widget.e.(anonymous function).(anonymous function) @ beakerVendor.js:95906(anonymous function) @ beakerVendor.js:95906jQuery.extend.each @ beakerVendor.js:13801jQuery.fn.jQuery.each @ beakerVendor.js:13448e.widget.bridge.e.fn.(anonymous function) @ beakerVendor.js:95906link @ beakerOutputDisplay.js:7092nodeLinkFn @ beakerVendor.js:48485compositeLinkFn @ beakerVendor.js:47879publicLinkFn @ beakerVendor.js:47775link.refresh @ beakerApp.js:7543(anonymous function) @ beakerApp.js:7546$get.Scope.$digest @ beakerVendor.js:54383$get.Scope.$apply @ beakerVendor.js:54648done @ beakerVendor.js:50183completeRequest @ beakerVendor.js:50397xhr.onreadystatechange @ beakerVendor.js:50336

PR #1981, opened 2015-07-27
fix 404 error for image, close Issue #1618

kpaharelau:


samccone:
hey @kpaharelau thanks for the PR, can you describe how I can reproduce this issue / verify that it has fixed the issue?

kpaharelau:
Hi. You can try to upload  /core/config/tutorials/chartingTutorial.bkr file (File->Upload(.bkr)).  You will see 404 error in console log. The problem is in the gulpfile.js. Some vendor's image files did not get into the assembly. The problem disappears after applying my changes

samccone:
cool thanks @kpaharelau 

Wondering if it might be more appropriate to move this copy task into
https://github.com/twosigma/beaker-notebook/blob/master/core/build.gradle#L108

Since we are already copying vendor files in our gradle task

kpaharelau:
Interesting :-). I will analyze the problem.

kpaharelau:
Hi, I fixed this problem using gradle. Now I have access to repo and made pull request using branch from it. https://github.com/twosigma/beaker-notebook/pull/1994

samccone:
thanks @kpaharelau 

@spot I think we should be able to close this PR in favor of #1994

Issue #1452, opened 2015-04-13
chart with one point crashes


scottdraves:
![screen shot 2015-04-13 at 3 48 36 pm](https://cloud.githubusercontent.com/assets/963093/7124105/999ccd6c-e1f4-11e4-8a1d-c883c1af94a6.png)

on the JS console:

2beakerOutputDisplayVendor.js:596 Error: Invalid value for <line> attribute x1="NaN"
2beakerOutputDisplayVendor.js:596 Error: Invalid value for <line> attribute x2="NaN"
2beakerOutputDisplayVendor.js:596 Error: Invalid value for <line> attribute x1="NaN"
2beakerOutputDisplayVendor.js:596 Error: Invalid value for <line> attribute x2="NaN"

PR #1993, opened 2015-07-28
chart with one point crashes (#1452)

mnavasiolava:
The problem was in plotGridlines function - data for gridlines (scope.rpipeGridlines) had NaN values.
This happened because plot had only one point and model range had equal xl/xr values. As the result we've got division by zero during calculating element.x percent value in PlotPoint.prototype.applyAxis.
To fix it I suggest increasing the model range if all elements have the same x or y axis coordinates.

### DIFF
--- a/core/src/main/web/outputdisplay/bko-plot/plotitems/std/plotpoint.js
+++ b/core/src/main/web/outputdisplay/bko-plot/plotitems/std/plotpoint.js
@@ -79,6 +79,14 @@
         range.yl = Math.min(range.yl, ele.y);
         range.yr = Math.max(range.yr, ele.y);
       }
+      if(range.xl == range.xr){
+        range.xl = range.xl - 0.1;
+        range.xr = range.xr + 0.1;
+      }
+      if(range.yl == range.yr){
+        range.yl = range.yl - 0.1;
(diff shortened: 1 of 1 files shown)


core/src/main/web/outputdisplay/bko-plot/plotitems/std/plotpoint.js
```diff
@@ -79,6 +79,14 @@
         range.yl = Math.min(range.yl, ele.y);
         range.yr = Math.max(range.yr, ele.y);
       }
+      if(range.xl == range.xr){
```
samccone: we should use strict equality here

core/src/main/web/outputdisplay/bko-plot/plotitems/std/plotpoint.js
```diff
@@ -79,6 +79,14 @@
...
+        range.xl = range.xl - 0.1;
+        range.xr = range.xr + 0.1;
+      }
+      if(range.yl == range.yr){
```
samccone: and here

scottdraves:
:+1:

ildipo:
everyone seems happy so I merge it :-)

PR #1994, opened 2015-07-28
fix 404 error for JQueryUI images, close Issue #1618

kpaharelau:
use gradle to copy jquery-ui images from original vendor location to dist/images

For reproducing this bug you can try open Help menu, select "Tutorial", and then click on the link "Interactive Charts".
 You can see 404 error in console log. This error will disappear after applying my changes


scottdraves:
:+1:

samccone:
:clap: :+1:

Issue #1450, opened 2015-04-13
legend can overflow cell output


scottdraves:
too many lines in the legend can overflow the output

![screen shot 2015-04-13 at 3 34 49 pm](https://cloud.githubusercontent.com/assets/963093/7123804/bd0341c0-e1f2-11e4-86e4-1930a968fdac.png)

PR #1995, opened 2015-07-28
legend can overflow cell output (#1450)

mnavasiolava:
Legend is wrapped in scrollable container

### DIFF
--- a/core/src/main/web/outputdisplay/bko-plot/bko-plot.js
+++ b/core/src/main/web/outputdisplay/bko-plot/bko-plot.js
@@ -544,17 +544,25 @@
           scope.jqcontainer.find("#legends").remove();
 
           scope.legendDone = true;
-          var legend = $("<table></table>").appendTo(scope.jqcontainer)
-            .attr("id", "legends")
-            .attr("class", "plot-legendcontainer")
+          var legendScrollableContainer = $("<div></div>").appendTo(scope.jqcontainer)
+            .attr("class", "plot-legendscrollablecontainer")
             .draggable({
               stop : function(event, ui) {
                 scope.legendPosition = {
                   "left" : ui.position.left,
                   "top" : ui.position.top
                 };
-              }
-            });
+              },
+              handle : "#legendDraggableContainer"
+            })
+            .css("max-height", scope.jqsvg.height() - scope.layout.bottomLayoutMargin);
+
+          var legendDraggableContainer = $("<div></div>").appendTo(legendScrollableContainer)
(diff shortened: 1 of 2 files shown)


core/src/main/web/outputdisplay/bko-plot/bko-plot.js
```diff
@@ -544,17 +544,25 @@
...
                 };
-              }
-            });
+              },
+              handle : "#legendDraggableContainer"
+            })
+            .css("max-height", scope.jqsvg.height() - scope.layout.bottomLayoutMargin);
+
+          var legendDraggableContainer = $("<div></div>").appendTo(legendScrollableContainer)
+            .attr("id", "legendDraggableContainer")
```
samccone, line 561: since a page should only have 1 ID I wonder what this is going to do on a page with multiple IDs

scottdraves:
good, i got the scroll bar if the legend is too long.  but If a line's label has more than one word, this is causing them to be broken into multiple lines:
screen shot 2015-07-28 at 9 57 02 am
should be
screen shot 2015-07-28 at 9 57 58 am

see the first example in the charting tutorial to reproduce this.

mnavasiolava:
@scottdraves fix it in 54c59a519f745f3c9c3dd2a60ca6635f0a486006

core/src/main/web/outputdisplay/bko-plot/bko-plot.js
mnavasiolava, line 561: You mean #legendDraggableContainer? jQuery draggable.handle works only with elements that descend from the draggable element. I've tested 2 plots with legend on the same page – legends are ok

scottdraves:
:+1:

core/src/main/web/outputdisplay/bko-plot/bko-plot.js
samccone, line 561: I am guessing this is just the browser letting us be wrong without blowing up.

we are not supposed to use the same ID more than once in a document.
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

> The id global attribute defines a unique identifier (ID) which must be unique in the whole document. Its purpose is to identify the element when linking (using a fragment identifier), scripting, or styling (with CSS).

scottdraves, line 561: It would be better not to reuse the ID but it looks like it works fine and before this PR it was already reusing the ID, so we don't have to fix it here and now.

Issue #900, opened 2014-11-20
plots can have too many ticks and lines


scottdraves:
the following groovy code produces a plot with too many and too big numbers/ticks on the axes.

```
def y = (1..30000).collect({ Math.random() * 10e13 })
def p = new Plot()
p << new Line(y: y)
return p
```

PR #1997, opened 2015-07-28
fix plots can have too many ticks and lines #900

kpaharelau:
I did a algorithm for  calculation grid lines more flexible. Algorithm  looks for an interval where the number of intervals stops decreasing and begins to grow. The intervals may be added during the algorithm working


scottdraves:
:+1:

Issue #1460, opened 2015-04-13
higher level API to plotting


scottdraves:
from a user:

When I have a timeseries one of the first things I want to do is plot some simple timeseries and summary statistics (histograms, quartiles, whatever). Probably the first thing everyone wants to do.
I think it’s realistic to want to be able to type something like:
table =
{ something complicated }
plot = SimpleTimePlot(table, ["column1", "column2", “column3”, “column4”], [other args to control whatever axes, labels, colors])

Instead, we have:
timeCol = table*.getTimeInMillis() <- spread operator = super implementation detail
plot = new TimPlot(...)
plot << new Line(x:timeCol, y: table.column1, ...)
plot << new Line(x:timeCol, y: table.column2, ...)
plot << new Line(x:timeCol, y: table.column3, ...)  starting to get tired of typing
plot << new Line(x:timeCol, y: table.column4, ...)
At which point I'm wondering why my plot has lines but no legend – it’s because I need to go back and add a "displayName" to all the lines... where a "simple" default implementation could just default to the column name provided. Since I don’t remember all of the arguments and the spread operator and whatever, I end up looking at the documentation every time just to make a simple time plot of the data in my frame.

It would be nice to be able to write summarize(table) or something and get some basic stats and plots.

PR #2007, opened 2015-07-30
fix higher level API to plotting #1460

kpaharelau:
fix higher level API to plotting #1460

More simple way to create simlpe TimePlot was added.

Now you can try to use SimpleTimePlot (plot = new SimpleTimePlot(table, ["column1", "column2", “column3”, “column4”], [other args to control whatever axes, labels, colors])).
Use 1460.bkr (https://www.dropbox.com/s/lbb2r7qfknf8l0n/1460.bkr?dl=0) to try.

Notes

default the names for the lines and points to the same as the column
default the xLabel to "Time"
default usetoolTip and showLegend to true
it can select colors automatically using a generated nice colors

parameres :

timeColumn: determines what column will be used for a time axis. "time"  by default

colors:[X, Y, Z...]
each X could be [r,g,b] or just a string, like you can do with css
including names:
white
WHITE
lightGray
LIGHT_GRAY
gray  
GRAY
darkGray
DARK_GRAY
black 
BLACK
red 
RED
pink
PINK
orange
ORANGE
yellow
YELLOW
green 
GREEN
magenta
MAGENTA
cyan
CYAN
blue
BLUE

displayLineNames: display names for Lines
displayPointNames: display names for Points

displayLines: Lines will be added on the plot if true
displayPoints:  Points will be added on the plot if true

### DIFF
--- a/plugin/jvm/src/main/java/com/twosigma/beaker/chart/xychart/SimpleTimePlot.java
+++ b/plugin/jvm/src/main/java/com/twosigma/beaker/chart/xychart/SimpleTimePlot.java
@@ -42,7 +42,7 @@ public class SimpleTimePlot extends TimePlot {
     private static final Color[] NICE_COLORS = {
             new Color(54, 54, 217),   //blue hue = 240
             new Color(216, 54, 54),   //red hue = 0
-            new Color(54, 216, 170),  //green hue = 163
+            new Color(54, 216, 54),   //green hue = 120
             new Color(192, 54, 216),  //purple hue = 291
             new Color(54, 126, 186),  //cyan hue = 169
             new Color(216, 178, 54),  //yellow hue = 46
--- a/plugin/jvm/src/main/java/com/twosigma/beaker/chart/xychart/SimpleTimePlot.java
+++ b/plugin/jvm/src/main/java/com/twosigma/beaker/chart/xychart/SimpleTimePlot.java
@@ -51,7 +51,9 @@ public class SimpleTimePlot extends TimePlot {
     this(null, data, columns);
   }
 
-  public SimpleTimePlot(Map<String, Object> parameters, List<Map<String, Object>> data, List<String> columns) {
+  public SimpleTimePlot(Map<String, Object> parameters,
+                        List<Map<String, Object>> data,
+                        List<String> columns) {
 
     this.data = data;
     this.columns = columns;
@@ -118,7 +120,9 @@ public class SimpleTimePlot extends TimePlot {
   private Color createChartColor(Object color) {
     if (color instanceof List) {
       try {
-        return new Color((int) ((List) color).get(0), (int) ((List) color).get(1), (int) ((List) color).get(2));
+        return new Color((int) ((List) color).get(0),
+                         (int) ((List) color).get(1),
+                         (int) ((List) color).get(2));
       } catch (IndexOutOfBoundsException x) {
         throw new RuntimeException("Color list too short");
       }
--- a/plugin/jvm/src/main/java/com/twosigma/beaker/chart/xychart/SimpleTimePlot.java
+++ b/plugin/jvm/src/main/java/com/twosigma/beaker/chart/xychart/SimpleTimePlot.java
@@ -27,361 +27,359 @@ import java.util.*;
 
 public class SimpleTimePlot extends TimePlot {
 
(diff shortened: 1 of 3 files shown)


scottdraves:
This looks great.  I am seeing a problem with the legend check boxes though.  Here's a really simple use of the API: http://sharing.beakernotebook.com/gist/anonymous/b02f32043b02cbaf003a
I then clicked both boxes for the "low" line, but it did not vanish.

scottdraves:
Use less saturated colors by default (75%) and put them in this order:
blue red green purple cyan yellow 
also less bright, 85%.

first color is: 54 54 217

kpaharelau:
This looks great. I am seeing a problem with the legend check boxes though. Here's a really simple use of the API: http://sharing.beakernotebook.com/gist/anonymous/b02f32043b02cbaf003a
I then clicked both boxes for the "low" line, but it did not vanish. -FIXED

Use less saturated colors by default (75%) and put them in this order:
blue red green purple cyan yellow
also less bright, 85%.

first color is: 54 54 217

I used collection nice colors for first 6 elements
//saturation 75%
//brightness 85%
private static final Color[] NICE_COLORS = {
    new Color(54, 54, 217),   //blue hue = 240
    new Color(216, 54, 54),   //red hue = 0
    new Color(54, 216, 170),  //green hue = 163
    new Color(192, 54, 216),  //purple hue = 291
    new Color(54, 126, 186),  //cyan hue = 169
    new Color(216, 178, 54),  //yellow hue = 46
}

next  elements gets generated color (random hue parameter, saturation 75% and brightness 85%).

But I have doubt that this is exactly what you need. What is your opinion? Probably better to use an algorithm from the XYChartSerializer?

 private static class ColorPalette {
    static final Color[] colors = new Color[]{
      new Color(192, 80, 77),
      new Color(79, 129, 189),
      new Color(155, 187, 89),
      new Color(247, 150, 70),
      new Color(128, 100, 162),
      new Color(75, 172, 198)
    };

```
 static Color getColor(int i) {
  return colors[i % colors.length];
}
```

  }

...
 int i = 0;
    for (XYGraphics g : xychart.getGraphics()) {
      if (g.getColor() == null) {
        g.setColori(ColorPalette.getColor(i++));
      }
    }
...

scottdraves:
Thanks that's much better.  But green hue is off, it should be 120, not 163.
Don't worry about more than 6 colors this is fine for a start.
Continuing my review...

kpaharelau:
But green hue is off, it should be 120, not 163. - fix

scottdraves:
follow our style guide (indentation 2)

kpaharelau:
follow our style guide (indentation 2) - fixed

plugin/jvm/src/main/java/com/twosigma/beaker/chart/xychart/SimpleTimePlot.java
```diff
@@ -0,0 +1,385 @@
...
+        String colorAsStr = (String) color;
+        if (colorAsStr.indexOf("#") == 0) {
+          return Color.decode(colorAsStr);
+        } else {
+          Field field = Class.forName("com.twosigma.beaker.chart.Color").getField(colorAsStr);
+          return (Color) field.get(null);
+        }
+      }
+    } catch (Throwable ignored) {
+      //expected
```
scottdraves: what error is expected?  probably better to avoid any error.

scottdraves:
make the default for displayPoints to false.
replace displayLineNames and displayPointNames with displayNames.

kpaharelau:
User can make mistakes. For example 
 colors: ["#ffff00AA", "green", [64, 64], "rad"]

In this case color will be taken from nice colors or generated

scottdraves:
If the user gives an illegal color then you should throw an error and say what is wrong.

kpaharelau:
Ireplace displayLineNames and displayPointNames with displayNames. - Use same for lines and points?

If the user gives an illegal color then you should throw an error and say what is wrong. - Ok

scottdraves:
> Use same for lines and points?

yes

kpaharelau:
Color select algorithm has been improved according Scott remarks

make the default for displayPoints to false.
replace displayLineNames and displayPointNames with displayNames.  - FIXED

scottdraves:
:+1:

Issue #1459, opened 2015-04-13
plot of horizontal line broken


scottdraves:
should get a horizontal line in the middle of the vertical space with decent axes and labels.
instead get:
http://sharing.beakernotebook.com/gist/anonymous/835fe84ecb06b633c950
![screen shot 2015-04-13 at 6 49 20 pm](https://cloud.githubusercontent.com/assets/963093/7127144/f2ace982-e20d-11e4-8298-cdc25ad54900.png)

PR #2011, opened 2015-07-30
plot of horizontal line broken (#1459)

mnavasiolava:
Wide datarange when all ys/xs are the same. Works both for point and lines


scottdraves:
:+1:

ildipo:
:+1:

mnavasiolava:
not vs, it just was made in different place to make it work both for points and lines.
"daterange.yl - daterange.yl / 10" means that we want to widen the range with the value 10 times lower than the initial yl

samccone:
ok, I am just trying to understand how this fixes it :)

Thanks for explaining!

PR #2012, opened 2015-07-30
Add memory tests

samccone:
it works!
https://travis-ci.org/twosigma/beaker-notebook/builds/73565086#L3970-L4009

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,15 @@
 language: java
 jdk:
  - oraclejdk7
+env:
+  - chromeBinaryPath=/home/travis/build/twosigma/beaker-notebook/chrome-linux/chrome
 before_install:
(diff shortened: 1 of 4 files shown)


test/memory-tests.js
```diff
@@ -0,0 +1,162 @@
+var drool = require('drool');
```
scottdraves, line 17: needs banner

samccone, line 17: good catch! one moment

.travis.yml
```diff
@@ -1,10 +1,15 @@
 language: java
 jdk:
  - oraclejdk7
```
scottdraves, line 3: should be 8

scottdraves, line 3: i filed a separate issue for this

Issue #1457, opened 2015-04-13
add ability to add 2nd Y axis to plots


scottdraves:
in the charting tutorial you should be able to have a cell like:

def priceChart = new TimePlot(xLabel: "Time", yLabel: "Price", useToolTip: true, showLegend: true,
                              crosshair: new Crosshair(color: Color.gray, width: 2, style: StrokeType.DOT))
priceChart << new Line(x: ibm2008.time, y: ibm2008.low, displayName: "Low Price", color: new Color(33, 87, 141))
priceChart << new YAxis(label: "Volume", upperMargin: 2.0)
priceChart << new Line(x: ibm2008.time, y: ibm2008.volume, displayName: "Volume", yAxis: "Volume")

that produces an axis on the right side for volume.

http://sharing.beakernotebook.com/gist/anonymous/7be3d2800684732febc3

PR #2039, opened 2015-08-04
add ability to add 2nd Y axis to plots #1457

mnavasiolava:
If more than one yAxis is added, then the 2nd yAxis is displayed on the right side of the graph.
To display right yAxis rightLayoutMargin was increased.
Crosshair labels are displayed both on left and right yAxes.
Line relates to right yAxis, if line.yAxis is equal to yAxis label.


scottdraves:
:+1:

scottdraves:
Will merge as soon as conflicts are resolved.

PR #2043, opened 2015-08-05
Default LoD should be river, not line #2021

mnavasiolava:
change default lod type index for plotlinelodloader


scottdraves:
:+1:

Issue #903, opened 2014-11-21
plots should have a resize handle.


scottdraves:
and they should remember how you resized them, so if you reevaluate, it isn't lost.

PR #2044, opened 2015-08-05
fix plots should have a resize handle. #903

kpaharelau:
1. Detect plot size changes by css
2. Changed model state now saved (call  kSessionManager.setNotebookModelEdited(true))
3. Model state now saved on "destroy" event

### DIFF
--- a/core/src/main/web/outputdisplay/bko-plot/bko-plot.js
+++ b/core/src/main/web/outputdisplay/bko-plot/bko-plot.js
@@ -88,8 +88,6 @@
         scope.initLayout = function() {
           var model = scope.stdmodel;
 
-          element.find(".ui-icon-gripsmall-diagonal-se")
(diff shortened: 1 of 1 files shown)


scottdraves:
1) should allow dragging in X and Y
2) show a handle to drag in the lower-right corner like this one:
screen shot 2015-08-05 at 10 22 28 am
3) I do this to the chart in the LoD tutorial, it leaves a trail:
screen shot 2015-08-05 at 10 19 45 am

kpaharelau:
1) should allow dragging in X and Y (fix)
2) show a handle to drag in the lower-right corner like this one:
(fix)
3) I do this to the chart in the LoD tutorial, it leaves a trail: - this problem was fixed by Marina

scottdraves:
:+1:

Issue #2022, opened 2015-07-31
plot typefaces


scottdraves:
the typeface for plot tick labels should be pt-sans, and it should be size 13px (same as the main body font for markdown cells).  axis labels should also be pt-sans, at 15px.

PR #2045, opened 2015-08-05
plot typefaces #2022

mnavasiolava:
change the font size and the typeface for plot tick labes and axis labels


scottdraves:
:+1:

PR #2056, opened 2015-08-06
EasyForm: api to set width and height for textarea

maxyeg:


scottdraves:
requested 10x10 but got 12x10:
screen shot 2015-08-06 at 9 35 44 am

scottdraves:
:+1:

Issue #898, opened 2014-11-20
"Level-of-Detail (LOD) is enabled" banner gets annoying


scottdraves:
every time you evaluate a cell that results in a plot with LoD, you get a fresh warning banner.
we should show it at most once, or have some clear but unobtrusive indicator for it.

scottdraves:
other LoD improvements:
increase the LoD threshold 3x.
should allow control of threshold for enabling LoD.
should allow API control of display method (avg, river, box)

PR #2060, opened 2015-08-06
fix "Level-of-Detail (LOD) is enabled" banner gets annoying #898

kpaharelau:
The LoD indicator has been removed (the legend is enough).

The default  LoD threshold has been increased 3x .

LoD threshold now can be changed in the breaker.pref.json (for example "lod-threshold": 4500) (I can add API to XYChart. Is it necessary?)

API control of display method (avg, river, box and etc.) has been added in all XYGraphics successors.
Filter getLodFilter()
void setLodFilter(Filter lodFilter)

### DIFF
--- a/core/src/main/web/app/mainapp/components/notebook/notebook-directive.js
+++ b/core/src/main/web/app/mainapp/components/notebook/notebook-directive.js
@@ -86,6 +86,12 @@
             },
             isDebugging: function () {
               return this._debugging;
+            },
+            getLodThreshold: function () {
+              return this._lodThreshold;
+            },
+            setLodThreshold: function (lodThreshold) {
+               this._lodThreshold = lodThreshold;
             }
           },
           getViewModel: function () {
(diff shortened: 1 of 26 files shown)


scottdraves:
1) Sorry there was a mistake in the specification of this issue, my mistake.
Please return the default LoD threshold back to 1500.

2) The constants that should change are lodSteps, which has different values for each lodType.  Please substitute these values:
line 1
river 3
area 1
bar 3
point 3
(leave the others unchanged).

3) yes add lodThreshold API to XYChart

4) please provide demo for setLodFilter

kpaharelau:
fix "Level-of-Detail (LOD) is enabled" banner gets annoying #898
1) Sorry there was a mistake in the specification of this issue, my mistake.
Please return the default LoD threshold back to 1500. FIXED

2) The constants that should change are lodSteps, which has different values for each lodType. Please substitute these values:
line 1
river 3
area 1
bar 3
point 3
(leave the others unchanged). FIXED

3) yes add lodThreshold API to XYChart FIXED

4) please provide demo for setLodFilter FIXED (download from https://www.dropbox.com/s/6x1va1z47fyowsj/898.bkr?dl=0)

scottdraves:
:+1:

PR #2062, opened 2015-08-06
Vertically center text cell output.

samccone:


scottdraves:
this doesn't seem right, a line of text is a little bit taller than a regular result now.

samccone:
screen shot 2015-08-06 at 4 34 17 pm
screen shot 2015-08-06 at 4 33 56 pm

Issue #2029, opened 2015-08-03
plots need more lines


scottdraves:
about 2x the current number.

scottdraves:
https://github.com/twosigma/beaker-notebook/pull/2075

PR #2075, opened 2015-08-10
fix plots need more lines #2029

kpaharelau:
Algorithm for calculation time increment has been fully changed. Use notebook for tests - https://www.dropbox.com/s/xwiz4max5htq83e/2029.bkr?dl=0


kpaharelau:
I went for lunch and found that the new algorithm is nonsense :-( . I will change it Please not review

kpaharelau:
I made predefined time values more detailed. It should solve all problems with gridlines

kpaharelau:
This is a correct link for the test notebook 
 https://www.dropbox.com/s/xwiz4max5htq83e/2029.bkr?dl=0

scottdraves:
Thanks.
This is a step in the right direction.  But still, the lines are never closer than about 104 pixels.
can you make them go down to 50?

scottdraves:
:+1:

Issue #2077, opened 2015-08-10
Plot can crash browser when output area is hidden


ildipo:
create a notebook with groovy
create a cell containing:

``` groovy
def a = new Plot()
a
```

execute
hide the output area
execute again
-> the browser is dead. :skull:

ildipo:
verified on Chrome Version 44.0.2403.130 (64-bit)

samccone:
I am looking into this now thanks for the find @ildipo

PR #2082, opened 2015-08-10
📠Add graph output smoke tests

samccone:
Will prevent regressions around #2077


samccone:
once tests :green_apple: then :ship:

Issue #1617, opened 2015-05-15
more plot legend improvements


scottdraves:
1) option to layout horizontally instead of stacking vertically
2) option to omit checkboxes
3) on mouseover highlight the line
4) cursor should be hand for dragging not I-bar for editing

scottdraves:
https://github.com/twosigma/beaker-notebook/pull/2087

PR #2087, opened 2015-08-11
more plot legend improvements #1617

mnavasiolava:
1) option to layout horizontally instead of stacking vertically: legendLayout property (LegendLayout.HORIZONTAL/VERTICAL) was added to API
2) option to omit checkboxes: omitCheckboxes property (true/false) was added to API
3) cursor should be hand for dragging not I-bar for editing

please find usage examples in https://www.dropbox.com/s/vlzhwr2ifgcuodv/1617.bkr?dl=0


scottdraves:
:+1:

PR #2088, opened 2015-08-11
Fix the publication api route needed for signing in

grin:


ildipo:
@grin I cannot even login in the pub server website

ildipo:
see pub.beakernotebook.com

grin:
ok, seems like there are some more issues...looking now.

ildipo:
also this is not enough:

(index):1 XMLHttpRequest cannot load https://pub.beakernotebook.com/user/v1/sessions. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://127.0.0.1:8801' is therefore not allowed access. The response had HTTP status code 500.

grin:
@ildipo I would try this together with the other env var fix for which I opened another PR. It looks like the `No 'Access-Control-Allow-Origin` is actually due to an error response obscuring CORS headers

PR #2090, opened 2015-08-11
Add http_host variable to fix redirects when nginx is not reached on …

ernetas:
…the same port it listens on.

This fixes incorrect 301 redirects. nginx uses its configuration to respond with full `Location` header in 301 response. However, the port in configuration does not always match the port in $http_host, especially when Beaker is behind NAT.

### DIFF
--- a/core/nginx/nginx.conf.template
+++ b/core/nginx/nginx.conf.template
@@ -61,10 +61,10 @@ http {
 
     # redirect to the starting page
     location = / {
-      return 301 %(start_page)s;
+      return 301 $scheme://$http_host%(start_page)s;
     }
     location = /beaker {
-      return 301 %(start_page)s;
+      return 301 $scheme://$http_host%(start_page)s;
     }
 
     # user directory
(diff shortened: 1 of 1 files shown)


scottdraves:
Thank you Ernetas, can you explain more about what situation this is good for?
Also though I would love to merge this, to do that I need a Contributer License Agreement.
See http://beakernotebook.com/community
Thanks, -Scott

ernetas:
E-mailed the signed version of the CLA agreement to the beaker-cla@twosigma.com address.

Essentially, if you're running Beaker behind NAT with a different external port than internal, you won't get it working with the default nginx config, because nginx uses only `server_name` and port from `listen` (if not `80`) to respond with `Location` header in 301 reponses.

Steps to reproduce the issue:

```
docker run -p 8800 -t beakernotebook/beaker
```

Try to connect to Beaker on the ephemeral port that Docker assigned to it and it will fail, as nginx will be issuing redirects to port 8800.

scottdraves:
Thanks!

That sounds reasonable, let me double check it...

ildipo:
:+1:

Issue #2093, opened 2015-08-12
cells stuck running


scottdraves:
a user reports:

> I am working with IPython code cells and  found some strange behavior. I ran my code cell (it worked fine earlier), code cell starts running.... 1 minute, 5 minutes. Cannot rerun, cannot cancel. Even if I restart Beaker and load my notebook again, this code cell is still in running(or cancelled) state. Tried to delete nginx cache, but still in cancelled... state. This situation occured a few times.
> So I cannot stop code running and cannot create new IPython cell. Only after restarting the app i can add python code cells.

reply:

> i am not sure why the original cell was not killable
> the Beaker bug is that the state of "running" can be saved in a file even though it makes no sense
> ignore what you see, you should be able to run it anyway

the fix: when loading notebook, any cells that are "in progress" should be changed to an
error state which was just "interrupted".

PR #2096, opened 2015-08-12
2093. Cells stuck running

maxyeg:
Fix for issue #2093

### DIFF
--- a/core/src/main/web/app/mainapp/mainapp.js
+++ b/core/src/main/web/app/mainapp/mainapp.js
@@ -474,7 +474,6 @@
             var deferred = bkUtils.newDeferred();
             var fileSaver = bkCoreManager.getFileSaver(uriType);
             bkSessionManager.dumpDisplayStatus();
-            bkSessionManager.clearEvaluatingCells();
             $timeout(function() {
               var content = bkSessionManager.getSaveData().notebookModelAsString;
               return fileSaver.save(uri, content, true);}, 1).then(function() {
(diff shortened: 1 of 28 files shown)


ildipo:
Hi Maxy, I think this should be done when you save the notebook: do not save cells 'under evaluation'

scottdraves:
sorry maxy, davide is right, but then the error message saved should be a different from a normal "interrupted" state, like "interrupted, unavailable at save time".

core/src/main/web/app/mainapp/mainapp.js
```diff
@@ -208,6 +208,19 @@
             showLoadingStatusMessage("Loading notebook");
             $scope.loading = true;
 
+            //Clear previous evaluation in progress state.
+            //Set state to Error with message "Interrupted."
+            if (notebookModel && notebookModel.cells) {
```
samccone: Thoughts on moving this into the`bk.notebookCellModelManager` helper?

core/src/main/web/app/mainapp/services/notebookcellmodelmanager.js
```diff
@@ -453,6 +453,18 @@
...
         }
       },
+      clearEvaluatingCells: function () {
+        if (cells) {
+          for (var i = 0; i < cells.length; i++) {
+            var currentCell = cells[i];
+            if (currentCell && currentCell.output && currentCell.output.result
+                && currentCell.output.result.innertype === 'Progress') {
+              currentCell.output.result.innertype = 'Error';
+              currentCell.output.result.object = 'Interrupted.'
```
scottdraves: error message should read:

```
Interrupted, saved while running
```

scottdraves:
This isn't quite right because if a cell is running, and you save the notebook, it actually stops the cell,
and puts the "Interrupted" error state in the notebook.

It should only change the model of the notebook that is being saved to disk, it should not alter the current model in RAM.

maxyeg:
Please, review only the last commit: https://github.com/twosigma/beaker-notebook/commit/d4b751247384b3f38dce4ad6c9fc109ce2af4fab
Rebasing against master brought many new commits to this branch :(
Please, tell me if it'll be better to create new PR.

scottdraves:
yes new PR please

PR #2101, opened 2015-08-12
more plot legend improvements #1617 (lines highlighting)

mnavasiolava:
Line highlighting on legend mouseover was added
To highlight element increase its size (+2)
Works for elements with LOD as well
Please find an example here https://www.dropbox.com/s/q2l5xdqppf3twaw/1617_highlight.bkr?dl=0

### DIFF
--- a/core/src/main/web/outputdisplay/bko-plot/bko-plot.js
+++ b/core/src/main/web/outputdisplay/bko-plot/bko-plot.js
@@ -1663,11 +1663,12 @@
         scope.$watch('getCellModel()', function() {
           scope.init();
         });
-        
+
         scope.$on('$destroy', function() {
           scope.setDumpState(scope.dumpState());
           $(window).off('resize',scope.resizeFunction);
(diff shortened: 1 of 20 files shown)


scottdraves:
:+1:

core/src/main/web/outputdisplay/bko-plot/bko-plot.js
```diff
@@ -841,7 +841,13 @@
             if (line.legend == null || line.legend === "") { continue; }
             var unit = $(legendLineUnit).appendTo(legend)
               .attr("id", "legend_" + id)
-              .addClass("plot-legendline");
+              .addClass("plot-legendline")
+              .mouseenter(function(e){
```
samccone, line 845: I think we need to make sure we clean up these event listeners, otherwise we are going to leak

scottdraves, line 845: that's right

scottdraves:
:+1:

PR #2114, opened 2015-08-13
Maxyeg/2093

maxyeg:
PR for #2093 Cells stuck running.
Save running cells with state='Error', message='Interrupted, saved while running.'
Do not change current notebook model.
Code moved from bk.notebookCellModelManager because I consider it's not a part of cell model management, it's a part of generation data for save.


ildipo:
:+1:

scottdraves:
:+1:

Issue #2081, opened 2015-08-10
Factor repeated Date into the axis label


scottdraves:
For a plot that covers a small interval in time, each line is labeled with a date but many of them are the same:
![screen shot 2015-08-10 at 1 25 21 pm](https://cloud.githubusercontent.com/assets/963093/9178067/4da8c862-3f63-11e5-9188-caa642cd94f5.png)
So "Jan 02 Fri" is the same for all the lines, so it should go into the axis label.  Only the times should remain on each line.  and where did the year go?  it should be in the axis label too.

Issue #1454, opened 2015-04-13
TimePlot tz problems


scottdraves:
1) the default should be UTC instead of EST
https://github.com/twosigma/beaker-notebook/blob/master/core/src/main/web/outputdisplay/bko-plot/plotaxis.js#L25

2) if you do set it, then it is ignored. perhaps there is a typo in the name of the timezone field
https://github.com/twosigma/beaker-notebook/blob/master/core/src/main/web/outputdisplay/bko-plot/plotformatter.js#L42
vs
https://github.com/twosigma/beaker-notebook/blob/master/plugin/jvm/src/main/java/com/twosigma/beaker/chart/serializer/XYChartSerializer.java#L86
?

PR #2130, opened 2015-08-14
fix TimePlot tz problems #1454

kpaharelau:
Default timezone has been changed to the 'UTC'. Typo in the XYChartSerializer has been fixed. Also algorithm for an extraction common part from time values has been improved (see https://github.com/twosigma/beaker-notebook/issues/2081).

Notebook for tests https://www.dropbox.com/s/sipau5hou3hlywx/1454.bkr?dl=0


scottdraves:
:+1:

PR #2135, opened 2015-08-14
changed docker build to make it faster

ildipo:
or rather split it in half so we can 'develop' faster


ildipo:
or rather split it in half so we can 'develop' faster

PR #2141, opened 2015-08-14
allow autocomplete choices window to grow as wide as needed.  it shou…

scottdraves:
…ld max out at the screen width and then scroll, this remains todo


samccone:
:ship:

PR #2147, opened 2015-08-16
Temporary fix for install-dependencies.sh for the Big Split

gef756:
Using a straight pip install ipython now downloads ipython 4. However, the ipython notebook has been spun off as jupyter starting in 4, causing some errors in the start up script (ImportError: No module named notebook.notebookapp). This can be seen by running `./beaker.command` or by running `ipython notebook`.

This fix forces pip to install version 3.2.1, where the command still works, until Beaker supports "The Big Split".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_script:
   - sudo apt-get install nginx
   - sudo apt-get install libzmq3-dbg libzmq3-dev libzmq3
   - sudo apt-get install python-pip python-dev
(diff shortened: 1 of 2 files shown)


scottdraves:
Thank you Gabe.  Indeed it looks like some work is required for 4.0 so we'll need this (and also in the Dockerfile) until it's done.  Can you please send a contributor's agreement (http://beakernotebook.com/community)?  Thanks.

scottdraves:
CLA received thank you Gabe.

Issue #1461, opened 2015-04-13
plots should support log axes


scottdraves:
def plot = new Plot(logY: true) // also X

PR #2148, opened 2015-08-17
plots should support log axes #1461

mnavasiolava:
Log y-axis was fixed. Logbase for x/y-axes was added to API. Example was added to tutorial


scottdraves:
1) `f(x) = log(x)` should be `f(x) = exp(x)`
2) `Liner` should be `Linear`
3) use like 100 points instead of 4, so you can see a smooth curve
4) add a log-log example (both X and Y log axes)

scottdraves:
screen shot 2015-08-17 at 12 24 53 pm

show regular numbers for the labels

mnavasiolava:
when we have about 100 points in exp(x) plot the axis labels look like
image

but I think this is another issue, right?

scottdraves:
Yea in general there is a problem with the formatting of the numbers can be too large.
We will file an issue for it later.

scottdraves:
:+1:

Issue #2085, opened 2015-08-11
control-click from start page should open in new window


scottdraves:
![screen shot 2015-08-11 at 10 58 22 am](https://cloud.githubusercontent.com/assets/963093/9200862/ea7d847a-4017-11e5-8afb-5704637ed0cf.png)

if you click on a notebook listed on the start page, and control or command key is down, then it should open in a new tab, like a regular html link.

PR #2155, opened 2015-08-17
#2085 control-click from start page should open in new window

dlukyanyuk:


samccone:
Can we just use an anchor tag here? Feels like we are reinventing the wheel

scottdraves:
anchor tag sounds good, is that possible dmitry?

scottdraves:
our style guide says no tabs, and indentation of JS is two spaces.
https://github.com/twosigma/beaker-notebook/wiki/Style-Guide

core/src/main/web/app/styles/dashboard.scss
```diff
@@ -81,4 +81,9 @@
       color: $green;
     }
   }
-}
+
+  .notebook-link {
+    display: block;
+    color: #333333;
+  }
+}
```
scottdraves: end files with newline

core/src/main/web/app/controlpanel/table.jst.html
```diff
@@ -28,4 +28,4 @@
       {{session.openedDate | date:'medium'}}
     </div>
   </li>
-</ul>
+</ul>
```
scottdraves: end files with newline

scottdraves:
:+1:

Issue #2127, opened 2015-08-13
console error on page init


scottdraves:
https://pub.beakernotebook.com/user/v1/current_user Failed to load resource: the server responded with a status of 404 (Not Found)

grin:
Just for the record, this error is not really an error. That request is used to probe the api to verify whether a user is currently signed in...that being said, I can make it so it doesn't show up the error on the javascript console if that's the intent of this issue.

ildipo:
yes please

PR #2165, opened 2015-08-19
Silence 4XX errors when probing for user session

grin:
With this change the 404 request will be logged to dev console when the publish modal is opened and not on Beaker start-up.

Fixes https://github.com/twosigma/beaker-notebook/issues/2127

### DIFF
--- a/core/src/main/web/app/beaker.js
+++ b/core/src/main/web/app/beaker.js
@@ -411,10 +411,6 @@
         };
       });
     });
-
-    beaker.run(function(bkPublicationAuth) {
-      return bkPublicationAuth.initSession();
-    });
   };
   var bootstrapBkApp = function() {
     // make sure requirejs reports error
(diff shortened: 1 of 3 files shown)


grin:
@ildipo turns out you can't silence the 404 in Chrome by try/catching it...it's logged natively. As a work-around I moved the check to a different spot, so it will only show up once to users who clicked the publish option. Let me know if it's :+1:  or :-1: 
Alternatively, I can introduce a new API endpoint that always responds with 2XX code and change Beaker to use that new endpoint.

ildipo:
I think this is fine - can we change the error code to 401?

scottdraves:
Thanks for moving this, that helps and is definitely right.
But I still don't like even a 401 on the console.
Can you please have the server respond with a 200 and false?
The reason is when there is a problem users will open the console and see this error and send it to us, instead of looking further to find the real problem.

grin:
Ok, this is now updated. No more 4XXs will show up in the console.

Issue #2154, opened 2015-08-17
combined plot Y label truncated


scottdraves:
the 1st example in the plotting tutorial, the Y axis label is truncated.

<img width="121" alt="screen shot 2015-08-17 at 10 46 00 am" src="https://cloud.githubusercontent.com/assets/963093/9307634/97e6d990-44cd-11e5-9fee-466ca90dc309.png">

PR #2166, opened 2015-08-19
combined plot Y label truncated #2154

mnavasiolava:
If gridline labels are shown, increase topLayoutMargin by the half of the labels height


scottdraves:
:+1: CI problem seems unrelated

PR #2168, opened 2015-08-19
menus: Only show custom content region if content.

samccone:
cc @ildipo


ildipo:
thanks Sam, just waiting on CI

samccone:
:green_apple:

Issue #2010, opened 2015-07-30
calling exit() in Python hangs


scottdraves:
if the Python kernel exits for any reason, it gets restarted and the following messages appear on the console:

```
IPython-stderr>[I 12:43:39.798 NotebookApp] KernelRestarter: restarting kernel (1/5)
IPython-stderr>WARNING:root:kernel 0c36cc0e-a827-4b82-b32f-eadee2e7607b restarted
```

but the JS plugin never gets the memo and is left in a confused state where it thinks the cell is executing and it can't be interrupted.

it should just warn that user of the restart and clean up its state.

other languages may have a similar problem.

dlukyanyuk:
Are there any steps to reproduce this issue?

dlukyanyuk:
@scottdraves , what we need to see as result of 'exit()' function calling (in addition to the normal cell execution)?

scottdraves:
backend should restart. try in ipython to see it work correctly.

dlukyanyuk:
What about 

> it should just warn that user of the restart and clean up its state.

Should we print some some message like 'Kernel is restarted' in the output cell or in a popup?

I fixed the error and now it is:

Cell execution result:

![cell](https://cloud.githubusercontent.com/assets/3629655/9359755/54e47d9e-469b-11e5-84fe-4cf440ccfbdd.png)

and console:
![console](https://cloud.githubusercontent.com/assets/3629655/9359858/bf3ea3ea-469b-11e5-9614-9dc9f134bfff.png)

Is this OK?

scottdraves:
great.  popup please.

PR #2174, opened 2015-08-20
calling exit() in Python hangs #2010

dlukyanyuk:


scottdraves:
this looks good.  please apply to julia.js and iruby.js as well.

also, change the message from 'Kernel restart completed' to 'Kernel exited, restart completed'

scottdraves:
:+1:

Issue #2157, opened 2015-08-17
plots rendered inside a closed section have size 0


ildipo:
load the test/notebooks/jvm-autotranslation-test notebook, wait for the initialization window to disappear then expand the section and scroll down.

PR #2183, opened 2015-08-24
plots rendered inside a closed section have size 0 #2157

mnavasiolava:
scope.container.node().getBoundingClientRect().width/height gives 0 when section is collapsed, use jQuery width()/height() instead


scottdraves:
:+1:

ildipo:
:+1:

Issue #2181, opened 2015-08-24
shift-control/command-enter in a code cell


scottdraves:
should run current code selection, or current line if none (and advance to the next line).
useful for debugging, this should be familiar to matlab and rstudio users.
http://opiateforthemass.es/articles/why-i-dont-like-jupyter-fka-ipython-notebook/

PR #2206, opened 2015-08-26
shift-control/command-enter in a code cell #2181

mnavasiolava:
evaluate selection or current line (if no selection) in a code cell

### DIFF
--- a/core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
+++ b/core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
@@ -335,6 +335,12 @@
           'Cmd-Alt-H': function(cm) { // cell hide
             scope.cellmodel.input.hidden = true;
             bkUtils.refreshRootScope();
+          },
+          'Shift-Ctrl': function(cm) {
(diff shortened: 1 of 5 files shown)


core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
```diff
@@ -420,6 +426,32 @@
           }
         });
 
+        scope.evaluateSelection = function(cm) {
+          var evalCode;
+          var currentLine;
+          if(cm.somethingSelected()){
```
scottdraves, line 432: official style is to have a space between "if" and "(" and between ")" and "{".
also should use "} else {" etc.
you don't have to fix this PR, it's not that important, but please do it this way going forward.

scottdraves:
:+1:

core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
mnavasiolava, line 432: ok, i'll do it

Issue #2203, opened 2015-08-25
show spinner during switch


scottdraves:
If you use the start page to switch between notebooks, you just see a new noteobook with the insert buttons while it is opening.  instead you should get a spinner.  also it says "New Notebook" instead of the name that it is opening, it should update this first.

![screen shot 2015-08-25 at 1 19 06 pm](https://cloud.githubusercontent.com/assets/963093/9474054/3d082cb2-4b2c-11e5-9593-457a267ce49c.png)

to reproduce:
run beaker.
open tutorial via the Help menu.
click logo to return to start page
click the tutorial in the list of notebooks.
while the tutorial is reopening, you see the buttons (bug), "New Notebook" (bug) and no spinner (bug).

PR #2216, opened 2015-08-27
show spinner during switch #2203

mnavasiolava:
The problem was in angular binding. If variables are assigned in $http callback angular'll invoke $digest (which binds the variables) only after the whole callback is executed.
In our case variable for notebook’s title and flag to show spinner were bound only after the notebook had been loaded.
To enforce angular $digest I firstly update notebook’s name and spinner flag and then call method for loading notebook in timeout function.


scottdraves:
:+1:

ildipo:
:+1:

Issue #2215, opened 2015-08-27
easyform crashes pub server


scottdraves:
https://pub.beakernotebook.com/#/publications/55de588f-98a4-471d-8cba-0eb880a141e7
`childScope.evaluatorId = $scope.model.getEvaluatorId();`
does not work

PR #2224, opened 2015-08-28
#2215. easyform crashes pub server

maxyeg:
EvaluatorExist flag added


ildipo:
This does work - eazyforms show up but the style sheet is wrong.
I'll create a new ticket and look into it since it is probably namespacing so related to the pub server itself.

Issue #2209, opened 2015-08-26
language popup position broken


scottdraves:
in both regular and advanced mode the language popup menu is positioned so that if you
click twice, you end up selecting the 1st menu item.

ideally it should be positioned so that a 2nd click just closes the menu with no change (like it selects current one).  the badge that pops up should be exactly positioned over the badge on the cell.

if that is difficult,  the menu should be moved down to be completely below the button that brings it up,
so a 2nd click just hides it.

PR #2240, opened 2015-08-31
fix language popup vertical position broken #2209

kpaharelau:
Menu has been moved down completely below the button. Second click hides menu.


scottdraves:
:+1:

Issue #2234, opened 2015-08-29
unset easyform field should be "" not null


scottdraves:
<img width="318" alt="screen shot 2015-08-29 at 11 56 27 am" src="https://cloud.githubusercontent.com/assets/963093/9562823/863852e2-4e45-11e5-84ce-e01cfb3e91cb.png">
https://pub.beakernotebook.com/#/publications/55e1d774-224c-4651-8333-5e4e4f11da74

PR #2241, opened 2015-08-31
#2234. unset easyform field should be "" not null

maxyeg:
Change EasyFormComponent 'value' field type to be String.
Set empty string value in super-component's constructor.


ildipo:
:+1:

scottdraves:
:+1:

Issue #2229, opened 2015-08-29
Combo Boxes and Editable Combo Boxes not updating


scottdraves:
changing the value is not sent back to the form on the server.

PR #2245, opened 2015-08-31
#2229. Combo Boxes and Editable Combo Boxes not updating

maxyeg:
Explicit call to update model value for custom.combobox jquery widget


scottdraves:
:+1:

Issue #2197, opened 2015-08-25
create new account flow error


ildipo:
1) start beaker and create/open a notebook
2) click publish (while not logged in)
3) click create new account
4) create the account
5) go back to beaker
6) click publish again -> beaker should be already logged in

grin:
implemented in: https://github.com/twosigma/beaker-notebook/pull/2253

ildipo:
@grin: This does not work very well. I think you should modify it.

The current behavior is:
1) put the login popup
2) click on 'new account' -> open it in a new window
3) in the new window you register and press ok
4) the new window tab is redirected to beaker
you endup with two beaker tabs pointing to the same notebook

The proper behavior should be:
1) put the login popup
2) click on 'new account' -> open it in a new window and close the popup in beaker
3) in the new window you register and press ok
4) close the new window tab

grin:
how about if in the step 2 I open the publications sign up in the current beaker window...and then at the end redirect it back to beaker ?

grin:
actually nevermind, I can implement the proper flow quite easily

ildipo:
the issue is that if you use the same tab you'll have first to backup the current session and reload it when you came back

ildipo:
I verified that if you simply close the new tab the old one is logged in properly

grin:
Fixed in: https://github.com/twosigma/beaker-notebook/pull/2275

ildipo:
merged

PR #2253, opened 2015-08-31
After successful sign up to Publications pass the auth token to Beaker

grin:
via redirection back to Beaker with the auth token passed as a query param.

Fixes: https://github.com/twosigma/beaker-notebook/issues/2197
Depends on: https://github.com/twosigma/bunsen/pull/1098


ildipo:
:+1:  I'll merge this after I merge and deploy bunsen side

Issue #798, opened 2014-10-23
call updateDocumentModelFromDOM  automatically on save/share


scottdraves:
and allow it to apply to the whole document, not just one cell.

PR #2262, opened 2015-09-01
call updateDocumentModelFromDOM automatically on save/share #798

mnavasiolava:
updateCellsFromDOM method was added to one or more code cells.
this method is called on save/share.


scottdraves:
:+1:

ildipo:
:+1:

Issue #2230, opened 2015-08-29
insert button bars should fade in/out


scottdraves:
<img width="266" alt="screen shot 2015-08-28 at 11 54 03 pm" src="https://cloud.githubusercontent.com/assets/963093/9560401/3e51f830-4de0-11e5-8808-d8b7d5e5496d.png">
<img width="359" alt="screen shot 2015-08-28 at 11 54 18 pm" src="https://cloud.githubusercontent.com/assets/963093/9560402/3e56ccca-4de0-11e5-997d-8cb59c8e3056.png">

on the mouse-over, ~ 0.3s duration.

PR #2271, opened 2015-09-02
insert button bars should fade in/out #2230

mnavasiolava:
use opacity property to show/hide button bars with transition = 0.3s


scottdraves:
:+1:

Issue #2143, opened 2015-08-14
runt column in data frame UI


scottdraves:
when you sort a dataframe, sometimes a runt column appears on the left.
![screen shot 2015-08-14 at 5](https://cloud.githubusercontent.com/assets/963093/9284378/5f82bdba-42a9-11e5-94a8-dc64b301c400.png)
repro:
1) start with ipython examples in the tutorial
2) remove ".head()", run the cell
3) click on "Open"

PR #2272, opened 2015-09-02
fix runt column in data frame UI #2143

kpaharelau:
Problem was in a fixedcolumns plugin behaviour. I has called fnRedrawLayout function for an adjust layout


ildipo:
:+1:

scottdraves:
:+1:

Issue #2232, opened 2015-08-29
mouseover on legend needs delay & fade


scottdraves:
<img width="227" alt="screen shot 2015-08-29 at 12 50 05 am" src="https://cloud.githubusercontent.com/assets/963093/9560588/624e05bc-4dea-11e5-80a1-40f4c465ecb1.png">
so if the mouse moves over it quickly it should not highlight the line.
only if you move slowly or stop the mouse over the legend, should the line react.

PR #2273, opened 2015-09-02
mouseover on legend needs delay & fade #2232

mnavasiolava:
lines will only be highlighted if the mouse is over the ligend's item is more than 0.3s
also duration was added to highlighting (0.1s)


scottdraves:
:+1:

Issue #2267, opened 2015-09-01
space in windows username breaks ipython


scottdraves:
```
starting nginx instance (C:\Users\With Space\Downloads\bn/nginx)
[main] INFO org.eclipse.jetty.server.Server - jetty-8.1.13.v20130916
[main] INFO org.eclipse.jetty.server.AbstractConnector - Started SelectChannelConnector@127.0.0.1:8802

Connecting to http://127.0.0.1:8801/

[qtp1297836716-20] WARN org.eclipse.jetty.servlet.ServletHandler -
javax.servlet.ServletException: java.io.FileNotFoundException: C:\Users\With Space\.beaker\v1\nginx2993394519491328222\p
rofile_beaker_backend_IPython\ipython_notebook_config.py (The system cannot find the path specified)
```

scottdraves:
probably missing quotes somewhere in pythonBasePath or hashIPythonPassword or generateIPythonConfig in core/src/main/java/com/twosigma/beaker/core/rest/PluginServiceLocatorRest.java

PR #2281, opened 2015-09-03
space in windows username breaks ipython #2267

mnavasiolava:
added quotes in nginx path for windows


scottdraves:
please remove conditional, this should apply to all platforms

scottdraves:
:+1:

Issue #2279, opened 2015-09-03
rename plot log axis properties


scottdraves:
they should be 
xLogBase and 
yLogBase

PR #2282, opened 2015-09-03
rename plot log axis properties #2279

mnavasiolava:
plot log axis properties were renamed, tutorial was updated


scottdraves:
:+1:

Issue #2231, opened 2015-08-29
h scrollbar on markdown cells is broken


scottdraves:
if you click on it, it vanishes and the cell formats.

PR #2286, opened 2015-09-03
fix h scrollbar on markdown cells is broken #2231

kpaharelau:
I found that problem exists only for chrome browser.OnBlur event for a codemirror editor is triggered after click on a h-scrollbar after notebook vertical scrolling. But without vertical scrolling all ok. I have used  scrollbarStyle from codemirror 'SimpleScrollbar' addon instead native scrollbars


scottdraves:
:+1:

ildipo:
:+1:

Issue #2255, opened 2015-09-01
default language for empty notebook should be JavaScript


scottdraves:
<img width="340" alt="screen shot 2015-08-31 at 10 07 39 pm" src="https://cloud.githubusercontent.com/assets/963093/9594730/e41b5812-502c-11e5-8093-cc3c8f4128ab.png">

scottdraves:
if JavaScript plugin does not exist, then can use anything, like the 1st in the list.

PR #2293, opened 2015-09-04
default language for empty notebook should be JavaScript #2255. error fix

mnavasiolava:
by mistake i used .keys function in a wrong way


ildipo:
:+1:

Issue #2261, opened 2015-09-01
codemirror should unhighlight on blur


scottdraves:
codemirror turns matching parens bright green.  but if you switch to another codemirror or just click on the background so the cm blurs, it remains highlighted.  seems that if you click enough times, it disables.  i have also noticed a bit of jankiness in the editor, where some clicks or arrow keystrokes are ignored.  this is very disturbing.

seems like a regression but i have not fully researched this yet.

PR #2299, opened 2015-09-07
fix codemirror should unhighlight on blur #2261

kpaharelau:
I found that highlighting not working if selection in an editor is not changed. I has added event listeners on "blur" and "focus" for a solving this problem.
On "blur" selection is changed and  "cursorActivity" event fired. On "focus" only "cursorActivity" event fired.

Notebook for tests - https://www.dropbox.com/s/lys4nixenlzaci9/2261.bkr?dl=0


scottdraves:
:+1:

Issue #2290, opened 2015-09-03
replace resize thumb


scottdraves:
![screen shot 2015-09-03 at 5 43 15 pm](https://cloud.githubusercontent.com/assets/963093/9671814/8e74b8f2-5263-11e5-9ef7-30e94544d883.png)

our thumb is a low-resolution bitmap.  is it part of the SVG?  then it can easily be drawn with 3 lines at high resolution.  if outside then let's discuss options.

PR #2303, opened 2015-09-08
fix replace resize thumb #2290

kpaharelau:
resize thumb has been replaced ('ui-icon-grip-diagonal-se' instead 'ui-icon-gripsmall-diagonal-se' from jquery-ui package).

ui-icon-gripsmall-diagonal-se
image

ui-icon-grip-diagonal-se
image


scottdraves:
:+1:

Issue #2300, opened 2015-09-08
right arrow on last position of cell


scottdraves:
should move to first position of next cell, similar to down arrow.
similarly, left arrow on first char should move to previous cell.

PR #2309, opened 2015-09-09
fix right arrow on last position of cell #2300

kpaharelau:
goCharRightOrMoveFocusDown and goCharLeftOrMoveFocusDown functions has been added to a core.js

Use Interactive charts tutorial for a test


scottdraves:
:+1:

Issue #2304, opened 2015-09-08
charts width can differ after rerun


scottdraves:
![screen shot 2015-09-08 at 10 55 36 am](https://cloud.githubusercontent.com/assets/963093/9738159/4479d828-5618-11e5-9d9c-a9a42d1856dc.png)

if you drag to change the size of the chart and the rerun the cell, the widths of the charts differ.

PR #2310, opened 2015-09-09
fix charts width can differ after rerun #2304

kpaharelau:
'watchCellSize' has been improved in a bko-plot.js. scope.model.isShowOutput always undefined. Treatment this has been added


scottdraves:
:+1:

Issue #2312, opened 2015-09-09
insert cell menus change cursor


scottdraves:
mouse should change to finger when selecting dropdown menus from "code" and "section".
![screen shot 2015-09-09 at 11 28 12 am 2](https://cloud.githubusercontent.com/assets/963093/9766324/45ed9680-56e6-11e5-9fc2-9ff4f36607a1.png)
![screen shot 2015-09-09 at 11 27 52 am 2](https://cloud.githubusercontent.com/assets/963093/9766335/5037c750-56e6-11e5-9f28-7f39564bae1d.png)

PR #2317, opened 2015-09-10
fix insert cell menus change cursor #2312

kpaharelau:
new css class new-cell-menu-item has been added


scottdraves:
cursor is still text caret on the last item "Other languages..."

scottdraves:
:+1:

Issue #2244, opened 2015-08-31
bokeh problem


scottdraves:
open the bokeh tutorial, then close it with file->close menu item.

i get this stack trace:

TypeError: Cannot read property 'unsubscribe' of undefined
    at disconnectcontrol (sessionmanager.js:850)
    at Object.clear (sessionmanager.js:923)
    at close (sessionmanager.js:940)
    at wrappedCallback (angular.js:11682)
    at angular.js:11768
    at Scope.$eval (angular.js:12811)
    at Scope.$digest (angular.js:12623)
    at Scope.$apply (angular.js:12915)
    at done (angular.js:8450)
    at completeRequest (angular.js:8664)(anonymous function) @ angular.js:10126(anonymous function) @ angular.js:7398wrappedCallback @ angular.js:11685(anonymous function) @ angular.js:11768Scope.$eval @ angular.js:12811Scope.$digest @ angular.js:12623Scope.$apply @ angular.js:12915done @ angular.js:8450completeRequest @ angular.js:8664xhr.onreadystatechange @ angular.js:8603
VM269:20 DELETE http://127.0.0.1:8801/8e780a6/python3.522082917704/api/kernels/ec7d9854-68d9-4574-9894-131829e1850f 404 (Not Found)

scottdraves:
also if i ran the cells i got some errors because it seems bokeh has changed a bit with more recent versions.  and then if i close the notebook, i get some errors that look like the underscore module being overwritten problem is back.

PR #2329, opened 2015-09-14
fix bokeh problem #2244

kpaharelau:
I found that problem is in the saved bokeh.bkr notebook. I reran output_notebook() in the python cell. BokenJS has been reloaded.  After that file was saved. The problem disappeared.
FYI
You should executet python cell one by one or  "Run all cells". Otherwise you get error


scottdraves:
:+1:

Issue #2332, opened 2015-09-14
stack trace on blur cm


scottdraves:
```
Error: [$rootScope:inprog] $apply already in progress
http://errors.angularjs.org/1.2.28/$rootScope/inprog?p0=%24apply
    at angular.js:78
    at beginPhase (angular.js:13122)
    at Scope.$apply (angular.js:12908)
    at codecell-directive.js:380
    at CodeMirror.signal (codemirror.js:8033)
    at onFocus (codemirror.js:4134)
    at HTMLTextAreaElement.<anonymous> (codemirror.js:8199)
    at TextareaInput.copyObj.focus (codemirror.js:1337)
    at CodeMirror.focus (codemirror.js:4797)
    at Scope.$scope.backgroundClick (codecell-directive.js:117)
```

repro:
1. Insert code cell.
2. Select some string in it.
3. Click below code cell's input, at the level of 'Run' button.

PR #2346, opened 2015-09-15
fix stack trace on blur cm #2332

kpaharelau:
block  scope.$apply(function ()....) has been deleted as unneeded


scottdraves:
:+1:

PR #2354, opened 2015-09-16
 Fix lines mangling in JS code.

maxyeg:
gulp-header version updated, it was fixed there in https://github.com/tracker1/gulp-header/issues/7


scottdraves:
:+1:

Issue #2353, opened 2015-09-15
cursor ends up in wrong location


scottdraves:
to reproduce:
1) create an empty notebook
2) add a groovy cell
3) type "[0, null]".
4) observe that the cursor is between the "l" and the "]", instead of at the end.

scottdraves:
![screen shot 2015-09-15 at 2 45 47 pm](https://cloud.githubusercontent.com/assets/963093/9886530/cdbc2c1c-5bb8-11e5-93b7-bb6919c0a68c.png)

PR #2366, opened 2015-09-17
fix cursor ends up in wrong location #2353

kpaharelau:
It is very strange but Robot mono font doesn't work correctly with codemirror editor. I have changed robot mono on ubuntu mono font.


core/src/main/web/app/styles/vendor_overrides.scss
```diff
@@ -26,6 +26,7 @@
 
 .CodeMirror pre {
   z-index: initial;
+  font-size: 14px;
```
scottdraves, line 29: why is this needed?

kpaharelau:
Font ubuntu-mono working good, but it looks small for default size

image

image

image

scottdraves:
OK this looks good.  Please remove the Roboto Mono font, and everywhere it appeared replace it with Ubuntu Mono.

scottdraves:
:+1:

Issue #2352, opened 2015-09-15
windows python error


scottdraves:
from the 2nd example in the python tutorial in windows:

```
C:\Users\Beaker\beaker-notebook\core\config\plugins\eval\ipythonPlugins\ipython\beaker_runtime.py in default(self, obj)
    309             out = {}
    310             out['type'] = "Date"
--> 311             out['timestamp'] = int(obj.strftime("%s"))*1000
    312             return out
    313         if type(obj) == pandas.core.frame.DataFrame:

ValueError: Invalid format string
```

scottdraves:
http://stackoverflow.com/questions/10807164/python-time-formatting-different-in-windows
cap S?

maxyeg:
It fixes #2292

Issue #2333, opened 2015-09-14
Python Dates showing up as NaN


scottdraves:
In the Python tutorial, in the table, the dates are showing up as NaN:
![screen shot 2015-09-14 at 3 38 19 pm](https://cloud.githubusercontent.com/assets/963093/9859886/b9d1cfee-5af6-11e5-983f-2eb84395b0f6.png)

scottdraves:
the same? https://github.com/twosigma/beaker-notebook/issues/2292

PR #2367, opened 2015-09-17
#2352. windows python error

maxyeg:
According to http://stackoverflow.com/questions/11743019/convert-python-datetime-to-epoch-with-strftime

Python `strftime` function doesn't support `%s` option. It works on Unix because Python `strftime` calls native `strftime` and `%s` is supported there on Unix, but not on Windows.

It's replaced with `calendar.timegm()`.

But it still doesn't fix Python Dates NaN error(issue #2333)


scottdraves:
:+1:

PR #2368, opened 2015-09-17
#2333. Python Dates showing up as NaN

maxyeg:


maxyeg:
Works fine before this commit:
https://github.com/twosigma/beaker-notebook/commit/0dbf3afc8b347b01f76c3a49f824bad0a6fd8e7e

ildipo:
Hi Maxy, actually this is not the correct fix.

The above changes (which you rollback) is used to automatically translate the data frame index between python, R and other languages.

The real issue for #2333 is that the DataTable output display does not consider the possibility of the index being a date.

If you look at the JSON that represents the table you can see:

{"columnNames":["Index","Open","High","Low","Close","Volume","Adj Close"],"subtype":"TableDisplay","values":[[{"timestamp":1262581200000,"type":"Date"},112.370003,113.389999,111.510002,113.33000200000001,118944600,101.445015],[{"timestamp":1262667600000,"type":"Date"},113.260002,113.68,112.849998,113.629997,111579900,101.713549],[{"timestamp":1262754000000,"type":"Date"},113.519997,113.989998,113.43,113.709999,116074400,101.785161],[{"timestamp":1262840400000,"type":"Date"},113.5,114.33000200000001,113.18,114.190002,131091100,102.214826],[{"timestamp":1262926800000,"type":"Date"},113.889999,114.620003,113.660004,114.57,126402800,102.554973]],"hasIndex":"true","type":"TableDisplay","types":["time","double","double","double","double","double","double"]}

Please look at the datable output display and ensure we can support index columns with any datatype, not only time or integer.
When you look at it please consider that the datable output display will use the existing index (if hasIndex: true is in the JSON) or will make its own index column.

maxyeg:
Now TableDisplay can support index column with any datatype.
But still in python script we put type definition in json like "time".

```
if typ.startswith("date") or typ.startswith("Time"):
        return "time"
```

And in javascript for "time" we show only time, not date.
Is it ok?

ildipo:
Hi Maxy, that's correct. At the moment we fully support only DateTime (which is called time).
We should really support Date, Time and DateTime for every language and output display but I didn't have the time to do it...

ildipo:
actually is called Date in JSON and time in the column times :-)

maxyeg:
Yes, you're right.
I mean maybe we should use other converter(for "date" or "datetime") in bko-tabledisplay.
Now converter for type "time" will be used.

scottdraves:
it should be datetime.

ildipo:
better create another issue to clean up the date vs time vs datetime everywhere

Issue #2313, opened 2015-09-09
NPE in CI on /rest/session-backup/setEdited


scottdraves:
```
WARN org.eclipse.jetty.servlet.ServletHandler - /rest/session-backup/setEdited
java.lang.NullPointerException
    at com.twosigma.beaker.core.rest.SessionBackupRest.setEdited(SessionBackupRest.java:207)
```

scottdraves:
related? https://github.com/twosigma/beaker-notebook/issues/1814

Issue #1814, opened 2015-06-22
NPE in session backup


scottdraves:
[qtp106931168-25] WARN org.eclipse.jetty.servlet.ServletHandler - /rest/session-backup/backup/InUWcP
java.lang.NullPointerException
        at com.twosigma.beaker.core.rest.SessionBackupRest.backup(SessionBackupRest.java:130)

scottdraves:
https://github.com/twosigma/beaker-notebook/pull/1817

PR #2369, opened 2015-09-17
fix NPE in CI on /rest/session-backup/setEdited #2313

kpaharelau:
In some cases sessionID parameters is empty. Check for this situation was added. https://github.com/twosigma/beaker-notebook/issues/1814 is not reproduced because method backup has been changed (rev 23e7617634cac1437c9c1501c49a8ead4e0eae38)


scottdraves:
can you please check the client code?
why is the param missing sometimes?

BeakerNotebook.com
ScottDraves.com
twitter.com/Scott_Draves

Issue #2380, opened 2015-09-18
Windows OS - electron application setup


kpaharelau:
runElectron and installElectron doesn't work. Add information about Inno Setup and unzip installations to the readme

PR #2381, opened 2015-09-18
fix Windows OS - electron application setup #2380

kpaharelau:
"cmd /c" prefix need to bee used for run all commands for Windows - has been added for installElectron task. Filter for path to application has been added (runElectron did not work without this filter). Information about UnZip installer for Windows OS has been added to the readme


ildipo:
:+1:

PR #2382, opened 2015-09-18
Sql plugin

maksim-lauryk:
sql plugin


ildipo:
Ah I think you changed the CR into CRLF...
Can you update this PR by using CR only please?

ildipo:
:+1:

Issue #2307, opened 2015-09-09
scroll to error cell


scottdraves:
if you "run all cells" and one of them produces an error, then scroll so the cell with the error is at the top of the window.

PR #2415, opened 2015-09-21
fix scroll to error cell #2307

kpaharelau:
First cell with error is found after evaluate all cells. We scrolls to the  cell and sets focus to it.


ildipo:
:+1:

scottdraves:
does not work for JavaScript syntax errors like
screen shot 2015-09-21 at 10 30 12 am

scottdraves:
:+1:

Issue #2386, opened 2015-09-18
copy dataframe to clipboard omits index column


scottdraves:
if it's a datetime at least you need it

PR #2417, opened 2015-09-21
#2386. copy dataframe to clipboard omits index column

maxyeg:


ildipo:
:+1:

core/src/main/web/outputdisplay/bko-tabledisplay/bko-tabledisplay.js
```diff
@@ -137,12 +137,18 @@
...
           var j;
+          var startingColumnIndex = 1;
           var order;
           var out = '';
           var eol = '\n';
           var sep = ',';
           var qot = '"';
           var fix = function(s) { return s.replace(/"/g, '""');};
+          var model = $scope.model.getCellModel();
+          var hasIndex = model.hasIndex === "true";
```
scottdraves, line 148: why is this the string true instead of the value true?
that is a bit disturbing.

scottdraves:
:+1:

core/src/main/web/outputdisplay/bko-tabledisplay/bko-tabledisplay.js
maxyeg, line 148: Should I refactor it now?
It will require changes in `beaker.r`, `beaker_runtime.py`, `beaker_runtime3.py`, `bko-tabledisplay.js` and `sessionmanager.js`.

scottdraves, line 148: thanks for explaining.
no but i will file an issue for future work.

Issue #1920, opened 2015-07-11
The input line is too long -- affects R plugin and perhaps others


aolney:
Seems to be a classic cmd bug in 1.3.1. The problem is with set CLASSPATH being too long. This problem exists for the R plugin and may exist elsewhere:

```
C:\Users\aolne_000\beaker-notebook-v1.3.1-0-g3072e5e-windows\config\plugins\eval\r>bin\r
The input line is too long.
The syntax of the command is incorrect.
C:\Users\aolne_000\beaker-notebook-v1.3.1-0-g3072e5e-windows\config\plugins\eval\r>
```

This bug prevents the R plugin from working on my machine (Win 8.1 Pro x64). The offending line is line 72. I just changed it to:

```
set CLASSPATH=%APP_HOME%\lib\*
```

which works because all the jars on the offending line were in this directory anyways. 

This change makes the R plugin work for me. It also has the possible advantage of being easier to read/maintain.

scottdraves:
Interesting, thanks for the report.  Sounds like a gradle issue as this file is generated by our java build tool.  Looks similar to https://discuss.gradle.org/t/filename-too-long-in-windows/9222
there is a suggested workaround there...

aolney:
FWIW here is a SO thread that talks about various ways of addressing this:

http://stackoverflow.com/questions/201816/how-to-set-a-long-java-classpath-in-windows

scottdraves:
https://issues.gradle.org/browse/GRADLE-2992
see the last workaround suggested here, that sounds pretty good.

PR #2419, opened 2015-09-21
#1920. The input line is too long -- affects R plugin and perhaps others

maxyeg:
CLASSPATH length violation fixed for R plugin


scottdraves:
cool.
shouldn't this be applied to all the plugins?
is there a way without copying the code?

maxyeg:
Applied to other plugins.
Need task to improve gradle build structure to avoid repeating code.

scottdraves:
:+1:

Issue #2134, opened 2015-08-14
Markdown cell does not support blockquotes.


dlukyanyuk:
See http://daringfireball.net/projects/markdown/syntax#blockquote for examples.
Now all blockquotes symbols (>) are displayed as plain text in markdown cell.

PR #2420, opened 2015-09-21
#2134. Markdown cell does not support blockquotes.

maxyeg:
Using `.html()` escapes reserved characters and turns them into character entities(like `&gt;`).
To make blockquotes working in markdown plugin I unescape `&gt;` back to `>`.


scottdraves:
cool but please do not let the text size change:
screen shot 2015-09-21 at 12 58 06 pm

scottdraves:
:+1:

Issue #2404, opened 2015-09-19
autocomplete click selection broken


scottdraves:
in python if you hit tab to bring up autocomplete, then click with the mouse to select a choice,
the menu vanishes but completion doesn't happen.  using the kbd (arrow down plus return) works fine.

PR #2428, opened 2015-09-22
fix autocomplete click selection broken #2404

kpaharelau:
Hints are considered now in the onBlur and onFocus handlers


ildipo:
:+1:

scottdraves:
:+1:

Issue #1697, opened 2015-06-02
Broken 'new folder' button on 'Save as...' dialog


DiegoAlbertoTorres:
JS console error when clicking the new folder button:

```
TypeError: bkUtils.httpPost(...).complete is not a function
    at fileChooserStrategy.newFolder (core.js:298)
    at angular.js:10903
    at angular-touch.js:441
    at Scope.$eval (angular.js:12811)
    at Scope.$apply (angular.js:12909)
    at HTMLElement.<anonymous> (angular-touch.js:440)
    at HTMLElement.jQuery.event.dispatch (jquery.js:4676)
    at HTMLElement.$event.dispatch (jquery.event.drag.js:374)
    at HTMLElement.elemData.handle (jquery.js:4360)(anonymous function) @ angular.js:10126(anonymous function) @ angular.js:7398Scope.$apply @ angular.js:12911(anonymous function) @ angular-touch.js:440jQuery.event.dispatch @ jquery.js:4676$event.dispatch @ jquery.event.drag.js:374elemData.handle @ jquery.js:4360
angular.js:8632POST http://127.0.0.1:8801/a48e9bd/beaker/rest/file-io/createDirectory 409 (Conflict)(anonymous function) @ angular.js:8632sendReq @ angular.js:8426serverRequest @ angular.js:8146wrappedCallback @ angular.js:11682wrappedCallback @ angular.js:11682(anonymous function) @ angular.js:11768Scope.$eval @ angular.js:12811Scope.$digest @ angular.js:12623Scope.$apply @ angular.js:12915(anonymous function) @ angular-touch.js:440jQuery.event.dispatch @ jquery.js:4676$event.dispatch @ jquery.event.drag.js:374elemData.handle @ jquery.js:4360
2kernel.js:605 Uncaught Error: kernel is not connected
```

DiegoAlbertoTorres:
Coming back to this issue, I don't think the button works at all :/

samccone:
@DiegoAlbertoTorres I think you fixed this one right?

scottdraves:
no this is live

PR #2430, opened 2015-09-22
fix Broken 'new folder' button on 'Save as...' dialog #1697

kpaharelau:
Problem with signature httpPost has been fixed. I see that our file save and file open dialog has very non standard view and behaviour, May be have sense use something like this https://github.com/eligrey/FileSaver.js?


ildipo:
Isn't FileSaver.js meant for 'client side' save?

kpaharelau:
We can use FileSaver.js for 'client side' save. See example
http://codepen.io/davidelrizzo/pen/cxsGb

scottdraves:
:+1:

Issue #2448, opened 2015-09-22
after leaving and returning, cursor broken


scottdraves:
1) open a new notebook
2) insert a JS cell and type a string like "asdf", leaving the cursor to the right of the "f"
3) use Alt-TAB to switch to another window
4) hit alt-tab again to switch back
5) type "x", and see that the cell now contains "xsdf" instead of "asdfx".

this is a serious regression, perhaps related to https://github.com/twosigma/beaker-notebook/pull/2299 ??

PR #2456, opened 2015-09-23
fix after leaving and returning, cursor broken #2448

kpaharelau:
this is a serious regression, perhaps related to #2299 ?? Yes. Problem has been fixed but matching brackets stays highlighted on click outside editor (for example on the area near buttons).


scottdraves:
:+1:

Issue #2412, opened 2015-09-20
escape key press


scottdraves:
The document URI somehow gets set to this string, like when I escape out of saving or publishing or something.

scottdraves:
<img width="254" alt="screen shot 2015-09-20 at 5 29 40 pm" src="https://cloud.githubusercontent.com/assets/963093/9983008/a672757c-5fbd-11e5-905c-dcf3a9e583c4.png">

scottdraves:
it gets in recently opened notebooks too:
<img width="196" alt="screen shot 2015-09-20 at 5 54 32 pm" src="https://cloud.githubusercontent.com/assets/963093/9983148/b714b400-5fc0-11e5-825e-78d3ecb641e6.png">

PR #2457, opened 2015-09-23
escape key press #2412

mnavasiolava:
When modal is dismissed make callback with empty result (like when Close btn clicked)


ildipo:
:+1:

scottdraves:
:+1:

Issue #2359, opened 2015-09-16
output breaks saving plot as png/svg


scottdraves:
https://pub.beakernotebook.com/#/publications/55f99dff-cdd6-427a-92ab-7395fd1a631b

PR #2461, opened 2015-09-23
fix output breaks saving plot as png/svg #2359

kpaharelau:
Support for the "Results" output type has been added


scottdraves:
menu entry appears but it gives

```
TypeError: $scope.outputDisplayModel.saveAsSvg is not a function
    at Object._saveAsItems.action (beakerApp.js:8170)
```

scottdraves:
:+1:

Issue #2446, opened 2015-09-22
old notebook flashes during load


scottdraves:
to reproduce:
1) run fresh beaker
2) open the main tutorial
3) click on JavaScript to open that tutorial
4) click on logo to return to start page
5) click on "tutorial" in the list

notice the page returns to the JS tutorial and display that until the main tutorial loads.
this is a regression.

scottdraves:
related? https://github.com/twosigma/beaker-notebook/issues/2406

maxyeg:
I've checked and it's not related to #2406.

PR #2464, opened 2015-09-23
#2446. old notebook flashes during load

maxyeg:
`initBeakerPrefs()` here makes `this.isEmpty()` falsy and cells are not cleared.
Move call to `initBeakerPrefs()` upwards and make it later.


scottdraves:
:+1:

Issue #2422, opened 2015-09-21
do these work for SQL?


scottdraves:
![screen shot 2015-09-21 at 12 48 51 pm](https://cloud.githubusercontent.com/assets/963093/9998452/2b6522ba-605f-11e5-8e21-8b4541f26aa0.png)

maksim-lauryk:
I checked and improved it. This should work.

PR #2466, opened 2015-09-23
#2422 'cancel' and 'kill all threads' for sql plugin

maksim-lauryk:
'cancel' and 'kill all threads' for sql plugin


scottdraves:
:+1:

Issue #2408, opened 2015-09-19
missing menu items


scottdraves:
there appears to be a race condition in how menu or init plugins are loaded because sometimes the ipynb item is missing:
<img width="496" alt="screen shot 2015-09-19 at 7 19 10 pm" src="https://cloud.githubusercontent.com/assets/963093/9978558/4c561ab2-5f04-11e5-84a3-c33554be447d.png">
vs
<img width="388" alt="screen shot 2015-09-19 at 7 20 46 pm" src="https://cloud.githubusercontent.com/assets/963093/9978559/51bb74ca-5f04-11e5-802b-2c2e596b5ae1.png">

PR #2473, opened 2015-09-24
missing menu items #2408

mnavasiolava:
The problem is in asynchronous calls. Sometimes file_ipython.js menu plugin is loaded before file.js menu plugin. In this case we create menu File -> Open -> Open.. IPython. Then when file.js menu is loaded we find existing File menu and start adding items to it. File.js menu also has Open submenu item and based on items priority we just replace existing Open submenu item (with Open.. IPython item) with the new one. That's why we have missing Open.. IPython menu item.
I suggest the following fix: if we have submenu with the same name and priority merge their items instead of replacing menus


ildipo:
:+1:  It looks like is a good solution (and simple) for the issue

Issue #2305, opened 2015-09-08
File open dialog sort should persist


scottdraves:
The sorting for the file open dialog should persist, even if i quit and restart Beaker.
It should be saved the same way advanced mode is.
![screen shot 2015-09-08 at 1 20 16 pm](https://cloud.githubusercontent.com/assets/963093/9742161/8f9fdc26-562c-11e5-93b3-7047ad3b7339.png)

PR #2487, opened 2015-09-25
fix File open dialog sort should persist #2305

kpaharelau:
sort options saved as preferences

### DIFF
--- a/core/src/main/java/com/twosigma/beaker/core/rest/UtilRest.java
+++ b/core/src/main/java/com/twosigma/beaker/core/rest/UtilRest.java
@@ -22,7 +22,7 @@ import com.twosigma.beaker.shared.module.util.GeneralUtils;
 import java.io.IOException;
 import java.nio.file.Paths;
 import java.nio.file.Files;
-import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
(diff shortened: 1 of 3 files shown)


ildipo:
Hi,
It works but I'm getting this exception when I change the sort by:

java.nio.file.NoSuchFileException: /Users/davidepasetto/.beaker/v1/config/beaker.pref.json.tmp
    at sun.nio.fs.UnixException.translateToIOException(UnixException.java:86)
    at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
    at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
    at sun.nio.fs.UnixCopyFile.move(UnixCopyFile.java:409)
    at sun.nio.fs.UnixFileSystemProvider.move(UnixFileSystemProvider.java:262)
    at java.nio.file.Files.move(Files.java:1395)
....

scottdraves:
For me, it corrupted the prefs file to end like this:

```
  "advanced-mode" : true,
  "fs-reverse" : true
} true
}
```

scottdraves:
why is this changing the mechanism to write the prefs file?
it should just add another name/value pair.

scottdraves:
we should look at this later, after 1.4 is done

Issue #1867, opened 2015-06-30
scala spark causes linkage error


scottdraves:
http://sharing.beakernotebook.com/gist/anonymous/6c7539f37aadfad9d00a

java.lang.LinkageError: loader constraint violation: when resolving method "java.lang.management.ManagementFactory.newPlatformMXBeanProxy(Ljavax/management/MBeanServerConnection;Ljava/lang/String;Ljava/lang/Class;)Ljava/lang/Object;" the class loader (instance of com/twosigma/beaker/jvm/classloader/JarClassLoader) of the current class, org/apache/spark/util/SizeEstimator$, and the class loader (instance of ) for the method's defining class, java/lang/management/ManagementFactory, have different Class objects for the type javax/management/MBeanServerConnection used in the signature

syepes:
Is there currently any kind of workaround to make Spark work with Scala?

scottdraves:
no sorry, I am baffled by the error and haven't had a chance to really dig into it yet.
maybe someone who knows spark better can help.

syepes:
I think this must be some kind of run-time class loading issue. I have done some similar tests with Groovy and I get the same issue running it from the console. But if I compile the same code in a fatJar it works as expected.

Are there any docs on how the backend works? to try and debug this issue.

PR #2493, opened 2015-09-25
Fix for scala spark - make sure you use the one compiled for scala 2.11

ildipo:
Fix issue #1867 - you MUST also use spark built for scala 2.11


syepes:
Is there anyway we could also make this work with Scala 2.10, as 2.11 is not used everywhere

scottdraves:
beaker comes only with scala 2.11 so this restriction does not seem out of order.
i would love it if we could improve it but i don't know how and we are out of time for v1.4 release.

ildipo:
We can compile the plugin with 2.10 - even distribute two plugins (one 2.11 and one 2.10 scala)

ildipo:
It looks like 2.11 has some breaking changes since all previous versions work with the same spark

scottdraves:
i am OK with moving plugin back to 2.10, Sebastian you probably know more about scala users than I do, what do you think?

scottdraves:
please move discussion to Issue #2499

Issue #2500, opened 2015-09-26
markdown h scrollbar broken


scottdraves:
when i click to scroll it vanishes
<img width="289" alt="screen shot 2015-09-26 at 6 07 07 pm" src="https://cloud.githubusercontent.com/assets/963093/10120093/7dc5b146-6479-11e5-8852-63b527701f86.png">

clicking on thumb part of scroll bar works, but clicking on bkg part (which normally means scroll-by-page),
instead exits the interaction.

PR #2510, opened 2015-09-28
markdown h scrollbar broken #2500

mnavasiolava:
seems like a defect in codemirror: blur event fires when we click on bkg part of a scroll.
but we can check cm.state.focused property to get to know whether the editor lost focus


scottdraves:
:+1:

Issue #2498, opened 2015-09-26
bad error handling


scottdraves:
if you try to save a notebook to a nonexistent location, you get a horrible stack trace.
it should give as short, clear error, as specific as possible, like "directory does not exist", "no write access to", or just "error saving to",
and give the name it tried to save.  and then when you clik "ok" it should take you back to the file dialog to try again.

PR #2512, opened 2015-09-28
fix bad error handling #2498

kpaharelau:
I shows very simple error message: ' "Error saving to " + uri'. We needs serious changes on backend for more clear error messaging. But I have opinion that it have sense after release. I can do it


scottdraves:
:+1:

Issue #2496, opened 2015-09-26
other fonts that should be roboto mono


scottdraves:
results from IPython:
<img width="129" alt="screen shot 2015-09-26 at 3 28 22 pm" src="https://cloud.githubusercontent.com/assets/963093/10119418/3deff1f0-6463-11e5-8a2f-ccd5b0a0e805.png">
errors:
<img width="801" alt="screen shot 2015-09-26 at 3 33 56 pm" src="https://cloud.githubusercontent.com/assets/963093/10119447/067da11c-6464-11e5-951c-7f312403fa25.png">
stdout, stderr:
<img width="530" alt="screen shot 2015-09-26 at 3 36 50 pm" src="https://cloud.githubusercontent.com/assets/963093/10119460/77e727f6-6464-11e5-9a80-575e22b88aab.png">
`code` typeface in markdown:
<img width="221" alt="screen shot 2015-09-26 at 6 35 57 pm" src="https://cloud.githubusercontent.com/assets/963093/10120181/7558583e-647d-11e5-99db-5bc4b9635f54.png">

PR #2513, opened 2015-09-28
#2496. Other fonts that should be roboto mono

maxyeg:


core/src/main/web/app/styles/variables.scss
```diff
@@ -25,6 +25,7 @@ $grid-gutter-width: 20px;
 $grid-row-height: 18px;
 
 $font-family-sans-serif: 'pt-sans', Helvetica, sans-serif;
+$font-family-roboto-mono: 'Roboto Mono';
```
scottdraves: please name variable $font-family-mono

scottdraves: and give "monospace" as a backup alternative

Issue #2507, opened 2015-09-27
stack trace on undefined name


scottdraves:
user error should only go in cell output, do not present stack trace in stderr window:

<img width="783" alt="screen shot 2015-09-27 at 1 27 50 pm" src="https://cloud.githubusercontent.com/assets/963093/10124146/c70437b8-651b-11e5-94b6-78ea4337122c.png">

scottdraves:
syntax errors too:

```
SEVERE: null
org.h2.jdbc.JdbcSQLException: Syntax error in SQL statement "INSERT INTO CITY VALUES ($ CITIES[*][0].CID , ?, ?, ?) "; expected "[, ::, *, /, %, +, -, ||, ~, !~, NOT, LIKE, REGEXP, IS, IN, BETWEEN, AND, OR, ,, )"; SQL statement:
INSERT INTO city VALUES ($ cities[0].cid , ?, ?, ?) [42001-189]
    at org.h2.message.DbException.getJdbcSQLException(DbException.java:345)
    at org.h2.message.DbException.getSyntaxError(DbException.java:205)
    at org.h2.command.Parser.getSyntaxError(Parser.java:524)
    at org.h2.command.Parser.read(Parser.java:3120)
```

scottdraves:
missing column:

```
SEVERE: null
org.h2.jdbc.JdbcSQLException: Column "COLOR" not found; SQL statement:
SELECT * FROM color WHERE color LIKE 'A%' [42122-189]
    at org.h2.message.DbException.getJdbcSQLException(DbException.java:345)
    at org.h2.message.DbException.get(DbException.java:179)
    at org.h2.message.DbException.get(DbException.java:155)
    at org.h2.expression.ExpressionColumn.optimize(ExpressionColumn.java:147)
    at org.h2.expression.CompareLike.optimize(CompareLike.java:87)
    at org.h2.command.dml.Select.prepare(Select.java:838)
    at org.h2.command.Parser.prepareCommand(Parser.java:246)
    at org.h2.engine.Session.prepareLocal(Session.java:461)
    at org.h2.engine.Session.prepareCommand(Session.java:403)
    at org.h2.jdbc.JdbcConnection.prepareCommand(JdbcConnection.java:1188)
    at org.h2.jdbc.JdbcPreparedStatement.<init>(JdbcPreparedStatement.java:72)
    at org.h2.jdbc.JdbcConnection.prepareStatement(JdbcConnection.java:276)
[...]

PR #2514, opened 2015-09-28
#2507  stack trace on undefined name

maksim-lauryk:
#2507  user error is shown only in cell output


scottdraves:
Don't say "Error" twice.  put the main error on the 1st line so user does not have to open the error.
screen shot 2015-09-28 at 10 52 19 am
screen shot 2015-09-28 at 10 52 32 am

scottdraves:
but these seem to be the wrong errors.  if there is an error in %% or ${} directive, then it should be printed clearly.  For example, instead of saying "null" it should say the bad name:

screen shot 2015-09-28 at 10 59 58 am

And in the 1st error originally reported where I put in ${sales[0]} but it was not defined, the error displayed was "named not defined: sales" that was good.  please leave that.  just git rid of the stack trace!  now I am getting a horrible SQL parsing error.

maksim-lauryk:
updated

Issue #2323, opened 2015-09-11
Clojure code should run in a notebook unique namespace


farrellm:
Currently, all Clojure code runs in the "clojure.core" namespace, but this can cause problems when importing external code and allows bad interactions between notebooks. This can be avoided by giving each notebook a default namespace where code is evaluated by default.

Possible function to do this:

```
(defn load-string-in-ns [n s]
  (binding [*ns* (find-ns n)]
    (load-string s)))
```

References:
http://stackoverflow.com/questions/6443594/how-to-load-string-to-current-namespace-not-core

scottdraves:
Thanks Matt, Clojure is due for an overhaul in about a month, we'll fix this then.

PR #2515, opened 2015-09-28
#2323  Clojure code should run in a notebook unique namespace 

maksim-lauryk:
#2323  Clojure code should run in a notebook unique namespace


scottdraves:
didn't work, i created two notebooks and in one ran (def x 1) and in the other i ran x and got 1.

scottdraves:
resubmit when it's fixed

Issue #2506, opened 2015-09-27
default JDBC config?


scottdraves:
Is there a way to have a default database for all the SQL cells in a notebook?

The "Data sources" field in the Language Manager allows me to define names which are then used like `%%beakerDB = H2MemDB2`.  but then still every cell needs `%%beakerDB` something.

There should either be another field for the default source, or if you enter a line in Data sources that does not have an "=", then it would be the default.  Or am I just missing how to do this?

maksim-lauryk:
in current realisation we have three ways for defined data source, e.g: 
%%beakerDB="jdbc:h2:mem:db1"
-- loading of connection string from plugin settings
%%beakerDB = H2MemDB2
-- load connection string from beaker object
%%beakerDB=${H2MemDB2}

There's not any way for setup default notebook data source, but it can be realized if necessary.

ildipo:
Please add a default database option in the language manager

scottdraves:
new field should be "Default data source"
current field should be renamed "Named data sources"

PR #2519, opened 2015-09-28
#2506  default JDBC config 

maksim-lauryk:
added default JDBC config, updated sql_H2_plugin_test.bkr


scottdraves:
This change looks correct but I cannot merge it because your other work is also on this branch.
Each branch should only have one change.

I cherry-picked the good commit.

Issue #2451, opened 2015-09-22
Publish As New removes images on old publication


planyon:
Publish a notebook with a preview image.
Click Publish from the menu again and remove the image and then Publish As New.
The preview image on the original publication gets deleted when it shouldn't.
The new publication does not have a preview image, as expected.

grin:
Fixed in: https://github.com/twosigma/beaker-notebook/pull/2522

PR #2522, opened 2015-09-28
Lazily delete attachments when publishing notebooks

grin:
Fixes: https://github.com/twosigma/beaker-notebook/issues/2451

### DIFF
--- a/core/src/main/web/app/mainapp/components/publication/publication.js
+++ b/core/src/main/web/app/mainapp/components/publication/publication.js
@@ -170,13 +170,21 @@
         return $scope.publish(action);
       };
 
-      $scope.publish = function(action) {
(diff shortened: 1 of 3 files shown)


scottdraves:
the 1st publication ended up with a broken image like this one:
https://pub.beakernotebook.com/#/publications/5609dbcb-27ef-4dd2-a459-17946a59d936

grin:
I tested image upload with code in this PR and the publication api url in Beaker set to `https://pub-staging.beakernotebook.com`...and it worked for me, see the publication titled `3` by bartek. What was your setup when you were testing this?

grin:
Also, this should only work on https://pub-staging.beakernotebook.com/#/publications because image scaling was only deployed there.

ildipo:
That's correct - I'll re-deploy as soon as beaker build works

PR #2528, opened 2015-09-29
Maxy/2525

maxyeg:


scottdraves:
this looks good.  can you add a menu/triangle button on the right (like the combo-box),
and make it only open if you click that button.  if you click the text, then you can edit just as text without the widget.

scottdraves:
ok, this is pretty close.  just needs a few details:
1) if you click the menu button again, the date picker should close.
2) the text part is too wide, it should have room for 9 chars.

scottdraves:
:+1:

Issue #2504, opened 2015-09-27
login flash in dialog


scottdraves:
If you are already logged in, and you publish a notebook, when the dialog opens
the "login" screen flashes before the correct dialog appears.

grin:
Fixed in: https://github.com/twosigma/beaker-notebook/pull/2529

PR #2529, opened 2015-09-29
Do not flash the sign in view if users is already signed in

grin:
Fixes: https://github.com/twosigma/beaker-notebook/issues/2504


ildipo:
:+1:

Issue #2524, opened 2015-09-28
sql should not require %%inputs %%outputs


scottdraves:
they are only needed for the [*] version i believe.

maksim-lauryk:
updated, now %%inputs %%outputs are not required in all cases

PR #2533, opened 2015-09-29
SQL plugin improvements

maksim-lauryk:
added custom error on unfound beaker objects
improved canceling of sql queries


scottdraves:
i'm not seeing the custom error
screen shot 2015-09-29 at 11 35 22 am

scottdraves:
also, please remember, one change in each pull request.

maksim-lauryk:
in this case we have to merge code with the branch:
#2524 sql should not require %%inputs %%outputs

or test it on following SQL: 
%%inputs: idx
select \* from color where id = ${idx}

It should be seems on:

badvar

PR #2547, opened 2015-09-30
#2323  Clojure code should run in a notebook unique namespace 

maksim-lauryk:
#2323  Clojure code should run in a notebook unique namespace

fixed bug with substitution of namespaces in previous version


scottdraves:
:+1:

Issue #2531, opened 2015-09-29
selecting code can fail


scottdraves:
if you drag to select text in a code cell, and the endpoint (where you release the mouse) is
outside the box, then the selection fails.

PR #2548, opened 2015-09-30
selecting code can fail #2531

mnavasiolava:
the problem was in ng-click which fires on mouse up event (https://github.com/codef0rmer/angular-dragdrop/issues/50). 
when we release the mouse inside code cell but outside codemirror element backgroundClick function is called and it clears selection and set up cursor position in editor.
i used ng-mousedown directive instead of ng-click.


scottdraves:
:+1:

Issue #2549, opened 2015-09-30
Java plugin doesn't work in the electron application for the windows OS


kpaharelau:
![image](https://cloud.githubusercontent.com/assets/13516511/10189055/b1ebabd8-676c-11e5-90db-d43da2a1e1d3.png)

PR #2550, opened 2015-09-30
fix Windows OS - Java plugin doesn't work in the electron application…

kpaharelau:
… for the windows OS #2549

 \resources\jre has been renamed to \resources\jdk because for JAVA_HOME with name 'jre' tools.jar is searched on the level jre folder.

### DIFF
--- a/launcher/electron/app/backend-runner.js
+++ b/launcher/electron/app/backend-runner.js
@@ -34,14 +34,14 @@ module.exports = (function() {
       var eventEmitter = new events.EventEmitter();
 
       if (_osName.startsWith('Windows')) {
-        process.env['JAVA_HOME'] = path.resolve(__dirname + '/../jre');
+        process.env['JAVA_HOME'] = path.resolve(__dirname + '/../jdk');
         process.chdir(__dirname + '/../dist');
         _backend = spawn(path.resolve(__dirname + '/../dist/beaker.command.bat'), ['--open-browser', 'false']);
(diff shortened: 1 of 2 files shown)


kpaharelau:
see 
javax.tools.ToolsProvider

```
private static final String[] defaultToolsLocation = { "lib", "tools.jar" };

   private Class<?> findSystemToolClass(String toolClassName)
       throws MalformedURLException, ClassNotFoundException
   {
       // try loading class directly, in case tool is on the bootclasspath
       try {
           return Class.forName(toolClassName, false, null);
       } catch (ClassNotFoundException e) {
           trace(FINE, e);

           // if tool not on bootclasspath, look in default tools location (tools.jar)
           ClassLoader cl = (refToolClassLoader == null ? null : refToolClassLoader.get());
           if (cl == null) {
               File file = new File(System.getProperty("java.home"));
               if (file.getName().equalsIgnoreCase("jre"))
                   file = file.getParentFile();
               for (String name : defaultToolsLocation)
                   file = new File(file, name);

               // if tools not found, no point in trying a URLClassLoader
               // so rethrow the original exception.
               if (!file.exists())
                   throw e;

               URL[] urls = { file.toURI().toURL() };
               trace(FINE, urls[0].toString());

               cl = URLClassLoader.newInstance(urls);
               refToolClassLoader = new WeakReference<ClassLoader>(cl);
           }

           return Class.forName(toolClassName, false, cl);
       }
   }
```

scottdraves:
:+1:

Issue #2553, opened 2015-09-30
refactor to expose text processing


scottdraves:
as requested https://github.com/twosigma/beaker-notebook/issues/2545

PR #2559, opened 2015-09-30
#2553. refactor to expose text processing

maxyeg:
Make common method in helper.js

### DIFF
--- a/core/src/main/web/app/helpers/helper.js
+++ b/core/src/main/web/app/helpers/helper.js
@@ -305,6 +305,108 @@
           return false;
         }
       },
+      markupCellContent: function(cellContent) {
+        var markupDeferred = bkHelper.newDeferred();
+
+        if (!this.bkRenderer) {
+          // Override markdown link renderer to always have `target="_blank"`
+          // Mostly from Renderer.prototype.link
+          // https://github.com/chjj/marked/blob/master/lib/marked.js#L862-L881
+          var bkRenderer = new marked.Renderer();
+          bkRenderer.link = function(href, title, text) {
+            var prot;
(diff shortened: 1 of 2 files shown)


core/src/main/web/app/helpers/helper.js
```diff
@@ -305,6 +305,108 @@
...
+          if (results.indexOf(text) === -1)
+            results.push(text);
+        }
+
+        var evaluateCode = function (index) {
+
+          if (index === results.length) {
+            markIt(cellContent);
+          } else {
+            bkHelper.evaluateCode("JavaScript", results[index][1]).then(
```
scottdraves: @grin will that work on pub?

grin: I think this will currently fail on Publications. We don't initialize Beaker app so `getCurrentApp` implemented as (and needed to eval code):

``` js
    var getCurrentApp = function() {
      return bkCoreManager.getBkApp();
    };
```

will be `undefined`. If we could somehow explicitly evaluate this code with plain javascript `eval()` without going through Beaker's evaluators it would work...

scottdraves: how about we make this new helper function take a 2nd argument that injects the evaluator?
then in Beaker you can pass in evaluateCode, and in Pub you can pass in a function that only works for JS, it just reduces to a call to eval() more simply.  the only thing it has to do that's a little tricky is set up the beaker object to access the namespace in the document model, so you can say {{beaker.x}}.

grin: yeah, that's a good solution
:+1:

Issue #1806, opened 2015-06-21
clojure crashes on the following code:


scottdraves:
```
Jun 21, 2015 4:57:12 PM com.twosigma.beaker.jvm.serialization.BasicObjectSerializer runConfiguredSerializers
SEVERE: exception in serialization
java.lang.IllegalArgumentException: Don't know how to create ISeq from: clojure.core$fib_seq_iterative
    at clojure.lang.RT.seqFrom(RT.java:505)
    at clojure.lang.RT.seq(RT.java:486)
    at clojure.core$seq.invoke(core.clj:133)
    at clojure.core$take$fn__4270.invoke(core.clj:2626)
    at clojure.lang.LazySeq.sval(LazySeq.java:40)
    at clojure.lang.LazySeq.seq(LazySeq.java:49)
    at clojure.lang.LazySeq.isEmpty(LazySeq.java:171)
    at com.twosigma.beaker.jvm.serialization.BasicObjectSerializer.isListOfPrimitiveTypeMaps(BasicObjectSerializer.java:71)
```

from

```
(defn fib-step [[a b]]
  [b (+ a b)])

(defn fib-seq-iterative []
  (map first (iterate fib-step [0 1])))
(take 20 fib-seq-iterative)
```

maksim-lauryk:
I new in Clojure, but it looks like that 'fib-seq-iterative' function in last string should be surrounded by parentheses: (take 20 (fib-seq-iterative))

![fib_seq](https://cloud.githubusercontent.com/assets/13995034/10188900/add36d8e-676b-11e5-964f-d7063d1e8c64.PNG)

scottdraves:
the user should not be able to crash the system.
if there is an error it should be reported nicely.

maksim-lauryk:
added a check of corrupted clojure objects.
now error message looks like that:
![clojure_obj_error](https://cloud.githubusercontent.com/assets/13995034/10214583/3f6e8d08-6822-11e5-8aef-5f8efe153aa6.PNG)

scottdraves:
i think it should produce this:

```
#'sandbox10869/fib-step
```

but with a different namespace....
it is not an error.

maksim-lauryk:
I tried to run this code in a REPL. It executes code line by line and #'sandbox10869/fib-step is an output for the first line. If we execute our code fully, we get the same error:
[...]

PR #2568, opened 2015-10-01
#1806 clojure crashes on the following code

maksim-lauryk:
added a checking of corrupted clojure objects


scottdraves:
see original issue for comment

PR #2572, opened 2015-10-01
Edit menu items crash electron for windows os #2571

mnavasiolava:
electron doesn't crashes for windows and unix but i can't test it on mac os. could you please do this?
select all menu item doesn't work yet


scottdraves:
:+1:

PR #2579, opened 2015-10-01
Run SQL autocomplete queries in a separate thread

siarhei-shvaiko:
Fixed error: DBConnectionException; must be caught or declared to be thrown


scottdraves:
don't change the copyright banners.

also, i can't see the diff since the commit combines it with reformatting the code.

PR #2580, opened 2015-10-01
🔒 specify a version range for new npm packages

samccone:


scottdraves:
thanks sam.

samccone:
np

scottdraves:
this change looks good but we would need a CLA from you.... possible?

PR #2590, opened 2015-10-02
Electron on Windows OS hanging plugins processes fix #2555

### LINKED ISSUE #2555: Electron problems on Windows OS
1. Processes of plugins remain in memory after beaker closing
2. Plugins doesn't work  with paths that  contains spaces

mnavasiolava:
use python subprocess.call command to start plugins in the same process tree on windows.
so plugins' processes can be killed by "taskkill" command when electron is closed


scottdraves:
:+1:

Issue #2477, opened 2015-09-24
Clojure classpath


ildipo:
We want a settable classpath and dynamic classpath, shared with the other JVM languages.

maksim-lauryk:
Classpath and imports are available to set up in the Clojure plugin settings

PR #2608, opened 2015-10-05
#2477  Clojure classpath and imports

maksim-lauryk:
Classpath and imports are available to set up in the Clojure plugin settings


ildipo:
Hi Max,
please use `com.twosigma.beaker.jvm.classloader.DynamicClassLoaderSimple` as a class loader instead of the URLClassLoader - look at how this is used in javash plugin.
You should also include the outDir function - this will allow you to have a Java cell that defines a class and then a clojure cell that uses the class.

maksim-lauryk:
used com.twosigma.beaker.jvm.classloader.DynamicClassLoaderSimple instead of the URLClassLoader
improved a file path processing in the DynamicClassLoaderSimple loader
included the outDir function

ildipo:
:+1:

Issue #2597, opened 2015-10-03
Please modernize AngularUI usage


jonassmedegaard:
AngularJS used to be a small collection of modules distributed as a whole - that's what is currently (as of versions 1.3.4 and 1.4) included with Beaker Noterbook.

For Debian packaging we prefer to track each upstream source independently, which reveals that upstream is now split into individual modules.

Grepping Beaker Notebook code for instances of "angular.module" seems to indicate that the modules actually used are ui.keypress and ui.bootstrap (the latter not even included in the legacy bundle).

If correct, then I suggest to replace the bundle with those two modules, both to drop unneeded baggage and get eventual recent upstream improvements - in addition to making Debian packaging slightly easier :-)

jonassmedegaard:
Hmm - seems ui.keypress is not even part of the legacy ui-utils bundle any more: https://github.com/angular-ui/ui-utils

Seems to me that if not refactoring it away then essentially you should consider yourself current upstream of ui-keypress because upstream for that code has gone. :-(

scottdraves:
Thanks for the suggestion.  Can you refer me to a project that does what you mean?

jonassmedegaard:
Quoting Scott Draves (2015-10-03 19:18:51)

> Thanks for the suggestion. Can you refer me to a project that does 
> what you mean?

I am not experienced with AngularJS, nor do I understand how Beaker uses 
AngularUI.

It seems (but check - don't trust me on that!) Beaker only uses a single 
module - ui.keypress - from AngularUI module bundle ui-utils.

It also seems(!) that ui-utils is deprecated in favor of individually 
tracked modules.

Furthermore it seems(!) that ui.keypress is gone even from the 
deperecated bundle.

What I suggest is therefore to a) check if what I have found is correct, 
and if so then b) look into replacing uses of ui.keypress with something 
that is maintained.  Stackoverflow has a slew of entries for keypress 
and AngularJS, but I don't really know what might be relevant.

Hope that is of some use... :-/
- Jonas

## 
- Jonas Smedegaard - idealist & Internet-arkitekt
- Tlf.: +45 40843136  Website: http://dr.jones.dk/
  
  [x] quote me freely  [ ] ask before reusing  [ ] keep private

jonassmedegaard:
Perhaps core AngularJS is adequate: https://docs.angularjs.org/api/ng/directive/ngKeypress

scottdraves:
OK so you would like to get rid of this, right?
[...]

PR #2609, opened 2015-10-05
 #2597. Please modernize AngularUI usage

maxyeg:
As requested, `ui-utils.min.js` removed.
I found only one module from this lib in Beaker (`ui.keypress`).
Currently nothing from `ui.keypress` is used in Beaker. 

I didn't touch `ui.bootstrap`as it's not a part of `ui-utils`. It's a standalone project with active upstream(https://github.com/angular-ui/bootstrap).


scottdraves:
:+1:

Issue #2605, opened 2015-10-05
missing kdb results in stack trace and hang


scottdraves:
```
[qtp398572781-17] WARN org.eclipse.jetty.servlet.ServletHandler - 
javax.servlet.ServletException: java.lang.Exception: Could not start kdb
...
Caused by: java.lang.Exception: QHOME is not set
    at com.twosigma.beaker.kdb.KdbProcess.<init>(KdbProcess.java:71)
    at com.twosigma.beaker.kdb.KdbShell.<init>(KdbShell.java:64)
    at com.twosigma.beaker.kdb.rest.KdbRest.getShell(KdbRest.java:82)
    ... 47 more
[qtp398572781-17] WARN org.eclipse.jetty.servlet.ServletHandler - /rest/kdb/getShell
java.lang.Exception: Could not start kdb
    at com.twosigma.beaker.kdb.rest.KdbRest.getShell(KdbRest.java:85)
```

scottdraves:
it should pop up the dialog about how to install the language.
instead the light goes yellow, sticks and blocks any other language from loading.

scottdraves:
this looks great, please apply to all languages.

PR #2610, opened 2015-10-05
missing kdb results in stack trace and hang #2605

mnavasiolava:
angular promise was neither rejected nor resolved when creating new shell failed.
added error callback to plugin's newShell function where angular promise is rejected


scottdraves:
this looks great, please apply to all languages.

Issue #2544, opened 2015-09-30
ipython math display broken


scottdraves:
![screen shot 2015-09-29 at 11 09 17 pm](https://cloud.githubusercontent.com/assets/963093/10183584/4c11f8b6-66ff-11e5-881e-fc3a3d02394f.png)
https://pub.beakernotebook.com/#/publications/560b5279-8469-45ae-823e-03de070213b6

grin:
If I open that notebook in my local Beaker that output cell looks the same and also shows `$$`... is this a bug in Beaker? or we should somehow handle it differently in Publications?

maxyeg:
![ebb0e-clip-13kb](https://cloud.githubusercontent.com/assets/2286032/10279376/eca705ba-6b6d-11e5-9621-fbcbf348efbe.png)

That's how I see it in beaker(built from current master).

grin:
Ok, tested it in latest master:
![image](https://cloud.githubusercontent.com/assets/12468/10279532/cb4c6196-6b66-11e5-9bb5-47169bd34c93.png)

looks like it depends on the evaluator status which is not properly initialized for me locally. This could also explain why it isn't working on Publications where we don't initialize any evaluators...

ildipo:
`plugin/ipythonPlugins/src/dist/ipython/ipython.js` lines 339 to 356 are triggering the rendering.

I think this function should be _moved_ to the Html output display: if there's a div with a class whose name ends in `-latex-class` interpret it as latex output.
This way the publication server will automatically work.

grin:
@ildipo thanks, I got it working locally without ipython based on your suggestion. I will push a formal fix in a bit later today (need to do some refactorings around it)

ildipo:
ok it seems to work fine.

PR #2611, opened 2015-10-05
Move latex typesetting from ipython plugin to html output directive

grin:
Fixes: https://github.com/twosigma/beaker-notebook/issues/2544

### DIFF
--- a/core/src/main/web/app/helpers/helper.js
+++ b/core/src/main/web/app/helpers/helper.js
@@ -305,6 +305,24 @@
           return false;
         }
       },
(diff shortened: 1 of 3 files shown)


grin:
@ildipo I couldn't full test this locally because ipython isn't working for me (haven't tried to fix it yet)...so the only thing I didn't test was to see if latex output is properly re-renedered after ipython code is changed and re-run. But initial render works fine.

ildipo:
@grin it is working fine - merging

Issue #2591, opened 2015-10-02
ipython backends broken in Docker


scottdraves:
beaker no longer seems to work with ipython v4 and even if I got back to https://github.com/twosigma/beaker-notebook/pull/2280
it still doesn't work.  what happened??

scottdraves:
```
Python-stderr>Traceback (most recent call last):
IPython-stderr>  File "/Users/spot/beaker-notebook/core/config/plugins/eval/ipythonPlugins/ipython/ipythonPluginStart.py", line 24, in <module>
IPython-stderr>    sys.exit(start_ipython(["notebook", "--config=" + os.environ["beaker_ipython_notebook_config"]]))
IPython-stderr>  File "/Users/spot/anaconda/lib/python2.7/site-packages/IPython/__init__.py", line 118, in start_ipython
IPython-stderr>    return launch_new_instance(argv=argv, **kwargs)
IPython-stderr>  File "/Users/spot/anaconda/lib/python2.7/site-packages/traitlets/config/application.py", line 591, in launch_instance
IPython-stderr>    app.initialize(argv)
IPython-stderr>  File "<decorator-gen-111>", line 2, in initialize
IPython-stderr>  File "/Users/spot/anaconda/lib/python2.7/site-packages/traitlets/config/application.py", line 75, in catch_config_error
IPython-stderr>    return method(app, *args, **kwargs)
IPython-stderr>  File "/Users/spot/anaconda/lib/python2.7/site-packages/IPython/terminal/ipapp.py", line 305, in initialize
IPython-stderr>    super(TerminalIPythonApp, self).initialize(argv)
IPython-stderr>  File "<decorator-gen-7>", line 2, in initialize
IPython-stderr>  File "/Users/spot/anaconda/lib/python2.7/site-packages/traitlets/config/application.py", line 75, in catch_config_error
IPython-stderr>    return method(app, *args, **kwargs)
IPython-stderr>  File "/Users/spot/anaconda/lib/python2.7/site-packages/IPython/core/application.py", line 386, in initialize
IPython-stderr>    self.parse_command_line(argv)
IPython-stderr>  File "/Users/spot/anaconda/lib/python2.7/site-packages/IPython/terminal/ipapp.py", line 300, in parse_command_line
IPython-stderr>    return super(TerminalIPythonApp, self).parse_command_line(argv)
IPython-stderr>  File "<decorator-gen-4>", line 2, in parse_command_line
[...]

PR #2625, opened 2015-10-06
fix ipython backends broken in Docker #2591

scottdraves:
I got ipyversion[0] == "51" without decoding to utf-8. All python languages except ruby are working now on the docker. IPython version == 4.0.0


scottdraves:
tested with ipy 2, 3, 4 on mac, plus docker

Issue #1171, opened 2015-02-08
default notebook does not handle multiline cells


scottdraves:
normal beaker files are stored with the body of code cells as an array of strings.
these do not work in the defaulte notebook.

PR #2635, opened 2015-10-07
 #1171. default notebook does not handle multiline cells

maxyeg:


core/src/main/web/app/utils/utils.js
```diff
@@ -227,6 +227,11 @@
         var deferred = angularUtils.newDeferred();
         angularUtils.httpGet(serverUrl("beaker/rest/util/getDefaultNotebook")).
             success(function(data) {
+              if (!_.isString(data)) {
+                // angular $http auto-detects JSON response and deserialize it using a JSON parser
+                // we don't want this behavior, this is a hack to reverse it
+                data = JSON.stringify(data);
+              }
               deferred.resolve(angular.fromJson(data));
```
scottdraves: why not delete the conditional and just call resolve(data)?
looks you you convert from structure to string and back.

maxyeg: You're right, it was unneeded effort. We can go with just resolve(data) as importer in mainapp.js will handle it correctly.

scottdraves:
:+1:

Issue #2644, opened 2015-10-07
iruby backend broken in Docker container


scottdraves:

scottdraves:
continuation of https://github.com/twosigma/beaker-notebook/issues/2591

PR #2645, opened 2015-10-07
fix iruby backend broken in Docker container #2644

kpaharelau:
iruby doesn't work without registered IRuby kernel in the '~/.ipython/kernels/ruby/kernel.json'. This file can be created by command 'iruby register'

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,12 +36,7 @@ RUN su -m beaker -c "cd /home/beaker/src  && gradle build"
 #      Setup      #
 ###################
 
-RUN su -m beaker -c "julia --eval 'Pkg.add(\"IJulia\")'" && \
-    su -m beaker -c "julia --eval 'Pkg.add(\"Gadfly\")'"
-
-RUN su -m beaker -c "iruby register"
-
-RUN mkdir -p /home/beaker/.beaker/v1/config && \
+WhebRUN mkdir -p /home/beaker/.beaker/v1/config && \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -39,6 +39,8 @@ RUN su -m beaker -c "cd /home/beaker/src  && gradle build"
 RUN su -m beaker -c "julia --eval 'Pkg.add(\"IJulia\")'" && \
     su -m beaker -c "julia --eval 'Pkg.add(\"Gadfly\")'"
 
+RUN su -m beaker -c "iruby register"
```
scottdraves: This should go in the base_docker_image/Dockerfile

Dockerfile
```diff
@@ -36,10 +36,7 @@ RUN su -m beaker -c "cd /home/beaker/src  && gradle build"
 #      Setup      #
 ###################
 
-RUN su -m beaker -c "julia --eval 'Pkg.add(\"IJulia\")'" && \
-    su -m beaker -c "julia --eval 'Pkg.add(\"Gadfly\")'"
-
-RUN mkdir -p /home/beaker/.beaker/v1/config && \
+WhebRUN mkdir -p /home/beaker/.beaker/v1/config && \
```
scottdraves: what is this? "Wheb"

scottdraves:
please do not change julia, just fix iruby.
if there is a problem with julia, please file an issue.

PR #2660, opened 2015-10-08
Cit/2305

kpaharelau:
I found that problem was in concurrent access to pref file. I have used synchronized access to setPreference function


core/src/main/java/com/twosigma/beaker/core/rest/UtilRest.java
```diff
@@ -23,15 +23,8 @@
...
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.Arrays;
+
+import java.util.*;
```
scottdraves: style guide prohibits import of *, please leave all modules listed.
http://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s3.3.1-wildcard-imports

scottdraves:
approved pending removal of wildcard import.

PR #2662, opened 2015-10-08
saved plots typeface wrong #2187

mnavasiolava:
external fonts are embedded into svg as base64 string.

one thing i want to pay your attention to is that i get fonts synchronously (to not make a mess in the code with numerous async callbacks) but if you are against i will switch it to async calls

### DIFF
--- a/core/src/main/web/outputdisplay/bko-plot/plotutils.js
+++ b/core/src/main/web/outputdisplay/bko-plot/plotutils.js
@@ -532,16 +532,16 @@
 
       base64Fonts: {},
 
-      getFontToInject: function(font){
+      getFontToInject: function(font) {
         var defer = bkUtils.newDeferred();
         var src = '';
         for (var url in font.urlformats) {
           if (font.urlformats.hasOwnProperty(url)) {
             var format = font.urlformats[url];
-            if(this.base64Fonts[url] == null){
-              this.base64Fonts[url] =  this.base64Encode(this.getBinary(url));
+            if (this.base64Fonts[url] == null) {
+              this.base64Fonts[url] = this.base64Encode(this.getFileSynchronously(url));
             }
-            src += "url('data:application/font-" + format +";charset=utf-8;base64," + this.base64Fonts[url] + "') format('" + format + "'), ";
+            src += "url('data:application/font-" + format + ";charset=utf-8;base64," + this.base64Fonts[url] + "') format('" + format + "'), ";
           }
         }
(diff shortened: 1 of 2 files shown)


core/src/main/web/outputdisplay/bko-plot/plotutils.js
```diff
@@ -515,7 +528,69 @@
...
+          "font-family: '" + font.fontFamily + "';" +
+          "src: " + src + ";" +
+          "font-weight: " + font.fontWeight + ";" +
+          "font-style: " + font.fontStyle + ";" +
+          " }\n";
+        return defer.promise;
+      },
+
+      //http://stackoverflow.com/questions/7370943/retrieving-binary-file-content-using-javascript-base64-encode-it-and-reverse-de
+      base64Encode: function(str) {
```
scottdraves, line 558: we should be able to use btoa() right?
only old IE has a problem with it afaik.

mnavasiolava, line 558: i tried to use btoa(), but i got an error (in chrome)
"Error: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range."

scottdraves, line 558: http://stackoverflow.com/questions/23223718/failed-to-execute-btoa-on-window-the-string-to-be-encoded-contains-characte

scottdraves, line 558: but maybe with one of these instead of unescape()?  http://www.w3schools.com/jsref/jsref_unescape.asp

not sure

mnavasiolava, line 558: using btoa(unescape(encodeURIComponent(str))) causes error in result svg:
"Failed to decode downloaded font: data:application/font-woff;charset=utf-8;base64,d09GRgABAAAAAHjvn5wAEwAAAAD…6B75+9756FcABGWjgEByNhDA4TYQLvn4bvn4LvnoIC75+9AX3vn6NrMQAAAAFTcO+fmT0AAA==
SVG%20Legend%20(13).svg:1 OTS parsing error: incorrect file size in WOFF header"

scottdraves, line 558: thanks for trying.

scottdraves:
:+1:

PR #2663, opened 2015-10-08
saved plots are missing legend #2186

mnavasiolava:
foreignObject is used to embed html legend to svg.
for correct legend displaying this PR is needed https://github.com/twosigma/beaker-notebook/pull/2662


scottdraves:
:+1:

Issue #2434, opened 2015-09-22
if new folder is created during save


scottdraves:
need to reload the list of files so that folder becomes visible,
and also enter it, so you can type a notebook name and it goes into the new folder.

scottdraves:
can't verify this, need to look more closely

PR #2668, opened 2015-10-08
fix if new folder is created during save #2434

kpaharelau:
improvement for a filling input after create new folder


scottdraves:
:+1:

Issue #2634, opened 2015-10-07
JS exception about missing fonts in open/save dialogs


maxyeg:
1. Use Open dialog(or Save/Save as dialog) to view file system tree. 
2. Expand some folders which have files
3. When tree contains files exceptions about missing fonts are thrown

![f6eab-clip-54kb](https://cloud.githubusercontent.com/assets/2286032/10337102/33ef07a8-6d06-11e5-8d3c-84251de901b1.png)

![d0a82-clip-31kb](https://cloud.githubusercontent.com/assets/2286032/10337078/0faacd46-6d06-11e5-84d5-494376f2ada1.png)

PR #2680, opened 2015-10-09
fix JS exception about missing fonts in open/save dialogs #2634

kpaharelau:
path '../components/bootstrap/fonts/' was hardcoded in the core/src/main/web/app/styles/ipython/style/style.min.css. It is replaced to the correct path in the gulp file now


scottdraves:
:+1:

Issue #2647, opened 2015-10-07
use stable release of datatables-fixedcolumns


jonassmedegaard:
It seems the fix for issue #1905  has been included in stable release 3.1.0 of datatables-fixedcolumns by now.

I suggest you switch to tracking that instead of a development snapshot.

scottdraves:
Good point, thanks.

PR #2698, opened 2015-10-12
#2647. use stable release of datatables-fixedcolumns

maxyeg:


ildipo:
:+1:

scottdraves:
looks like this changes the color of the index column header.
can that be avoided?

scottdraves:
:+1:

ildipo:
Actually I think we should override the CSS rather than removing it from the checked in file.
When we update bower component this will be overwritten

scottdraves:
good point davide

Issue #2476, opened 2015-09-24
Clojure require configuration


ildipo:
The clojure language manager interface needs a textbox for 'require' that should automatically added - like the "imports" textbox for java plugin.

maksim-lauryk:
added the require configuration

PR #2707, opened 2015-10-13
#2476  Clojure require configuration 

maksim-lauryk:
added the require configuration
needs the issue  #2477  Clojure classpath and imports


ildipo:
:+1:

Issue #2706, opened 2015-10-12
CombinedPlots heights


scottdraves:
if you do not provide a 2nd argument to the add method, it should default to 1.

and if there are multiple plots, they weights do not seem to translate to heights linearly:
https://pub.beakernotebook.com/#/publications/561c37b1-7ce6-4232-b648-3397327458ea
these 3 charts should have the same height since they have the same weight.

scottdraves:
![screen shot 2015-10-12 at 6 44 00 pm](https://cloud.githubusercontent.com/assets/963093/10440913/3939d030-7111-11e5-90b5-66cda4d16e99.png)

PR #2711, opened 2015-10-14
CombinedPlots heights #2706

mnavasiolava:
1. API was updated: now we can add subplot without providing weight (default weight is 1)
2. Actually plots’ weights are translated linearly to heights, but plot’s height also includes margins (which are constant for plots) and that’s why visually it seems that subplots have different heights (http://clip2net.com/s/3oPBjYd). 
   I've changed the algorithm of heights calculation so it  includes margins as well.

Concerning your screenshot with different heights: maybe you created plots with different weights, then ran the cell and then tried to change weight?  
in this case, yes, the heights will be different because plot size is saved to state and remains unchanged between cell evaluations. If you create new cell it should work.

### DIFF
--- a/core/src/main/web/outputdisplay/bko-plot/combinedplotformatter.js
+++ b/core/src/main/web/outputdisplay/bko-plot/combinedplotformatter.js
@@ -58,13 +58,26 @@
         var plotType = model.plot_type;
         if (plotType == null) { plotType = "Plot"; }
 
+        var layout = {
+          bottomLayoutMargin : 30,
+          labelHeight: 12
+        };
         var sumweights = 0;
+        var sumvmargins = 0;
+        var vmargins = [];
         var weights = model.weights == null ? [] : model.weights;
         for(var i = 0; i < model.plots.length; i++) {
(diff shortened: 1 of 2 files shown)


scottdraves:
this looks great, it solves the weight problem.

i think that the initHeight spec is no longer exactly respected, since the margins are added later.
to get a completely accurate result i think you need to solve an equation.
but this is a minor problem by comparison, i'll file another issue for it if we need it.

:+1:

mnavasiolava:
Do you mean initHeight for combinedplot? Actually it works as previously. Margins are not added to combined plot height, they just participate in equation to calculate subplot’s height. So the result initHeight of the combined plot will not change

scottdraves:
are you sure because i tried to make a plot with initHeight and initWidth both 800 and the result was not square, it was a bit taller than wide.

mnavasiolava:
yes, i'm sure. i've checked what is wrong - this is because of subplots' title: even if title is not provided/visible it has margins that influence on the result height. if you provide subplots titles combinedplot will become even taller.

PR #2718, opened 2015-10-14
Alice/2389

alice-kroch:
Beginning of implementing 'About Beaker' menu item

### DIFF
--- a/core/src/main/web/app/utils/electron.js
+++ b/core/src/main/web/app/utils/electron.js
@@ -80,6 +80,9 @@
             }
           },
           {
+            type: 'separator'
+          },
+          {
             label: 'Quit',
             click: function() {
               IPC.send('quit');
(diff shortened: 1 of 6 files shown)


scottdraves:
cool, looks like a good start.  it needs
1) to fill in the templates for getBuildTime() and getVersion()
2) to be large enough that it does not need to scroll.

alice-kroch:
Added version and build time, gave proper window size

scottdraves:
we need a solution that does not duplicate the code.
it should use the same HTML as the regular web version does,
just in a top-level window.

alice-kroch:
We can't use the same HTML because it's modal window with own controller and it works without any URL changes. To render top-level window in electron it's required to use URL to render the content (as it is in case of all notebook windows), or to build new native window in electron app - as I did. 

For example window with keyboard shortcuts - when you open it it has it's own url and opens in separate window. So it works for electron application - because it's static window with it's own URL. For 'About' instance it doesn't work like this - because it's modal, not static window, that doesn't affect current URL.

PR #2721, opened 2015-10-14
Tom/2709 - Waiting for evaluators to load before rendering workspace

tomca32:


scottdraves:
don't block display of the whole notebook.

Issue #2684, opened 2015-10-09
easyform on pub server css problems


scottdraves:
css problems, there is garbage at the bottom of the page,
even more if you click the date picker.

https://pub.beakernotebook.com/#/publications/560ca655-12df-43b9-9ecd-5714766b8fc4?fullscreen=true

maxyeg:
In js console: ![8abef-clip-21kb](https://cloud.githubusercontent.com/assets/2286032/10396903/9ca5683e-6ead-11e5-98e2-3bf291626697.png)

looks like it lacks some ui resources when calls jquery autocomplete from easy-form impl.

sethkrasnianski:
We're also not including the datepicker's css file. When manually injecting it, it fixes all that garbage at the bottom

![screen shot 2015-10-09 at 10 58 46 am](https://cloud.githubusercontent.com/assets/1910114/10397283/cd0766aa-6e74-11e5-967c-36efe7871415.png)

sethkrasnianski:
Just need to get those assets squared away

sethkrasnianski:
Assets and css issues aside, it seems that at the time any form field which talks to the `easyform/set` endpoint is failing due to `window.languageServiceBase` not being accessible. Also,`evaluatorId` is not set at this time.

The culprit in the example of the combo boxes is line `11506` 

```
url: window.languageServiceBase[evaluatorId] + '/easyform/set'
```

When you select an option from the combo box, an ajax request it kicked off to that endpoint. The combination of `languageServiceBase` and `evaluatorId` being undefined prevents the request from occurring. This is most likely not be specific to the combo box only. Any EasyForm widgets that communicate with the `easyform` endpoint, and use those two values, are at risk to fail as well.

scottdraves:
on the pub server the form should be read-only.

planyon:
After downloading the publication and running it on my local copy of Beaker I'm seeing the same issue, with improperly displayed Date Pickers on Easy Forms.

scottdraves:
i can't reproduce the problem in local beaker.  what version do you have paul?

planyon:
I just pulled down the latest master, downloaded the notebook, opened it up and ran all cells. See attached screenshot. I had Seth try to do the same test and it worked for him, but Matt tried and he experienced the same issue as on publications and in my screenshot.

[...]

PR #2725, opened 2015-10-15
core: Don't crop easy form select elements

sethkrasnianski:
Supports https://github.com/twosigma/beaker-notebook/issues/2684


ildipo:
:+1:

Issue #2705, opened 2015-10-12
TimePlot autorange bug


scottdraves:
requested 0 in range but seems to be ignored:
https://pub.beakernotebook.com/#/publications/561c2126-6107-43d6-a0f4-1bd97b182d6e

PR #2728, opened 2015-10-15
TimePlot autorange bug #2705

mnavasiolava:
Added support of autoRangeIncludesZero property on frontend.
I'd like to note that this property is set to true on backend by default. And now after this property started working usual plots (e.g. in charting tutorial) look different. Maybe it makes sense to set autoRangeIncludesZero to false by default?


scottdraves:
you are right it should be false by default.

scottdraves:
:+1:

Issue #2688, opened 2015-10-09
typeface for autocomplete is not roboto mono


scottdraves:
<img width="216" alt="screen shot 2015-10-09 at 11 21 42 am" src="https://cloud.githubusercontent.com/assets/963093/10397879/f69cdf38-6e77-11e5-9ab8-0e3a4dcda4f8.png">

PR #2729, opened 2015-10-15
#2688 - Sets the autocomplete font to Roboto Mono

tomca32:


scottdraves:
possible to use $font-family-mono as defined in variables.scss?

Issue #2730, opened 2015-10-15
Header repaints on scroll


tomca32:
The whole header repaints on scroll. This is unnecessary and affects performance.

To reproduce enable paint flashing in Chrome and scroll the page. Green rectangle marks the areas that are constantly repainted.

![Repainting Header](http://cdn.makeagif.com/media/10-15-2015/-SeT-5.gif)

Promoting the header to a composite layer should solve the issue.

PR #2734, opened 2015-10-15
#2730 - Promoted the header to a layer to prevent repainting on scroll

tomca32:
Fixed elements repaint when content underneath them changes. To prevent repainting they have to be promoted to a composite layer.

More about this issue: http://benfrain.com/improving-css-performance-fixed-position-elements/

### DIFF
--- a/core/src/main/web/app/styles/header.scss
+++ b/core/src/main/web/app/styles/header.scss
@@ -24,6 +24,8 @@
 }
 
 header {
+  @include promote-to-layer();
+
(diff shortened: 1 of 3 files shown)


ildipo:
Might be better to put this as an added style on our navbar rather than changing the bootstrap css since when we'll update bootstrap this change will go away

tomca32:
Oh wow, I'm blind. Yes of course, didn't even notice that was bootstrap.

PR #2741, opened 2015-10-16
Maxl/#2710

maksim-lauryk:
added new features clojure demo notebook
needs the issue #2476 Clojure require configuration


ildipo:
:+1:

Issue #2762, opened 2015-10-19
Refactoring for Angular 1.4.x upgrade


tomca32:
In order to upgrade to  the latest stable angular version we need to refactor some pieces of the application. We are doing some non-standard operations which aren't supported by Angular. The biggest offenders are:
- mutating the resolve object in $routeProvider config - angular will copy the object when instantiated and ignore the mutations
- Accessing $scope.$root instead of $rootScope - $scope.$root can be null

Also there are still a couple of endpoints which return plaintext but with application/json content-type header.

scottdraves:
https://github.com/twosigma/beaker-notebook/pull/2763

PR #2763, opened 2015-10-19
#2762 - Refactoring in preparation for Angular 1.4.x

tomca32:


scottdraves:
crashes on the R and Py tuts with

```
TypeError: Cannot read property '_beaker_model_output_result' of undefined
    at Object._notebookModel.set (beakerApp.js:6824)
```

Issue #1352, opened 2015-03-29
short name for LaTeX should be Tx


scottdraves:
instead of La

scottdraves:
actually this is more TeX or KaTeX than LaTeX so how about Tx?  too bad we can't TX

scottdraves:
and long name should be TeX

scottdraves:
The key here is to rename the plugin when loading old notebooks.

scottdraves:
similarly rename Html -> HTML

scottdraves:
actuall forget that, this PR is good, make another PR for html under this issue: https://github.com/twosigma/beaker-notebook/issues/515

PR #2767, opened 2015-10-20
#1352 short name for LaTeX should be Tx

alice-kroch:
Please, check


scottdraves:
this looks good, at the end of config/tutorials/text.bkr please also change the section name to TeX from LaTeX.

scottdraves:
actually, this reminds me, in the same PR can you please also rename Html to HTML?

scottdraves:
:+1:

Issue #515, opened 2014-08-29
Html should be HTML


scottdraves:
![screen shot 2014-08-28 at 10 51 32 pm](https://cloud.githubusercontent.com/assets/963093/4085928/a4db9a1c-2f27-11e4-86e0-2b921dc200df.png)

scottdraves:
And Latex should be LaTeX

PR #2776, opened 2015-10-21
#515 Html should be HTML

alice-kroch:
Done. Please, check


tomca32:
Tests on CI failed: 
- No element found using locator: By.cssSelector(".plugin-manager .Html")

Seems like that caused a problem css class names.

alice-kroch:
Fixed.

scottdraves:
:+1:

Issue #2649, opened 2015-10-07
blueimp-canvas-to-blob and blueimp-load-image maybe redundant


jonassmedegaard:
It seems at least parts of blueimp projects are embedded in FileAPI.js of ng-file-upload.

Please try if possible to reduce avoid those possibly redundant dependencies.

scottdraves:
Not sure what you are suggesting here.  we have copies of the stuff downloaded from bower in our tree just to make us less dependent on their servers.  presumably you are replacing this with debian versions.
if there is a debian version of ng-file-upload then you can ignore ours and we are good, right?

jonassmedegaard:
Quoting Scott Draves (2015-10-07 23:30:12)

> Not sure what you are suggesting here. we have copies of the stuff 
> downloaded from bower in our tree just to make us less dependent on 
> their servers.
> presumably you are replacing this with debian versions.
> if there is a debian version of ng-file-upload then you can ignore 
> ours and we are good, right?

It seems src/main/web/vendor/bower_components/ng-file-upload/FileAPI.js 
is a superset of 
core/src/main/web/vendor/bower_components/blueimp-canvas-to-blob/js/canvas-to-blob.js 
and 
core/src/main/web/vendor/bower_components/blueimp-load-image/js/load-image-ios.js 
- and possibly of other files below 
  core/src/main/web/vendor/bower_components/blueimp-load-image as well.

If ng-file-upload can be used for all (not only some) of the cases where 
blueimp-canvas-to-blob and blueimp-load-image is currently used, then 
that should give a slight speedup and slightly lower memory consumption.  
Benefit for me would be 2 packages less to package.

But you are right, we need not change anything here - nothing is broken, 
just (possibly) less optimized.
- Jonas

## 
- Jonas Smedegaard - idealist & Internet-arkitekt
- Tlf.: +45 40843136  Website: http://dr.jones.dk/
  
  [x] quote me freely  [ ] ask before reusing  [ ] keep private

maxyeg:
We have `blueimp-canvas-to-blob`, `blueimp-load-image` and `blueimp-tmpl` included, because they are dependencies of `jquery-file-upload`. 
https://github.com/blueimp/jQuery-File-Upload/blob/master/bower.json#L44

Beaker uses `jquery-file-upload` to import notebooks.
https://github.com/twosigma/beaker-notebook/blob/master/core/src/main/web/app/helpers/core.js#L216

[...]

PR #2788, opened 2015-10-22
#2649. blueimp-canvas-to-blob and blueimp-load-image maybe redundant

maxyeg:
Remove jquery-file-upload plugin.
Make import notebook code use ng-file-upload.


scottdraves:
works well except the drop-zone seems to be limited.  on master i can drag anywhere in the window, but with this change, only on the top-half (below the notebook?).

scottdraves:
:+1:

PR #2797, opened 2015-10-22
Increased RequireJS timeout to 30s from the default 7s

tomca32:
This fixes the missing menus on large notebooks.

### DIFF
--- a/core/src/main/web/app/beaker.js
+++ b/core/src/main/web/app/beaker.js
@@ -157,7 +157,8 @@
       'bk.helper',
       'bk.utils',
       'bk.publication',
-      'bk.electron'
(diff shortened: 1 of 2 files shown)


scottdraves:
this sounds good but why 30s?
why is it timing out at all?
what is blocking it?

tomca32:
My best guess now is that the initial paint is blocking it. Initially the browser has to paint the whole notebook with a ton of cells, tables and other outputs and that locks it up. This will be addressed in performance optimizations and is one of the things I want to measure but at this moment it seemed critical to make the large notebooks usable so increasing the timeout is a stop-gap measure.

PR #2820, opened 2015-10-26
 #1806 clojure crashes on the following code

maksim-lauryk:
up to date with master
such exceptions will be shown in the cell output


scottdraves:
:+1:

Issue #472, opened 2014-08-14
consider replacing RJSONIO with jsonlite


scottdraves:
http://cran.r-project.org/web/packages/jsonlite/index.html

seems to be improved and more maintained

scottdraves:
though maybe it does not handle unboxed values like rjsonio does, so maybe a lose.

scottdraves:
maybe best to use jsonlite to convert TO json (since it has support for atomic vectors) and rjsonio to convert back (since it doesn't barf on scalars).

scottdraves:
first question is can this be done without changing semantics?

scottdraves:
https://github.com/rstudio/shiny/issues/572

scottdraves:
not needed

PR #2827, opened 2015-10-27
 #472. consider replacing RJSONIO with jsonlite

maxyeg:


scottdraves:
the handling of scalars and vectors looks good, but 
needs to run the notebooks in the test/notebooks subdirectory.

PR #2833, opened 2015-10-29
Setup performance testing - Tom/2807

tomca32:
Measuring time to render performance_noteboook as well as some rendering stats for section showing/hiding like:
- dropped frame count
- layout time
- mean frame time

to run the tests just run ./test/perftest

This is not a part of our build. They take a while to run and it would slow down the build too much.


scottdraves:
looks great.  please document perftest in README

tomca32:
Excellent, I'll update drool to the new version. Thanks!

PR #2854, opened 2015-11-04
Maxy/265

maxyeg:
Fixed typo and bug when use plugin env variables. 
But we retrieve `pluginEnvps` from config and they are never filled there. 
https://github.com/twosigma/beaker-notebook/blob/master/core/src/main/java/com/twosigma/beaker/core/module/config/DefaultBeakerConfig.java#L160
From where we should populated this map?


scottdraves:
:+1:

Issue #2692, opened 2015-10-10
less saturated code highlight colors


scottdraves:
beaker is on the left.  github on the right.  it gets more keywords like document and window too.

especially bad is beaker's bright green matching paren.
<img width="899" alt="screen shot 2015-10-09 at 9 36 33 pm" src="https://cloud.githubusercontent.com/assets/963093/10408553/6a776f12-6ece-11e5-8368-52614fa952f9.png">

scottdraves:
alice, please start out with a PR for just fixing the matching-paren green.
i think we should also use some bold, in addition to highlighting with color.
watch out for this though: https://github.com/codemirror/CodeMirror/issues/3603

scottdraves:
this should be done by adding a new theme.

PR #2859, opened 2015-11-04
#2692 less saturated code highlight colors

alice-kroch:
Please, check


scottdraves:
please skip bold until we can be sure to avoid https://github.com/codemirror/CodeMirror/issues/3603

PR #2862, opened 2015-11-04
Removing mouseleave listener on destroy

tomca32:
As @samccone pointed out here: https://github.com/twosigma/beaker-notebook/pull/2853

I didn't remove the listener when the scope gets destroyed. This PR fixes that.


samccone:
:+1: :clap:

PR #2866, opened 2015-11-05
update D3

maxyeg:
Update d3

Example notebook:
https://pub.beakernotebook.com/#/publications/563c9646-2277-44c2-89d5-4e16188fc7ac?fullscreen=true


scottdraves:
1) looks good but can you reproduce the style of the example (solid/merged drops, zoomable with mouse wheel).
2) don't include eventdrops inside of beaker, load it like we do the p5.js example (sorry i wasn't clear about this).

scottdraves:
OK much better -- but we only want the d3 changes for this repo, the notebook itself should go into the examples repo where i already merged it.

scottdraves:
:+1:

Issue #2864, opened 2015-11-04
level of detail on points should default to boxes


scottdraves:
just like how we changed it for lines to default to river.

PR #2867, opened 2015-11-05
level of detail on points should default to boxes #2864

mnavasiolava:
to see changes you will need to recreate the cell


scottdraves:
:+1:

Issue #2334, opened 2015-09-14
performance optimization: lazy instantiation of outputs


scottdraves:
lazy instantiation of outputs.  just like we only turn code blocks into code mirrors when they become visible on the screen, similarly we should only render outputs into plots and tables when they first become visible.  hopefully this can improve how long it takes to load large notebooks.

PR #2879, opened 2015-11-06
Issue #2334 - Lazy output instantiation Tom/2334

tomca32:
This gives us a dramatic performance boost on notebook loading. However, the drawback is that the scroll lags because the browser needs to render the output cells during scroll.
This is very noticeable in the performance notebook in the Giant Tables section since there is a table there with 1000 columns that takes a while to render once you scroll to it.

Still, I believe that this is a good tradeoff.


scottdraves:
:dolphin: :dolphin:

Issue #2679, opened 2015-10-09
click filename to change 


scottdraves:
<img width="259" alt="screen shot 2015-10-08 at 11 42 55 pm" src="https://cloud.githubusercontent.com/assets/963093/10385405/7f2d060a-6e16-11e5-9ba7-4feb68a1ebfc.png">

if you click on "/Users/spot/foo.bkr" it should open a dialog to rename the notebook.
open the regular filesystem dialog, but instead of writing a copy like save-as,
mv the original atomically (if possible, otherwise copy & delete (make sure they
are different!)).

should highlight on hover.

PR #2902, opened 2015-11-10
#2679. click filename to change

maxyeg:


scottdraves:
cool.  but:
1) if the file has not been saved yet, this will crash.  in that case, clicking on "New Notebook" should just be like an orderinary save.
2) if the file has been saved, the button at the bottom of the dialog should read "Rename" instead of "Save"

scottdraves:
if the document is unsaved and i rename it, then the \* in the title of the window should not vanish.

tomca32:
On the frontend we have a couple of modules/files that are essentially dumping grounds for everything, such as mainapp.js that is edited here.

I feel like we have an opportunity here to make the codebase a bit healthier by separating this "I/O" stuff into its own service - fileManipulation.js or something. I think it would make sense to have loading/saving/renaming logic in there rather than in a file called "mainapp.js".

If it's not a big refactor and I don't feel it is, maybe we could consider it.

scottdraves:
sure that's fine.

scottdraves:
:+1:

Issue #2546, opened 2015-09-30
shift-enter


scottdraves:
which evalutes and moves to next cell, should move to next _code_ cell, not into any text or section cell

PR #2908, opened 2015-11-11
shift-enter #2546

mnavasiolava:
i've also added the same behavior for shift-ctrl/cmd-enter if it evaluates the last line in the cell


scottdraves:
:+1:

Issue #2733, opened 2015-10-15
Wrong encoding on import of non-ascii .bkr file


altavir:
Loading via File -> open works fine, but if I try to drag and and drop file to main window (both web-version and electron) or upload it, all Russian letters are replaced with `?`-s. The .bkr file itself is in UTF-8 (I've explicitly checked that) and system encoding is cp1251.

scottdraves:
thanks for the report.  can you please provide notebook that reproduces this problem?
emailing an attachment to me or beaker-feedback at twosigma is fine.

altavir:
Here it is: http://sharing.beakernotebook.com/gist/anonymous/d3bec42962f02963ebbf . I just published it to server to show to students who failed to install beaker.

scottdraves:
indeed.  we'll get to this one later since there seems to be a decent workaround.

PR #2909, opened 2015-11-11
Wrong encoding on import of non-ascii .bkr file #2733

mnavasiolava:


scottdraves:
:+1:

scottdraves:
This looks good but it actually shows a problem if I import notebooks such as
https://pub.beakernotebook.com/#/publications/560b54da-3c14-4890-8b64-b204bc798320?fullscreen=false or
https://pub.beakernotebook.com/#/publications/560c510c-16a8-4a22-9bed-93071e9f88e8?fullscreen=false
The former has some chinese characters, and the latter has some non-breaking spaces.
The chinese characters appear as ? and the space as �.  So maybe we are not encoding correctly when we save?

mnavasiolava:
You are right, downloaded notebooks are not encoded correctly, and that's why they are opened with ?? symbols

scottdraves:
:+1:

PR #2910, opened 2015-11-11
Cit/2882

kpaharelau:


core/src/main/java/com/twosigma/beaker/core/rest/PluginServiceLocatorRest.java
```diff
@@ -599,11 +600,19 @@ private String hashIPythonPassword(String password, String pluginId, String comm
     // but it would be nice if there were a way to make this explicit. XXX
     bw.write("print(passwd('" + password + "'))\n");
     bw.close();
-    String hash = br.readLine();
-    if (null == hash) {
-      throw new RuntimeException("unable to get IPython hash");
+
+    Object[] array = br.lines().filter(new Predicate<String>() {
```
tomca32: Can this have a bit better name? Calling it "array" is not going to be very helpful to anyone that's going to work with this code later. Maybe calling it something like "hashes" or "linesWithSHA" would be better?

tomca32:
Is it possible that this PR slowed down Python initialization? Tests seem to time out on loading a small notebook after 12 seconds. I don't think it's flakiness.

scottdraves:
worked fine for me and i did not experience any slowdown. agreed that "array" should be improved.

kpaharelau:
I improved performance. Array does not using now

tomca32:
Great thanks. I don't understand why is that one test failing all the time. Let me check it out

kpaharelau:
May be some ipython process hangs in memory

tomca32:
Yeah maybe. It works fine on my machine but on CI it seems to timeout waiting for notebook to load. And it's a small notebook

kpaharelau:
try to call killall ipython on CI

scottdraves:
:+1:

Issue #2704, opened 2015-10-12
TimePlot class should accept java.util.Date


scottdraves:
not just java.lang.Number for X axis values.

altavir:
Consider working with Instants or LocalDateTimes instead of Dates. Java 8 Time library is much more convenient than legacy date.

scottdraves:
agreed, thanks.

PR #2916, opened 2015-11-12
TimePlot class should accept java.util.Date #2704

mnavasiolava:
added support of java.util.Date and java.times.LocalDateTime objects for x coordinates of the TimePlot.
examples can be found in the charting tutorial

### DIFF
--- a/core/config/tutorials/chartingTutorial.bkr
+++ b/core/config/tutorials/chartingTutorial.bkr
@@ -40,7 +40,7 @@
                     "background": "#E0FFE0"
                 }
             },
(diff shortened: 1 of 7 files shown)


scottdraves:
:+1:

scottdraves:
please remove Java8 feature LocalDateTime, that has to wait, sorry.

scottdraves:
You can leave the code for it in, just comment it out, because we'll need this later but if we use a branch i think it will get lost.

mnavasiolava:
done

scottdraves:
:+1:

PR #2919, opened 2015-11-12
Alice/448

alice-kroch:
I/O power button on initialization cell. Please, check


scottdraves:
:+1: 
i will merge this as it is.  but please make another PR with these improvements:
1) if the code is hidden, then show this indicator in the summary of the code, to the right of the number of lines.
2) if you click on the indicator, bring up a menu with one item that says "disable initialization"

Issue #2878, opened 2015-11-06
Cannot find julia on OS X


xiaoweiz:
julia version: 0.4
python version: 3.5 from Anaconda 2.4

`IJulia` is installed and `julia` can be used with `jupyter-notebook`. 

`R`, on the other hand, can be found by beaker. I've add the path info of `julia` in `beaker.pref.json` but it still doesn't work.

scottdraves:
Thanks for the report.  Unfortunately Julia put the version number into the kernel name, and Beaker is requesting version 0.3.  You can hack around this by editing julia.js line 89, which you can find inside the application.  I can tell you how more precisely if you say your OS and if you are using our Electron version.

Really Beaker should query the Jupyter backend so it can always match.

xiaoweiz:
Just found a similar issue #1602. 

Revising the line [https://github.com/twosigma/beaker-notebook/blob/master/plugin/ipythonPlugins/src/dist/julia/julia.js#L89] to change the kernel name from `julia-0.3` to `julia-0.4` solves the problem.

xiaoweiz:
Thanks for your quick response. I was typing my own finding while you're replying! :)

scottdraves:
no problem, thanks for the report.  since julia has moved 0.4 up to their default download we should at least update our default, if not query.

scottdraves:
cit, please find the API call jupyter uses to get the list of available kernels, and call find find a kernel that starts with "julia" and then use that to start our backend.

PR #2933, opened 2015-11-13
Cannot find julia on OS X #2878

kpaharelau:
I have created rest method for a receiving information about all jupiter kernels.  I am calling 'jupyter kernelspec list --json' in the rest method.


scottdraves:
unfortunately i don't think this is quite right.  you would need to set up the path the same as PluginServiceLocatorRest.java in order to be sure "jupyter" finds the right executable.

you can avoid that because the right jupyter has already been started.
just ask it which kernels it has, it must have a REST call of its own.
use network panel of chrome debugger to reverse engineer it.
also, look in https://github.com/twosigma/beaker-notebook/tree/master/plugin/ipythonPlugins/src/dist/vendor/ipython3
there is likely to be a JS wrapper for this call that you can use.

scottdraves:
I get:

```
Julia-stderr>WARNING:tornado.access:404 GET /api/kernelspecs (127.0.0.1) 0.40ms referer=http://127.0.0.1:8809/beaker/
```

when ipython is v2.  check the version and if less than 3 then use the default guess (now 0.4).

scottdraves:
cit can you please bring this branch up to date with master?
i can't test it well it is missing the TERM fix i think.
on my current laptop i cannot even run python with this branch but it works on master.

scottdraves:
:+1:

Issue #2757, opened 2015-10-18
please consider use request-frame polyfill instead of gist


jonassmedegaard:
Hi,

You currently use this gist from Paul Irish: https://gist.github.com/paulirish/1579671

A proper project has been formed, based on same gist: https://github.com/julienetie/request-frame

Please consider switch to use that, as I expect that to more likely be community maintained.

scottdraves:
thanks, sounds like a good idea.

PR #2934, opened 2015-11-13
request-frame polyfill instead of gist #2757

mnavasiolava:


core/src/main/web/app/template/index_template.html
```diff
@@ -109,6 +109,12 @@
   <script src="vendor/bower_components/datetimepicker/jquery.datetimepicker.js"></script>
   <!-- endbuild -->
 
+  <!-- htmlbuild:embedremove -->
+  <script>
+    (function(w){ w.rAF = requestFrame('request'); })(window);
```
scottdraves: "rAF" is too short a name for a global.  please call it "requestFrame"

scottdraves: and does it really have to be a global??

tomca32: It should be a global because it's filling for a browser API. Most modern browsers have window.requestAnimationFrame, this is supposed to implement it for a browser that doesn't have it.

However, the standard API is requestAnimationFrame and I don't think we should pollute the global namespace with an additional value. Request-frame docs say that we can just use:

```
requestFrame('native'); // re/ declares requestAnimationFrame & cancelAnimationFrame
```

without assigning it to any value. Is there a reason why we're not doing that?

mnavasiolava: i think you right, @tomca32 
i checked once again: the use of requestFrame('native') + window.requestAnimationFrame is the equivalent of how it worked when we used rAf.js
i will make the necessary changes

tomca32: Great thanks!

scottdraves:
:+1:

Issue #2925, opened 2015-11-12
down arrow on last line, with hidden source


scottdraves:
if you type the down arrow button on the last line of a cell,
and the next code cell has the input hidden, then the cursor should skip over the closed cell
and move on to the next cell.

PR #2935, opened 2015-11-13
#2925. down arrow on last line, with hidden source

maxyeg:
Please, inform me if we want to skip cells with hidden input when pressed Shift+Enter(evaluates current cell and goes to the next one).


scottdraves:
:+1:

PR #2936, opened 2015-11-13
fixed 2492

kpaharelau:
Codemirror has been upgraded to 5.8.0 version


scottdraves:
holding off on this until we have a plan for reindent-region.
i thought upgrading was required for that but cit says it has not been supported in a long time.

scottdraves:
turns out this is not needed for reindent.

Issue #448, opened 2014-08-12
Need easier way to tell if a cell is 'initialization cell' in new UI design


anglee:
Right now, one can click to open the cell menu and see if the 'initialization cell' item is checked or not. In the previous version of the UI, one can tell by the cell bracket color.
before
![image](https://cloud.githubusercontent.com/assets/1767921/3891367/474d9036-2229-11e4-838f-9094a32d8f5f.png)
after
![image](https://cloud.githubusercontent.com/assets/1767921/3891374/546f56a0-2229-11e4-9416-ebafd92db75d.png)

jchendy:
I'm not sure if y'all modified the design on purpose, but that was the intent of the I/O button in the bottom right corner on cells in the original design:

![bkinit](https://cloud.githubusercontent.com/assets/1740497/3891464/42224920-222a-11e4-8106-91bb894d790b.PNG)

anglee:
Ah, thanks for pointing it out, @jchendy! I didn't see that I/O button in the current implementation. @samccone, was this implemented or I missed something?

samccone:
It was not implemented, I think it must have been lost in one of the design iterations. 

currently you can tell if a cell is ready based on the fact that there is no spinner in the top right menu of each cell.

Perhaps that is not enough.

anglee:
A cell being ready and being an 'initialization cell' are different ideas. 'Initialization cell' is a tag that one can use to tag a cell. The cells that are tagged as "initialization cell" will be evaluate upon notebook open.

melmac48:
agreed, this was part of the initial design, we should add it back in. @samccone  please check with Ang if you need further clarification on how to identify an cell which is marked to run on open.

scottdraves:
in addition, would be nice to show tags more clearly than in the options dialog.
i am thinking small gray text just above the cell box, on the right, maybe with some UI for adding and removing.

scottdraves:
still need:
1) if the code is hidden, then show this indicator in the summary of the code, to the right of the number of lines.
2) if you click on the indicator, bring up a menu with one item that says "disable initialization"

PR #2949, opened 2015-11-16
#448 Need easier way to tell if a cell is 'initialization cell' in new UI design

alice-kroch:
1) if the code is hidden, then show this indicator in the summary of the code, to the right of the number of lines.
2) if you click on the indicator, bring up a menu with one item that says "disable initialization"

Please, check

### DIFF
--- a/core/src/main/web/app/mainapp/components/notebook/cell.jst.html
+++ b/core/src/main/web/app/mainapp/components/notebook/cell.jst.html
@@ -49,12 +49,5 @@
       <div class="loading-spinner rotating"></div>
     </div>
   </div>
-
-  <div data-ng-if="cellmodel.type=='code'" class="power-menu" data-ng-class="{'input-hidden': cellmodel.input.hidden}">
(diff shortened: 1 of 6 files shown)


tomca32:
Test failure seems legitimate. I guess you changed some behavior of cell menus in which case you will need to change the test as well.

Try running the test locally and see what happens. To run tests just do:
    cd test
    ./runner

However, this will run all tests which takes a long time. You can easily run a single test with a simple change. Using the failing test as an example.
This test is located in test/tests/cell-menu.js starting at line 35. Just change the line 35 from:
    it('closes when another menu is opened', function(done) {
to:
    fit('closes when another menu is opened', function(done) {

So yeah it's just a one character change. Now when you do ./runner only this test will run which should take just a couple of seconds. Don't forget to change it back to "it" before the push though.

scottdraves:
put about 5mm space to the left of the on/off symbol

screen shot 2015-11-18 at 11 24 31 am

scottdraves:
:+1:

Issue #1343, opened 2015-03-27
delete all output cell should NOT delete the one under execution


ildipo:
Start a 'long running cell'
use the 'delete all output cells' menu
-> you cannot interrupt the long running cell anymore.

samccone:
the goal here is to not delete running cells correct @ildipo ?

ildipo:
yea. and remove from the evaluation queue all cells you delete

PR #2959, opened 2015-11-17
delete all output cell should NOT delete the one under execution #1343

mnavasiolava:


scottdraves:
this doesn't work for me.  try this:
open the JavaScript tutorial.
run the 2nd cell (start the game).
Delete all output cells.
see that you cannot cancel anymore.

mnavasiolava:
fixed. also removed deleted cells from evaluation queue as Davide asked

scottdraves:
it looks like it will not remove it from the evaluation queue if it has no output, i think this is a mistake.

mnavasiolava:
but "delete all output cell" concerns only cells with output.
and btw, if the cell is in the evaluation queue, it will always have output "pending" or "running" and that's why will be deleted. am i wrong?

scottdraves:
:+1: you are right, thanks.

PR #2960, opened 2015-11-17
Minor refactoring of mainapp.js

maxyeg:
Move save/saveas/rename logic to different place

I moved methods which don't use scope.
I don't move methods which work with scope.
Please tell me if more cleaning is required.


tomca32:
Excellent! I think that this kind of small refactoring is really important to keep the codebase healthy. Thanks.

scottdraves:
:+1:

PR #2965, opened 2015-11-18
fix scrollbar jumping, save output heights

alice-kroch:
Please, check

### DIFF
--- a/core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
+++ b/core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
@@ -98,7 +98,7 @@
           var style,
               height = $scope.cellmodel.output && $scope.cellmodel.output.height;
           if (height)
-            style = {'height': height + 'px'};
+            style = {'min-height': height + 'px'};
 
           return style;
         };
@@ -403,11 +403,11 @@
         });
 
         scope.$watch('displayOutput', function(value) {
-          if (!value) return;
+          if (!value || !scope.hasOutput()) return;
 
           var output = element.find('.code-cell-output');
           $timeout(function() {
-            if (scope.hasOutput() && output)
+            if (output)
               scope.cellmodel.output.height = output[0].offsetHeight;
           }, 100);
         });
--- a/core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
+++ b/core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
@@ -94,13 +94,14 @@
           return $scope.cellmodel.output.result !== undefined;
         };
 
-        $scope.$watch('cellmodel.output.result', function () {
-          var output = $element.find('.code-cell-output');
-          $timeout(function() {
-            var height = output.height();
-            _.extend($scope.cellmodel.output, {height: height});
-          }, 100);
-        });
+        $scope.getOutputHeight = function() {
+          var style,
+              height = $scope.cellmodel.output && $scope.cellmodel.output.height;
+          if (height)
+            style = {'height': height + 'px'};
+
+          return style;
+        };
 
         $scope.backgroundClick = function(event) {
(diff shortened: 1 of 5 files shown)


core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
```diff
@@ -395,6 +402,16 @@
           }, 1);
         });
 
+        scope.$watch('displayOutput', function(value) {
+          if (!value || !scope.hasOutput()) return;
+
+          var output = element.find('.code-cell-output');
+          $timeout(function() {
```
tomca32: Just curious. Why is this timeout here?

core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
```diff
@@ -94,6 +94,15 @@
           return $scope.cellmodel.output.result !== undefined;
         };
 
+        $scope.getOutputHeight = function() {
+          var style,
+              height = $scope.cellmodel.output && $scope.cellmodel.output.height;
```
scottdraves: use "var" for every variable.

core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
```diff
@@ -94,6 +94,15 @@
           return $scope.cellmodel.output.result !== undefined;
         };
 
+        $scope.getOutputHeight = function() {
```
tomca32: If you put a console.log here, you'll see that this function gets run a lot of times, we would want to avoid that if possible.

tomca32:
This works but there are a couple of changes that would make it better.

First, we should really try to not use watchers, also we're getting height and setting style all the time - try putting a console.log into getOutputHeight function to see that. Lastly, we probably don't need that timeout.

Right above the displayOutput watcher in codecell-directive is a Scrollin tracker. That handler runs once - when the element enters the viewport. You can just call getOutputHeight from there and it will run only once per output.  That way we also don't need the watcher or the timeout. Makes sense?

Otherwise, it looks good. Thanks

alice-kroch:
We do need timeout because cell output element behaves in the DOM by ngIf directive. So that If you try to access element and it's height immediately you won't get it.

tomca32:
OK, so we don't want to depend on a timeout. There are a couple of reasons but generally timeouts are arbitrary numbers and we don't want to play the guessing game. Also, no matter what number we guess it's going to be wrong in some case, sooner or later.

So here's a couple of options to remove timeout:
1. If you need the element rendered before saving the height, try using requestAnimationFrame instead of timeout
2. Wrap display output div in a directive. The link function will be fired only after the element is ready so you can certainly get the height from the link function and save it there.
3. Experiment with [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) - I have never used these but I read somewhere that you can use them to listen for paints.

Of these I would probably prefer option 2 since I think that is the most stable solution. Link function fires after the element is ready so there are no unknowns there. If requestAnimationFrame works, I'm totally fine with that as well.

As for option 3 - sounds interesting. If you make it work I would be happy to learn something new.

core/src/main/web/app/mainapp/components/notebook/codecell.jst.html
```diff
@@ -36,6 +36,7 @@
         <div class="cell-power-button" title="initialization cell is on" data-ng-class="{'visible': isInitializationCell()}"><i class="fa fa-power-off"></i></div>
       </div>
     </div>
+    <div data-ng-if="!displayOutput || !hasOutput()" data-ng-style="{'min-height': (cellmodel.output && cellmodel.output.height) + 'px'}"></div>
```
tomca32, line 39: Good thinking about data- prefixing but we don't support any old browsers so let's just remove the prefixes so they don't cause confusion.

PR #2970, opened 2015-11-18
Added test that checks if IPython can be started

tomca32:
Build fails of course, that's correct because the build should be broken - IPython doesn't load


ildipo:
kick jenkins

ildipo:
kick jenkins

scottdraves:
kick jenkins

scottdraves:
kick jenkins

tomca32:
kick jenkins

Issue #2966, opened 2015-11-18
IPython fails to start


tomca32:
Steps:
1. Open Beaker landing page
2. Click New Default Notebook
3.  Wait for IPython to start
4. Observe the error alert

![screen shot 2015-11-18 at 10 28 40 am](https://cloud.githubusercontent.com/assets/3384784/11245138/30986f32-8ddf-11e5-9936-7e1aae6396b3.png)

tomca32:
Also in the browser console we get the following error:
http://127.0.0.1:8801/93017c3/ipython.732108985775/api/sessions/ Failed to load resource: the server responded with a status of 403 (Forbidden)

tomca32:
Seems to first happen in this commit: ed5a0ec37aa42452c21fab5a230b006a302627c9

kpaharelau:
I found little problem with windows quotes and fixed it.  Problem still exists for MacOS. I   have not  MacOS at the current moment

kpaharelau:
see my PR with fixes for windows os

https://github.com/twosigma/beaker-notebook/pull/2975

scottdraves:
https://github.com/twosigma/beaker-notebook/pull/2975

PR #2975, opened 2015-11-19
IPython fails to start #2966

kpaharelau:
This is a os dependent problem. In current moment tested for linux and windows os


kpaharelau:
Problem still exists for MacOS. I  do not have  MacOS at the current moment

scottdraves:
:+1:

KnowledgeGarden:
This thread is difficult (for me) to follow. Mentions "fixes" for windows. I'm running the windows download (recent) and still getting this very error. Did everything in terms of the config file; even installed ipython in python34 and python27; tried both in config file.  Just in case, my config is here: https://gist.github.com/KnowledgeGarden/3b2c714dd0c193ff8b75

Issue #2963, opened 2015-11-17
Y viewport not restored correctly


scottdraves:
If you adjust the viewport of a chart and then save the notebook, and then open the notebook, the resulting viewport is wrong in the Y coordinate only.

saved like this:
![screen shot 2015-11-17 at 1 39 29 pm](https://cloud.githubusercontent.com/assets/963093/11221021/dab9cb1e-8d30-11e5-8264-96fa0c6d16cb.png)

open and look and see this:
![screen shot 2015-11-17 at 1 39 54 pm](https://cloud.githubusercontent.com/assets/963093/11221022/daba5c64-8d30-11e5-8d23-1bb3971328d1.png)

```
                    "focus": {
                        "xl": 0.47381271098606886,
                        "xr": 0.4929045148928317
                    },
```

looks like the Y data is missing from the file completely.

For simple plots (not combined plots) the viewport is lost completely!

Issue #2930, opened 2015-11-12
remove plot memory across reruns


scottdraves:
currently plot resizes and and viewport changes and LoD settings are recorded so that even if you rerun the cell, they are reapplied.  this was a mistake -- if you run the cell again, it should always get the same results.  we do need to save the results of these interactions, but they are regular output.

PR #2976, opened 2015-11-19
Y viewport not restored correctly #2963

mnavasiolava:
changes in this PR:
1) PR #2930 was reverted, because it deleted save state functionality for plots not only on rerun.
2) plot memory across reruns was deleted
3) plot focus is restored from state

Please note that for now save state on Ctrl+S is broken, so please check using File-Save


scottdraves:
:+1:

Issue #2969, opened 2015-11-18
pasted tab should be converted to spaces


scottdraves:
when you paste a string into a code cell, any tab character should be converted to spaces.

scottdraves:
bad idea

PR #2977, opened 2015-11-19
#2969. pasted tab should be converted to spaces

maxyeg:


scottdraves:
this was requested by a user but the more i look at it the more i think it's a bad idea.
sorry i should have thought before you started....
but now i think this is better, if possible: https://github.com/twosigma/beaker-notebook/issues/2978

Issue #2972, opened 2015-11-18
modifying language settings needs spinner


scottdraves:
when you change the classpath or any of the settings in the language manager that result in restarting the backend, you should put the text "Restarting Groovy..." with a spinner in the upper right of the dialog (right justified in the same line as the title). and when it finishes it changes to say "Restarting Groovy... done".

PR #2981, opened 2015-11-23
#2972. modifying language settings needs spinner

maxyeg:


core/src/main/web/app/helpers/globals.js
```diff
@@ -24,7 +24,9 @@
       REQUIREJS_TIMEOUT: 30,
       RECONNECT_TIMEOUT: 90000, // 90 seconds
       EVENTS: {
-      	RECONNECT_FAILED: 'reconnect-failed'
+      	RECONNECT_FAILED: 'reconnect-failed',
+        LANGUAGE_MANAGER_SHOW_SPINNER: 'language-manager-show-spinner',
+        LANGUAGE_MANAGER_HIDE_SPINNER: 'language-manager-hide-spinner'
```
tomca32, line 29: Good place for putting these. I like it

core/src/main/web/app/mainapp/components/pluginmanager/pluginmanager-directive.js
```diff
@@ -145,5 +145,19 @@
       }
     };
 
+    $scope.$on(GLOBALS.EVENTS.LANGUAGE_MANAGER_SHOW_SPINNER, function(event, data) {
+      $scope.loading = true;
+      $scope.loadingMessage = 'Restarting ' + data.pluginName + '...';
+      if (!($scope.$$phase || $rootScope.$$phase))
+        $scope.$digest();
```
tomca32: $on handler would have to be in the digest loop already, no? Are you sure we need to kick off the digest loop here?

core/src/main/web/app/mainapp/components/pluginmanager/pluginmanager-directive.js
```diff
@@ -145,5 +145,19 @@
       }
     };
 
+    $scope.$on(GLOBALS.EVENTS.LANGUAGE_MANAGER_SHOW_SPINNER, function(event, data) {
```
tomca32: I prefer putting the handler on the rootScope and always using $emit instead of $broadcast, unless you really know what you're doing. 

http://toddmotto.com/all-about-angulars-emit-broadcast-on-publish-subscribing/

scottdraves:
please leave the text up after the operation is complete.
when it finishes it changes to say "Restarting Groovy... done".

scottdraves:
:+1:

core/src/main/web/app/mainapp/components/pluginmanager/pluginmanager-directive.js
```diff
@@ -145,5 +145,23 @@
...
+    $rootScope.$on(GLOBALS.EVENTS.LANGUAGE_MANAGER_HIDE_SPINNER, function(event, data) {
+      if (data.error) {
+        $scope.loadingMessage += 'failed';
+      } else {
+        $scope.loadingMessage += 'done';
+      }
+      $scope.showSpinner = false;
+      bkUtils.timeout(function() {
+        $scope.showMessage = false;
+      }, 3000);
```
tomca32, line 163: I just bumped into this, I guess I either didn't notice it, or didn't understand it before.

We seem to be guessing here that the language reset operation will take ~3s, right? Isn't there a mechanism in place that tells us when the environment is ready? Like when we first start the language, we poll to determine if backend is ready.

scottdraves, line 163: 3s is supposed to be how long it takes for the message to fade out after the reset is finished.

tomca32, line 163: Fair point. I thought the backend was async, but that doesn't seem to be the case.

Issue #2978, opened 2015-11-19
show tab with arrow to bar


scottdraves:
in the code mirror, any actual tab character should be displayed with http://www.fileformat.info/info/unicode/char/21e5/index.htm

PR #2983, opened 2015-11-24
#2978. show tab with arrow to bar

maxyeg:
Remind please, that in our code cells when we press Tab key it adds two spaces, not `\t`.

### DIFF
--- a/core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
+++ b/core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
@@ -379,6 +379,7 @@
           scope.cm = CodeMirror.fromTextArea(element.find('textarea')[0], codeMirrorOptions);
           scope.bkNotebook.registerCM(scope.cellmodel.id, scope.cm);
           scope.cm.on('change', changeHandler);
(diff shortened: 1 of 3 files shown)


maxyeg:
issues #2969 #2978

core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
```diff
@@ -379,6 +379,7 @@
           scope.cm = CodeMirror.fromTextArea(element.find('textarea')[0], codeMirrorOptions);
           scope.bkNotebook.registerCM(scope.cellmodel.id, scope.cm);
           scope.cm.on('change', changeHandler);
+          scope.cm.on('change', bkCoreManager.replaceTabsWithArrow);
```
tomca32, line 382: Maybe we can debounce this. If you're running this on every keystroke with a fast typist we could run into problems.

We use lodash, lodash has debounce: https://lodash.com/docs#debounce

scottdraves:
this is what i want actually, not to actually replace the tabs but to show them differently only:
https://codemirror.net/demo/visibletabs.html

maxyeg:
That's not what we need

PR #2985, opened 2015-11-24
#2978. show tab with arrow to bar

maxyeg:
Arrows for tabs using code mirror's css class.
They all disappear when Tab key is pressed, because this leads to autoIndent which is done using spaces, not tabs


scottdraves:
:+1:

PR #3000, opened 2015-11-30
installApp has been deprecated #2995

kpaharelau:
task installApp has been changed to task installDist

If you get error in the IntellyJ Idea IDE 
"Error when building : “Could not find property 'installDist'"

Please setup local gradle distribution  for a project more or equals  2.7

image


scottdraves:
:+1:

Issue #2982, opened 2015-11-23
warn user on closing unset language manager


scottdraves:
if the user edits the settings in language manager, and then closes the window without clicking "set", they should get a modal popup that says "Discard your changes to the settings?" ok/cancel

PR #3013, opened 2015-12-02
#2982. warn user on closing unset language manager

maxyeg:


scottdraves:
field should not turn red unless user closes the window without clicking "set".
then it should open the dialog (i could not find the dialog in this PR) and color the field.

scottdraves:
OK, I can get the dialog by clicking close at the bottom of the window.
but there are other ways of closing the window that do not bring up the dialog, eg ESC or clicking on bkg.
best would be to hook into manager close regardless of how it is done.
ok solution is to hook into each of these methods.

finally, still it should only turn red when the manager closes and unsaved changes are found.

scottdraves:
:+1:

scottdraves:
please bring up to date with master.

Issue #3001, opened 2015-11-30
init indicator improvements


scottdraves:
1) it's too light, please change color to #a0a0a0  

2) in advanced mode, if a cell has only 1 or 2 lines then the initialization indicator overlaps the cell menu.
move it to the right so it fits.

![screen shot 2015-11-30 at 10 26 42 am](https://cloud.githubusercontent.com/assets/963093/11475769/01233f8a-974d-11e5-96a2-f9b556613759.png)

PR #3015, opened 2015-12-02
Initialisation indicator improvements

alice-kroch:
Issue #3001. Please, check

### DIFF
--- a/core/src/main/web/app/mainapp/components/notebook/codecell.jst.html
+++ b/core/src/main/web/app/mainapp/components/notebook/codecell.jst.html
@@ -34,7 +34,7 @@
           {{ isJobCancellable() ? 'Stop' : 'Run' }}
         </a>
       </div>
-      <div class="power-menu" data-ng-class="{'input-hidden': cellmodel.input.hidden}">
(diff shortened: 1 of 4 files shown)


core/src/main/web/app/mainapp/components/notebook/codecell.jst.html
```diff
@@ -34,7 +34,7 @@
           {{ isJobCancellable() ? 'Stop' : 'Run' }}
         </a>
       </div>
-      <div class="power-menu" data-ng-class="{'input-hidden': cellmodel.input.hidden}">
```
scottdraves, line 37: don't ever use the data- prefix we don't support old browsers.

scottdraves:
i appreciate what you are doing but please follow the issue instrucitons exactly.  i think this will reduce performance, plus it is a little funny to have it animate.

the issue specifies moving the indicator to the _right_.  position should be constant, not depending on the lines.

scottdraves:
also, don't re-use the old branch.  this is a new issue so branch should be named "alice/3001"

Issue #2946, opened 2015-11-15
beaker object becomes disconnected


scottdraves:
I do not have a good way to reproduce this, it is pretty rare but it does happen and of course this is a big problem when it does because autotranslation stops working in a notebook, instead you just get errors like:

```
channel not found for session 0wTxtB
channel not found for session 0wTxtB
channel not found for session 0wTxtB
channel not found for session 4Ha29c
channel not found for session 4Ha29c
channel not found for session 4Ha29c
channel not found for session 4Ha29c
channel not found for session 4Ha29c
channel not found for session 4Ha29c
```

It seems to be more likely to happen if I am running Beaker and put my laptop to sleep and then come to to existing sessions.

PR #3016, opened 2015-12-02
beaker object becomes disconnected #2946

mnavasiolava:
resubscribe to the channel after handshake
more detailed logging was added


scottdraves:
:+1:

Issue #2994, opened 2015-11-28
do not copy default.bkr to ~/.beaker


scottdraves:
similar to https://github.com/twosigma/beaker-notebook/issues/101
right now we cannot update the default notebook.
we should not copy it, leave that to the user.  the problem is that many users have done this already.
so we should start using a new name for it, like "default-notebook.bkr".

PR #3020, opened 2015-12-03
do not copy default.bkr to ~/.beaker #2994

kpaharelau:
default.bkr has been renamed to notebook-default.bkr
Default notebook name has been changed in the DefaulBeakerConfig class


scottdraves:
:+1:

PR #3021, opened 2015-12-03
Init indicator improvements (new branch)

alice-kroch:
Issue #3001. Please, check


scottdraves:
please do not move the menu to the left.

scottdraves:
it should look like this:
screen shot 2015-12-03 at 222

alice-kroch:
Probably it's not a good idea. If i leave menu margin 20px it's buttons will overlap power button
overlap
overlap2

So I adjusted menu margin value only for advanced mode.

scottdraves:
indeed but this is a smaller problem, please just do it.

scottdraves:
:+1:

Issue #3023, opened 2015-12-03
do not show day of week in timeseries axis labels


scottdraves:
![screen shot 2015-12-03 at 9 18 19 am](https://cloud.githubusercontent.com/assets/963093/11562905/d15d1132-999e-11e5-9e65-424a04932a9d.png)

do not show Thu Sat Mon, etc.

PR #3031, opened 2015-12-04
do not show day of week in timeseries axis labels #3023

kpaharelau:
Date format "YYYY MMM DD ddd, HH:mm" has been changed to the "YYYY MMM DD, HH:mm"
Date format "YYYY MMM DD ddd" has been changed to the "YYYY MMM DD"

### DIFF
--- a/core/src/main/web/outputdisplay/bko-plot/plotaxis.js
+++ b/core/src/main/web/outputdisplay/bko-plot/plotaxis.js
@@ -424,9 +424,9 @@
           ret = moment(d).tz(this.axisTimezone).format("HH:mm:ss");
         }
       } else if (plotUtils.lte(span, this.DAY)) {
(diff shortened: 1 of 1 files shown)


scottdraves:
i not not seeing day numbers at all now.  
screen shot 2015-12-07 at 12 54 12 pm

scottdraves:
note that i actually think that not showing the days at all is best for this graph, if you do that then the lines have to be on the 1st of the month.  these lines are on various days of the month as the interval is like 45 days.

since fixing that is more complex, right now please just remove the day-of-week string.

kpaharelau:
> > since fixing that is more complex, right now please just remove the day-of-week string. 

 it was a  solution for  this issue. 

According "i not not seeing day numbers at all now.". We a re using time format "YYYY MMM" for time spans less then year but greater then month.

scottdraves:
:+1:

Issue #2907, opened 2015-11-11
to fix scrollbar jumping, save output heights


scottdraves:
since the outputs instantiate lazily, as you scroll down a notebook for the first time, the scroll bar can jump as the height changes.  we should initialize them with empty divs with the right height (save it in the notebook model).

PR #3032, opened 2015-12-04
Fix scrollbar jumping, save output heights #2907 

maxyeg:


scottdraves:
this still isn't right because it's based on the same timeout approach even though it has a nicer name.
for example, in the big performance notebook the big image is loaded from an external web site and
is not yet available when the deferred code runs.  alice's 150ms delay might have been enough to get this image but not others.  and what if due to some random interaction a cell changes size?

there is no hurry in getting the heights of the cells.  the height we want to save in the document isn't the height of the cells when they are created, it's the height when they are saved.  we already have a hook in the code to scan the notebook when it's saved to reflect the html from the DOM into the notebook model.  this is the right time to get the heights.

scottdraves:

maxyeg:
Please, check.
1. Set height when notebook saved. No timeouts anymore
2. Updated branch.

core/src/main/web/app/mainapp/services/notebookcellmodelmanager.js
```diff
@@ -443,8 +443,9 @@
               if(!runningJob || runningJob.cellId !== cell.id){
                 cell.output.result = undefined;
                 cell.output.elapsedTime = undefined;
+                cell.output.height = undefined;
```
tomca32: why are we initializing to undefined? It's an object property, it will be undefined by default.

maxyeg: I'm not sure why. It was taken from previous commits. Thanks for mentioning.

scottdraves:
:+1:

Issue #3039, opened 2015-12-05
scroll on code cell create


scottdraves:
when you you insert a code cell, the notebook scrolls to place it near the window.
this is unexpected and annoying.

if the code is is created **offscreen** then it should scroll the minimum amount needed to put it on the screen.  but if it's OK then just leave it.

PR #3047, opened 2015-12-07
scroll on code cell create #3039

kpaharelau:
My proposal is to using standard codemirror focus mechanism instead hand-made function scrollIn. When we are creating codecell (or markdown cell) we are calling codemirror focus function. This function moves offscreen codemiiror in the center of window. This is not full expected behaviour for this issue but I have opinion that using standard tested mechanism is a better solution. What is your opinion?


tomca32:
I agree with you but I would like to know the reason why the scrollTo function was written in the first place. Maybe there was a good reason, or a fix for the bug we forgot. Do you remember the reasoning for it?

kpaharelau:
I think nobody knew about codemirror focus behaviour. Function scrollto always hides  this one
.

scottdraves:
:+1:

Issue #3038, opened 2015-12-05
opening easyform tutorial


scottdraves:
produces stderr

```
Client is trying to subscribe to nonexisting object ace9d52f-f694-460d-974c-cd33bb90fedf
lient is trying to subscribe to nonexisting object ace9d52f-f694-460d-974c-cd33bb90fedfClient is trying to subscribe to nonexisting object d178101b-1f54-4d22-b363-8f4a04193f62
lient is trying to subscribe to nonexisting object d178101b-1f54-4d22-b363-8f4a04193f62Client is trying to subscribe to nonexisting object d178101b-1f54-4d22-b363-8f4a04193f62
lient is trying to subscribe to nonexisting object d178101b-1f54-4d22-b363-8f4a04193f62Client is trying to subscribe to nonexisting object b726ad08-7102-4c81-a810-cc278de23319
```

scottdraves:
if form is instantiated without object, it should show an error state (red box?) and be nonfunctional until properly connected.

mnavasiolava:
reproduced both on mac and windows: you should just run the 1st cell of the tutorial and then scroll the page to the end.

scottdraves:
i see it sometimes and not others.  it's not reliable.  every time i see a bunch of

```
TypeError: value.forEach is not a function
    at writeNgOptionsMultiple [as writeValue] (beakerVendor.js:66548)
    at ngModelCtrl.$render (beakerVendor.js:68572)
    at Object.ngModelWatch (beakerVendor.js:65477)
    at Scope.$digest (beakerVendor.js:55872)
    at Scope.$apply (beakerVendor.js:56151)
    at beakerVendor.js:57918
    at completeOutstandingRequest (beakerVendor.js:45561)
    at beakerVendor.js:45838
```

errors and i have to think it's the same basic cause: the form GUI is trying to sync with an object that no longer exists.  if you look into easyform.bkr you can see

```
"update_id": "ace9d52f-f694-460d-974c-cd33bb90fedf"
...
"update_id": "d178101b-1f54-4d22-b363-8f4a04193f62"
```

this should probably not be saved into the file at all since it will never make sense later.

when the file is opened and you read it in, and instantiate the form, you should not bother trying to sync it back to the JVM.  in this state, with an update_id, the form should be displayed in a read-only state,
with all the widgets disabled.

mnavasiolava:
yes, you are right about "update_id" in .bkr

[...]

PR #3066, opened 2015-12-08
opening easyform tutorial #3038

mnavasiolava:
this PR contains the following changes:
1. model's "update_id" is not being saved to .bkr anymore
2. if model doesn't have "update_id" all components on the form will be disabled
3. fixes for combobox and date disable/enable


scottdraves:
:+1:

Issue #3061, opened 2015-12-08
timeseries labels and lines should be at beginning of interval


scottdraves:
![screen shot 2015-12-08 at 11 27 59 am](https://cloud.githubusercontent.com/assets/963093/11661258/f2d72786-9d9e-11e5-906a-3f778190cfae.png)
as you can see the line that is labeled "March" is actually March 12.
this is terribly broken.
lines should only be drawn at the beginning of their named interval.  this goes for days, months, and years.
maybe related: https://github.com/twosigma/beaker-notebook/issues/3052

Issue #3052, opened 2015-12-08
plot date range incorrect


scottdraves:
https://pub.beakernotebook.com/#/publications/566627d4-7558-49dd-9432-a561d5c05d6a

kpaharelau:
I did not find javascript in the https://pub.beakernotebook.com/#/publications/566627d4-7558-49dd-9432-a561d5c05d6a.  But after analysing I think problem doesn't exist
![image](https://cloud.githubusercontent.com/assets/13516511/11653448/607a792c-9daf-11e5-98d3-a0b1ded886fc.png)

See on screenshot.  Date range is correct. But on the sharing notebook '2007' must be 2004 Dec 31, 23:59:59 000.  Problem only in the date formatting. I changed format (only for a test) and we can see

![image](https://cloud.githubusercontent.com/assets/13516511/11653547/fd7f75c4-9daf-11e5-88b8-b40ab7887961.png)

PR #3067, opened 2015-12-09
timeseries labels and lines should be at beginning of interval #3061

kpaharelau:
I adopted algorithm for for days, months, and years. I choose the beginning of the next range (day, month or year), if the calculated value closer to the  end of the period, otherwise - end of the previous range. This is also fix for the https://github.com/twosigma/beaker-notebook/issues/3052

### DIFF
--- a/core/src/main/web/outputdisplay/bko-plot/plotaxis.js
+++ b/core/src/main/web/outputdisplay/bko-plot/plotaxis.js
@@ -236,21 +236,21 @@
 
       var self = this;
 
-      var selectStartOrEndRange = function(value, range) {
-        var endOfRange = moment(value).tz(self.axisTimezone).endOf(range).add(1, "ms");
-        var startOfRange = moment(value).tz(self.axisTimezone).startOf(range);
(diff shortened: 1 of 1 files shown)


core/src/main/web/outputdisplay/bko-plot/plotaxis.js
```diff
@@ -233,6 +233,31 @@
...
+        if (self.axisType === "time") {
+          if (plotUtils.gt(w, self.DAY)) {
+            if (plotUtils.lte(w, self.MONTH)) {
+              value = selectStartOrEndInterval(value, "day");
+            } else if (plotUtils.lte(w, self.YEAR)) {
+              value = selectStartOrEndInterval(value, "month");
+            } else {
+              value = selectStartOrEndInterval(value, "year");
+            }
+            console.log("line:" + moment(value).tz("UTC").format("YYYY MMM DD ddd, HH:mm:ss .SSS"));
```
tomca32: Is console.log supposed to be here?

kpaharelau: No :-) . I have removed it.

scottdraves:
:+1:

Issue #1976, opened 2015-07-23
Unexplained Bayeux Null


DiegoAlbertoTorres:
Leads to having to catch an NPE in SessionBackupRest.java, which spits out:
Warning: Caught NPE of unknown origin.

DiegoAlbertoTorres:
More information on origin of this NPE:
When using an electron client to switch to a pre-existing backend, all backend-Electron communications will fall unto this NPE, instead of properly posting on the cometd channel.

DiegoAlbertoTorres:
The problem mentioned in the previous comment was caused because Faye was not being rebound to the new backend server. Then it seems that the NPE occurs when a publication is attempted on a channel on which nobody is listening, in which case this NPE is properly handled and no error should be output.

maxyeg:
The only way I can reproduce it is:
In beaker(web version), when notebook is opened, click menu File -> Close.
It will call REST Api to close session:
https://github.com/twosigma/beaker-notebook/blob/master/core/src/main/java/com/twosigma/beaker/core/rest/SessionBackupRest.java#L190
and it tries to publish session update to Electron channel which was never created.
So I will separate api for closing electron session from web.

scottdraves:
still getting these.

maxyeg:
Could you describe test case?

scottdraves:
you said you have seen these now...

scottdraves:
these are visbile as:

```
Warning: Caught NPE of unknown origin.
```

PR #3075, opened 2015-12-09
Unexplained Bayeux Null #1976

mnavasiolava:
subscription to 'sessionChange' channel was only in controlpanel-directive.
if we open notebook and then reload the page re-handshake will be performed and "sessionChange" subscriber will be removed. when we perform session backup we are trying to find "sessionChange" channel to publish data but we can't because subscriber was removed. 
to avoid this i added subscription to "sessionChange" channel also to notebookmanager.js


scottdraves:
```
Warning: Caught NPE of unknown origin. frontend
```

I got this right away from opening the tutorial and clicking on easyform.

mnavasiolava:
can't see the error after steps you described @scottdraves . maybe you could provide more details to reproduce it.

scottdraves:
:+1: 
must have been a cache problem

Issue #3069, opened 2015-12-09
Error in browser console when open R tutorial


mnavasiolava:
after update from master get this error

```
TypeError: bkNotebookWidget.getFocusable(...).scrollTo is not a function
    at go (beakerApp.js:4281)
    at Object._impl.go2FirstErrorCodeCell (beakerApp.js:4293)
    at Object.bkHelper.go2FirstErrorCodeCell (beakerApp.js:2718)
    at :8801/8fdd768/beaker/plugin/menu/notebook.js:57
```

tomca32:
Confirmed. But I only get it when I run all cells after opening.

This seems like it was caused by https://github.com/twosigma/beaker-notebook/pull/3047 - @kpaharelau can you take a look?

tomca32:
This is the line: https://github.com/twosigma/beaker-notebook/blob/492429eef95a040db134ae9cfa7b251ff294b46b/core/src/main/web/app/mainapp/mainapp.js#L794

scottdraves:
i'm not seeing this.

mnavasiolava:
Tomislav was right, you should also run all cells to see the error

scottdraves:
i did!

mnavasiolava:
and still nothing?

tomca32:
try the groovy one. It happens when the cell errors for any reason

scottdraves:
OK I get it from the groovy error.
we do want to scroll to show an error when they happen.
we just have to use a method that exists...

PR #3079, opened 2015-12-10
Error in browser console when open R tutorial #3069

kpaharelau:
Function scrollTo has been restored


tomca32:
Does CodeMirror exposes its own scrollTo functionality?

kpaharelau:
I think that old scrollTo function is more preferable in this place. CodeMirror shows error cell in the center of window. Old function  - in the top part of window.

Issue #3044, opened 2015-12-06
Time series autotranslation bug


scottdraves:
https://pub.beakernotebook.com/#/publications/5664a5f8-80d8-4091-89f8-df63b16f3df0

maxyeg:
![pnd](https://cloud.githubusercontent.com/assets/2286032/11629178/a441d1e4-9d07-11e5-9f23-09115546f184.png)

Here is what I see.
What should be the result like?

scottdraves:
here's a clearer explanation:
https://pub.beakernotebook.com/#/publications/56666486-57fe-4411-9e55-4b1a1efd7f60

scottdraves:
i get different results depending on what order i run the cells in, sometimes Python3->IPython works, sometimes i get the [object object], for R somtimes I get an error and somtimes I get an object object.

maxyeg:
It still the same for me with the new notebook.[object object]

PR #3095, opened 2015-12-11
#3044. Time series autotranslation bug

maxyeg:
I found that autotranslation in JS ruins DataTable with complex(object) key.
Here: `o[v.values[r][0]] =` index is an object with type=Date and timestamp field. As it converted to string as `[object object]` all values will be mapped to this key and we can see:
[object object] 0.1234123123

Confirm if that's what we need, please.

### DIFF
--- a/core/src/main/web/app/mainapp/services/sessionmanager.js
+++ b/core/src/main/web/app/mainapp/services/sessionmanager.js
@@ -371,7 +371,11 @@
           if (v.subtype === "Dictionary") {
             var o = {}
             for (var r in v.values) {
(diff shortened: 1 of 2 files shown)


scottdraves:
this is a step in the right direction but more is needed to fix it for R.

scottdraves:
:+1:

maxyeg:
I've fixed this in R now. But as now values are put under integer key(timestamp) in javascript - it changes original value.
You can add python3 cell with code: `beaker.ts` and see that value in the first column differs from initial value.

scottdraves:
:+1:

Issue #3080, opened 2015-12-10
unlocking notebook breaks code cells


scottdraves:
1) open the ipython tutorial (or any notebook with code cells).
2) lock it
3) reload
4) unlock

![screen shot 2015-12-10 at 11 00 48 am](https://cloud.githubusercontent.com/assets/963093/11720474/6b12cc70-9f2d-11e5-8cc1-6c1122884ee9.png)

PR #3096, opened 2015-12-11
unlocking notebook breaks code cells #3080

kpaharelau:
$scope.$watch('isShowInput()' ... is not acceptable in this case because div with codemirror editor still hiding. I check presence 'ng-hide' class.

### DIFF
--- a/core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
+++ b/core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
@@ -70,21 +70,24 @@
           if ($scope.isLocked()) {
             return false;
           }
-          if ($scope.cellmodel.input.hidden === true) {
-            return false;
-          }
-          return true;
+          return $scope.cellmodel.input.hidden !== true;
         };
 
         $scope.bkNotebook = getBkNotebookWidget();
+
         // ensure cm refreshes when 'unhide'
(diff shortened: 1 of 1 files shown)


core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
```diff
@@ -70,21 +70,24 @@
...
         $scope.bkNotebook = getBkNotebookWidget();
+
         // ensure cm refreshes when 'unhide'
-        $scope.$watch('isShowInput()', function(newValue, oldValue) {
-          if ($scope.cm && newValue === true && newValue !== oldValue) {
-            bkUtils.fcall(function() {
-              $scope.cm.refresh();
-            });
+        var lastInputHidden = $(".code-cell-input").hasClass("ng-hide");
+        window.setInterval(function () {
```
tomca32, line 80: setInterval is outside of Angular scope. There is angular's $interval service that we should use for that https://docs.angularjs.org/api/ng/service/$interval

However, do we really want to poll every 10 miliseconds for every cell? So if we have 100 cells that's 100 intervals going off on every frame, 60 frames per second = 6000 function calls per second and every single one of them touches the DOM. That doesn't sound great.

Or am I misreading this code?

Issue #3078, opened 2015-12-09
scroll wheel stops working over charts


scottdraves:
if you scroll down a notebook, it stops when you get to a plot.
it seems the plot is eating the scroll events because plots use the wheel for zoom.
but the plot does not zoom until you click on it.  that's actually fine.
so if you don't click, it should continue scrolling, right?
this drives me nuts.

on laptop, the two-finger scroll gesture has a similar problem.

scottdraves:
still happens to treemaps

PR #3100, opened 2015-12-11
scroll wheel stops working over charts #3078

mnavasiolava:
until we click on the chart mouse the wheel will be used to scroll the page
when we click on the chart then mouse wheel will be for d3 zoom
but then if we leave the chart, the mouse wheel will once again be used to scroll the page, even if the mouse is over the chart (until we click on the chart once again)
please check if this is what we need


scottdraves:
:+1:

PR #3101, opened 2015-12-11
Revert "Upgraded CodeMirror 5.2.0 -> 5.9.0"

scottdraves:
Reverts twosigma/beaker-notebook#3090


scottdraves:
this merge broke highlighting in ambiance theme.  makes me think this is too risky for this particular time, need to test more and find a good reason to upgrade.

PR #3114, opened 2015-12-14
Cit/3080

kpaharelau:


scottdraves:
test failure is real, this breaks hiding input side of cell.

scottdraves:
still doesn't work for me if the locked notebook has hidden code, eg the first big html cell of the main tut.
start with that notebook (with hidden html input), lock it, reload, unlock, open the input I get:
screen shot 2015-12-14 at 11 19 48 pm

Issue #3088, opened 2015-12-10
make logo color a parameter


scottdraves:
we need to adjust the color of the logo for themes.
right now the color white is baked in to the SVG.
it should be a parameter, and in the ambiance theme it should be our standard color for text #E6E1DC
in the regular theme, it should be #FFFFFF.
please make this fix without duplicating the SVG.

PR #3115, opened 2015-12-14
Make logo color a parameter

alice-kroch:
Issue #3088. Please, check.

### DIFF
--- a/core/src/main/web/app/helpers/core.js
+++ b/core/src/main/web/app/helpers/core.js
@@ -294,6 +294,7 @@
           preferencename: 'theme',
           preferencevalue: theme
         });
+        $rootScope.$broadcast('beaker.theme.set', theme);
       },
       getTheme: function () {
         return this._prefs.getTheme();
@@ -1097,6 +1098,7 @@
       preference: 'theme'
     }).success(function (theme) {
       bkCoreManager._prefs.setTheme(_.contains(_.values(GLOBALS.THEMES), theme) ? theme : GLOBALS.THEMES.DEFAULT);
+      $rootScope.$broadcast('beaker.theme.set', theme);
     }).error(function (response) {
       console.log(response);
       bkCoreManager._prefs.setTheme(GLOBALS.THEMES.DEFAULT);
--- a/core/src/main/web/app/utils/basic/commonui.js
+++ b/core/src/main/web/app/utils/basic/commonui.js
@@ -337,4 +337,30 @@
     };
   });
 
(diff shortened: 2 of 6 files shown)


core/src/main/web/app/helpers/core.js
```diff
@@ -294,6 +294,7 @@
           preferencename: 'theme',
           preferencevalue: theme
         });
+        $rootScope.$broadcast('beaker.theme.set', theme);
```
tomca32, line 297: Use $rootScope.$emit instead of $broadcast. $emit will only trigger the listener on the $rootScope, while $broadcast will go through all of the child scopes. Since the listener is on rootScope, $emit is the correct option.

http://toddmotto.com/all-about-angulars-emit-broadcast-on-publish-subscribing/

core/src/main/web/app/helpers/core.js
```diff
@@ -1097,6 +1098,7 @@
       preference: 'theme'
     }).success(function (theme) {
       bkCoreManager._prefs.setTheme(_.contains(_.values(GLOBALS.THEMES), theme) ? theme : GLOBALS.THEMES.DEFAULT);
+      $rootScope.$broadcast('beaker.theme.set', theme);
```
tomca32, line 1101: Here as well

core/src/main/web/app/utils/basic/commonui.js
```diff
@@ -337,4 +337,30 @@
...
+    return {
+      restrict: 'A',
+      scope: {
+        bkBrandFill: '='
+      },
+      link: function(scope) {
+        var brandColors = {
+          'default': '#FFFFFF',
+          'ambiance': '#E6E1DC'
+        };
```
tomca32: Why can't we put this with the rest of the ambiance theme scss?

tomca32:
Why is the logo a special case that can't be handled by the already existing theme code?

scottdraves:
it's special because "white" was hardcoded in the SVG.

scottdraves:
now that I see this, probably also a good time to factor the SVG so there is only one copy of it (in that new directive?).

tomca32:
But why can't we just put a class on it the way we handle all other theme elements, and put the styling in ambience.scss? SVG fill is just a css property.

scottdraves:
yes it should use a css class.

scottdraves:
cool.  still needs to refactor to avoid duplicating the svg.

scottdraves:
too complicated for right now

Issue #2191, opened 2015-08-25
in log plot, do not show lines below 0


scottdraves:
<img width="254" alt="screen shot 2015-08-24 at 11 37 46 pm" src="https://cloud.githubusercontent.com/assets/963093/9458684/587b601c-4abf-11e5-90f9-eb5a7ff67e47.png">

PR #3119, opened 2015-12-14
in log plot, do not show lines below 0 #2191

mnavasiolava:
lines below 0 are not shown both for x and y axes


scottdraves:
:+1:

PR #3127, opened 2015-12-15
unlocking notebook breaks code cells #3080

kpaharelau:
Functionality for the refresh codemirror after unhide cell has been added. In current momemt we are using 2 way for refreshing codemirror:
1. After changing 'IsLocked' property
   2/ After click on "Unhide" menu button

### DIFF
--- a/core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
+++ b/core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
@@ -18,6 +18,26 @@
   'use strict';
   var module = angular.module('bk.notebook');
 
+  module.directive('ngHideEx', function($animate) {
+    return {
+      scope: {
+        'ngHideEx': '=',
+        'afterShow': '&',
+        'afterHide': '&'
+      },
+      link: function(scope, element) {
+        scope.$watch('ngHideEx', function(hide, oldHide) {
+          if (hide) {
+            $animate.addClass(element, 'ng-hide', {tempClasses: 'ng-hide-animate'}).then(scope.afterHide);
+          }
+          if (!hide) {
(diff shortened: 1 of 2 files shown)


core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
```diff
@@ -25,7 +45,8 @@
       bkSessionManager,
       bkCoreManager,
       bkPublicationHelper,
-      $timeout) {
+      $timeout,
+      $animate) {
```
tomca32: why do we need animate here? It seems that it's used only in ngHideEx directive.

core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
```diff
@@ -70,22 +91,24 @@
...
+
+        //ensure cm refreshes when 'unhide'
+        $scope.afterShow = function () {
+          $scope.cm.refresh();
+        };
+        $scope.$watch('cellmodel.input.hidden', function(newValue, oldValue) {
+          if ($scope.cm && oldValue === true && newValue !== oldValue) {
             bkUtils.fcall(function() {
-              $scope.cm.refresh();
+              $scope.afterShow();
```
tomca32, line 105: Do we need the call to afterShow here if the ngHideEx directive handles that automatically?

kpaharelau, line 105: Yes. ngHideEx is not working for hidden cells

core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
kpaharelau: good point - will be deleted

scottdraves:
:+1:

Issue #3139, opened 2015-12-15
Cannot read property 'refresh' of undefined


tomca32:
Seems to happen for every cell in notebook.

Steps:
1. Open dev tools console
2. Create a new notebook
3. Add any cell

![screen shot 2015-12-15 at 2 45 11 pm](https://cloud.githubusercontent.com/assets/3384784/11822027/7cad13ee-a33a-11e5-8249-172c6e97628a.png)

@kpaharelau can you take a look? I'm pretty sure this has to do with: https://github.com/twosigma/beaker-notebook/pull/3127

PR #3141, opened 2015-12-16
Cannot read property 'refresh' of undefined #3139

kpaharelau:
Check that codemirror is defined  has been added


scottdraves:
:+1:

PR #3145, opened 2015-12-16
beaker object labels style change

alice-kroch:
Please, check.

### DIFF
--- a/core/src/main/web/app/styles/cell-outputs.scss
+++ b/core/src/main/web/app/styles/cell-outputs.scss
@@ -18,6 +18,10 @@ bk-code-cell-output {
   .ui-state-default {
     background: rgb(235, 235, 235);
   }
+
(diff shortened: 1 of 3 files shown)


core/src/main/web/app/styles/cell-outputs.scss
```diff
@@ -18,6 +18,10 @@ bk-code-cell-output {
   .ui-state-default {
     background: rgb(235, 235, 235);
   }
+
+  b {
```
scottdraves: https://pub.beakernotebook.com/#/publications/567171b3-fce4-4164-a2ad-536ca126d3cc

don't change all bold text in the output, just the beaker object labels.  please add a new class to the output of this part of the beaker object and then define that class to be bold&stdout.

scottdraves:
:+1:

Issue #3083, opened 2015-12-10
add small run button to sections


scottdraves:
just to the left of the trashcan, add a small run button like on code cells.
for sections, this would "run all cells" (put that in the tooltip), just like the menu item.

![screen shot 2015-12-10 at 11 42 11 am](https://cloud.githubusercontent.com/assets/963093/11721695/184560ba-9f33-11e5-8d46-a82baf8c74da.png)

alice-kroch:
and by the way - why do we have 'run cell' button in markdown cells? it even has no click action

scottdraves:
running a markdown cell just leaves edit mode and formats it for display.
there are other ways to do this but it is more consistent to have a run button.

PR #3147, opened 2015-12-16
Add small run button to sections

alice-kroch:
Issue #3083. Please, check

### DIFF
--- a/core/src/main/web/app/mainapp/components/notebook/cell-directive.js
+++ b/core/src/main/web/app/mainapp/components/notebook/cell-directive.js
@@ -128,11 +128,17 @@
           if ($event) {
             $event.stopPropagation();
           }
+          var toEval;
 
-          $scope.cellmodel.output.state = {};
+          if ($scope.cellmodel.type === 'section') {
+            toEval = $scope.cellmodel.id;
+          } else {
+            $scope.cellmodel.output.state = {};
+            toEval = $scope.cellmodel;
+          }
 
           bkCoreManager.getBkApp()
-            .evaluateRoot($scope.cellmodel)
+            .evaluateRoot(toEval)
             .catch(function(data) {
               console.error(data);
             });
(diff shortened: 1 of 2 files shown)


core/src/main/web/app/mainapp/components/notebook/cell-directive.js
```diff
@@ -217,6 +217,30 @@
         $scope.isCodeCell = function() {
           return $scope.cellmodel.type == 'code';
         };
+
+        $scope.isSection = function() {
```
tomca32: I don't think we need this function. You can just ng-if="cellmodel.type === 'section'"

Watching functions greatly reduces the performance of the watcher.

core/src/main/web/app/mainapp/components/notebook/cell.jst.html
```diff
@@ -45,6 +45,7 @@
     </div>
     <div class="cell-menu-item evaluate" ng-if="isMarkdownCell()" title="run cell"></div>
     <div class="cell-menu-item evaluate" ng-click="evaluate($event)" ng-if="isCodeCell()" title="run cell"></div>
+    <div class="cell-menu-item evaluate" ng-if="isSection()" ng-click="evaluateSection($event)" title="run all cells"></div>
```
tomca32: As I said above you can ng-if="cellmodel.type === 'section'" - no need for the isSection function, we already have the answer and it won't change, we don't need to compute it all the time.

Also, since the type of a section won't change, we don't need to watch it all the time, so we can use one-time binding, like this:

ng-if="::cellmodel.type === 'section'" - the double colon at the start of the expression marks it as one time evaluation. The watcher will run once when the section is created and never again. This helps us reduce the number of watchers that fire on every digest.

core/src/main/web/app/mainapp/components/notebook/cell-directive.js
alice-kroch: OK. And how about other similar functions?

tomca32: Yeah, you're right,  we are doing that a lot. It's not great and one day we should change it but at least we can stop adding new ones.

tomca32:
Looks much better. Thank you

core/src/main/web/app/mainapp/components/notebook/cell-directive.js
```diff
@@ -217,6 +217,26 @@
         $scope.isCodeCell = function() {
           return $scope.cellmodel.type == 'code';
         };
+
+        $scope.evaluateSection = function($evt) {
```
scottdraves: we already have this feature, no need to reimplement
https://github.com/twosigma/beaker-notebook/blob/a811e079ac5afc8ac107ea6dffc3c565509f00d5/core/src/main/web/app/mainapp/components/notebook/sectioncell-directive.js#L130

scottdraves:
:+1:

Issue #3065, opened 2015-12-08
R dataframe autotranslation breaks on control-character in string


jsbachsonata:
Hi Scott,

Now that cell A is runnign fine, I tried to set my all_comments dataframe as a beaker object "test" in cell B but the time elapsed counter keeps on going indefinitely when I run cell B. Is there something wrong with my autotransalation code or is this a bug?

cell A:
all_comments<-read.csv("\all_comments_20151208.csv")
0

cell B:
beaker::set('test',all_comments)
0

jsbachsonata:
keep me posted on the timeline for the fix. I'm looking to give a presentation on Beaker with my analytics group at Boston Consulting Group since we often go between R, Python, and Spark.It'll def be a useful tool for us once we can create R beaker objects reliably :)

scottdraves:
thanks, i am pretty sure this is a dup of https://github.com/twosigma/beaker-notebook/issues/2799
please watch it there.

scottdraves:
actually there's an additional problem here.  it is really really slow, which is the bug mentioned.
but there is also another bug, because after about 15 minutes the autotranslation of this dataframe does finish, but with an error.  it says

```
Error
 Error : Server Error
```

in the cell and I get the following stack trace from java:

```
org.codehaus.jackson.JsonParseException: Illegal unquoted character ((CTRL-CHAR, code 3)): has to be escaped using backslash to be included in string value
 at [Source: java.io.StringReader@6ab4fc74; line: 1, column: 2877113]
```

scottdraves:
which you can easily reproduce with

```
beaker::set('test','\003')
```

PR #3156, opened 2015-12-17
#3065. R dataframe autotranslation breaks on control-character in string

maxyeg:
Code for containsControlCharacters and escapeControlCharacters is duplicated.
The reason is that as I see core module(NamespaceRest) and jvm/shared module(SimpleEvaluationObject) are independent.
Maybe some proposals here?

### DIFF
--- a/plugin/jvm/src/main/java/com/twosigma/beaker/jvm/object/SimpleEvaluationObject.java
+++ b/plugin/jvm/src/main/java/com/twosigma/beaker/jvm/object/SimpleEvaluationObject.java
@@ -30,7 +30,10 @@ import java.util.List;
 import java.util.Observable;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.regex.Pattern;
 
+import org.apache.commons.lang3.StringEscapeUtils;
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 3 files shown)


plugin/jvm/src/main/java/com/twosigma/beaker/jvm/object/SimpleEvaluationObject.java
```diff
@@ -190,6 +193,33 @@ public synchronized int getProgressBar() {
 
     private final Provider<UpdateManager> updateManagerProvider;
     private final Provider<BeakerObjectConverter> objectSerializerProvider;
+    private static final Pattern CONTROL_CHARACTERS_IN_STRING_PATTERN
+        = Pattern.compile("[^\\p{Cc}]*[\\p{Cc}]+[^\\p{Cc}]*(?U)");
```
scottdraves: why use this complicated pattern instead of just calling isISOControl in a loop?

scottdraves:
we have a place for code shared between core and plugins:
https://github.com/twosigma/beaker-notebook/tree/master/shared/src/main/java/com/twosigma/beaker/shared

scottdraves:
:+1:

Issue #3154, opened 2015-12-16
new Text plotitem properties


scottdraves:
1) please add color: `new Text(... color: ...)`
2) when a Text item is drawn, it should actually have a little arrow that points at its location.
the direction of the line can be set with the `pointerAngle` property.  0 means straight left, Pi/2 means straight up, etc.
3) showPointer: true/false (turn it off, default is on)

PR #3162, opened 2015-12-18
new Text plotitem properties #3154

kpaharelau:
Please test using 
https://www.dropbox.com/s/i830o5j3on9zyaf/3154.bkr?dl=0

I used new svg element 'marker' for arrow drawing

### DIFF
--- a/core/config/tutorials/chartingTutorial.bkr
+++ b/core/config/tutorials/chartingTutorial.bkr
@@ -40,7 +40,7 @@
                     "background": "#E0FFE0"
                 }
             },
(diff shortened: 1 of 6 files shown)


scottdraves:
i like the use of marker.  but when `showPointer: false` i still see the arrows.

scottdraves:
text needs size property too (see the pdf i sent you).

scottdraves:
color is for the text itself, not for the pointer.

scottdraves:
the alignment of the pointer and the text is off.  please compare to results in the PDF.
i think you are using the lower-left corner of the text, but it should be the center.

scottdraves:
:+1:

Issue #3053, opened 2015-12-08
make plotly work in R


scottdraves:
https://pub.beakernotebook.com/#/publications/56663c28-1278-4b24-9bd3-34dc89d16a70
should be very similar to ggvis https://github.com/twosigma/beaker-notebook/issues/2301

PR #3163, opened 2015-12-18
make plotly work in R #3053

mnavasiolava:
https://pub.beakernotebook.com/#/publications/5674305d-bed2-4438-9c7f-9892a60e1638


plugin/r/src/dist/beaker/R/beaker.r
```diff
@@ -14,6 +14,7 @@
 
 library(RCurl, quietly=TRUE)
 library(RJSONIO, quietly=TRUE)
+library(plotly, quietly=TRUE)
```
scottdraves: don't load it by default, not all users want it.

scottdraves:
load plotly from bower.  i am pretty sure you can get htmlwidgets from bower too by specifying the git repo.

scottdraves:
i get a console error too:

```
htmlwidgets.js:555: Uncaught TypeError: Cannot read property 'render' of undefined  
```

scottdraves:
ideally we should get the JS from R and not have a separate copy in Beaker.

scottdraves:
:+1:

Issue #3022, opened 2015-12-03
save chart tooltips


scottdraves:
if any tooltips are open, their positions should be saved, restored, and published.
also, after you click to open a tooltip, it should get a little x in the upper-right to close it.
also, draw a marker from the plotitem to the nearest corner or edge midpoint of the tooltip.
there are 8 possible places for the marker line to be anchored on the tooltip, 4 corners, and 4 midpoints.
when the tooltip is dragged, the marker should stay attached, and move to new anchor as needed.

PR #3175, opened 2015-12-21
Chart tooltips improvement

alice-kroch:
Issue #3022.


scottdraves:
1) please make the closing "x" only appear if the tooltip has been pinned, if it is hovering then omit it.
2) let's skip the line for now, i don't like this curved one, doing this right is probably too complicated for 1.5

core/src/main/web/outputdisplay/bko-plot/bko-plot.js
```diff
@@ -482,6 +483,23 @@
           });
         };
 
+        scope.pinCloseIcon = function(d) {
+          var tipid = "tip_" + d.id;
```
scottdraves: looks like this code was copied.  instead, make a function that you can call from each place.

scottdraves:
thanks that's better but still, this code has been duplicated:

```
              delete scope.tips[d.id];
              if (d.isresp === true) {  // is interaction responsive element
                scope.jqsvg.find("#" + d.id).css("opacity", 0);
              } else {
                scope.jqsvg.find("#" + d.id).removeAttr("filter");
              }
              scope.interactMode = "remove";
```

scottdraves:
xlnt, save/restore works.
now just needs:
1) address the duplicated code mentioned above.
2) give a shot at drawing the line, it's similar to the one for text plotitems.
3) bring up to date with master.

kpaharelau:
2) give a shot at drawing the line, it's similar to the one for text plotitems.
FIXED - try to use

scottdraves:
cool, this looks basically right. to improve it:
1) make it have 8 possible attachments, not just 4 (include the middles of the edges, not just the corners).
2) the marker should never appear when the tooltip is visible due to hover, only if the user clicks and pins.
3) target of the marker should be in the coordinates of the chart so it moves with the rest of the chart when you drag and zoom.

scottdraves:
cool, almost done!  still two small problems:
1) the color of the marker should be the same as the outline of the tooltip (this will make it work in themes)
screen shot 2015-12-30 at 10 07 46 am

2) the middle-edge attachment should be chosen more often.  ideally it should choose the attachment to minimize the length of the marker. 
screen shot 2015-12-30 at 10 00 16 am

Issue #3126, opened 2015-12-15
section headings editing is off


scottdraves:
in particular if i double click on a word it cycles through 3 results:
just leaving the caret, selected in dark blue, and selected in light blue (here are the pics in reverse order, last to first):
<img width="169" alt="screen shot 2015-12-14 at 10 13 01 pm" src="https://cloud.githubusercontent.com/assets/963093/11801077/495c29c4-a2b0-11e5-84ad-7c5394473c79.png">
<img width="196" alt="screen shot 2015-12-14 at 10 12 54 pm" src="https://cloud.githubusercontent.com/assets/963093/11801078/495fa7e8-a2b0-11e5-8e15-90d3279c3604.png">
<img width="172" alt="screen shot 2015-12-14 at 10 12 47 pm" src="https://cloud.githubusercontent.com/assets/963093/11801079/49604e32-a2b0-11e5-9902-0fe3be07dc9d.png">
<img width="146" alt="screen shot 2015-12-14 at 10 12 39 pm" src="https://cloud.githubusercontent.com/assets/963093/11801080/4965e04a-a2b0-11e5-8bad-9a32c4881deb.png">
it should just go to the dark blue, and blink with each double-click.

scottdraves:
i believe this is a regression, try backing up.

scottdraves:
seems like an old bug...

scottdraves:
thanks maxy i understand this better now, so part of the behavior is explained (the first double click leaving the caret).  I get the two kinds of selection too.

what's still unexplained is why the 3rd double click does an html selection and not a codemirror selection.  after the 2nd double-clicks, the heading is in edit code.  so when you continue to click on it, why does it leave edit mode and allow an html selection?

scottdraves:
If i have a CM selection in a section header (from double click or from dragging) and then I click again I get two different behaviors: if i click _on_ the selected text it leaves edit mode.  if i click on other text in the heading, then i get a caret.  this seems like a bug, i should always get a caret.

maxyeg:
When I select text in CM(in edit mode, any CM, in code cell also) and after that I click inside this selection, CM fires `onblur` event.
Our `markdown-editable-directive` handles CM `blur` event and enters preview mode(caret lost).

I will still examine why this happens, but I don't think it's right to fire blur event in that case, because no matter of if I click inside or near selected text, I'm still in CM, so it's focused, not blurred.

scottdraves:
sounds like an important clue.

[...]

PR #3179, opened 2015-12-22
#3126. section headings editing is off

maxyeg:
This is the mentioned workaround.
Please check if it's ok.
The thing that is fixed by this: click inside CM selection doesn't switch CM into preview mode anymore.


scottdraves:
:+1:

Issue #3055, opened 2015-12-08
add treemap demo


scottdraves:
based on this dataset, using our treemap class
http://bl.ocks.org/mbostock/4063582

PR #3186, opened 2015-12-23
add treemap demo #3055

kpaharelau:
Look at the end of the tutorial. Also I fixed error with resizing treemap/ Problem was in the new version of the lodash


scottdraves:
colors are missing.  you can copy the array from here:
https://github.com/mbostock/d3/wiki/Ordinal-Scales

scottdraves:
OK looks good, but tests fail.

scottdraves:
:+1:

Issue #3196, opened 2015-12-24
ambiance bugs


scottdraves:
table H and V scrollbars should have dark gray bkg, like the codemirror scrollbar:
![screen shot 2015-12-24 at 12 21 15 pm](https://cloud.githubusercontent.com/assets/963093/11996816/825b2f82-aa39-11e5-833b-4d91dd352531.png)
![screen shot 2015-12-24 at 12 23 43 pm](https://cloud.githubusercontent.com/assets/963093/11996818/853389e8-aa39-11e5-92c3-82275de7514f.png)

hrule in the view menu should be cell border color:
![screen shot 2015-12-24 at 12 21 57 pm](https://cloud.githubusercontent.com/assets/963093/11996823/94e8c5b0-aa39-11e5-9651-bf82a61000c4.png)

the rectangle behind the show/hide button and output menu button should be transparent, same as in normal theme:
![screen shot 2015-12-24 at 12 21 49 pm](https://cloud.githubusercontent.com/assets/963093/11996853/5a812ba0-aa3a-11e5-869d-cc98d0125413.png)

PR #3198, opened 2015-12-29
Ambiance theme bugs fixes

alice-kroch:
Issue #3196. Please check. (Note: no scrollbars stylization here. Will be referenced in another issue)


scottdraves:
the menu line is good, but the rectangle in the cell output menu area is too transparent.  it should be like the default theme:
screen shot 2015-12-29 at 10 01 17 am
this is what i am seeing now:
screen shot 2015-12-29 at 10 03 26 am

alice-kroch:
Done. Please, check

Issue #3183, opened 2015-12-23
ambiance bugs


scottdraves:
box around legend should be same as cell border color, not bbbbbb.
<img width="251" alt="screen shot 2015-12-23 at 12 35 29 am" src="https://cloud.githubusercontent.com/assets/963093/11971528/7b588568-a90d-11e5-800b-2630a8d0803d.png">

scottdraves:
<img width="251" alt="screen shot 2015-12-23 at 12 35 29 am" src="https://cloud.githubusercontent.com/assets/963093/11971528/7b588568-a90d-11e5-800b-2630a8d0803d.png">

scottdraves:
and easyform scrollbars should have the same dark bkg as the horizontal codemirror scrollbar.
<img width="157" alt="screen shot 2015-12-23 at 12 58 21 am" src="https://cloud.githubusercontent.com/assets/963093/11971776/66ff09ea-a910-11e5-86dc-31e1ec2f4477.png">

and easyform checkboxes and radio buttons should have the dark bkg (until they are selected and then they go blue, keep that).

<img width="213" alt="screen shot 2015-12-23 at 12 59 36 am" src="https://cloud.githubusercontent.com/assets/963093/11971787/930dbdec-a910-11e5-9c4a-1c494cb0e0c2.png">

scottdraves:
python widgets need to be themed.
and the text came out black here, why?
<img width="562" alt="screen shot 2015-12-23 at 1 08 52 am" src="https://cloud.githubusercontent.com/assets/963093/11971874/ff0b7312-a911-11e5-8f8e-c1e9edce3c36.png">

scottdraves:
progress reporting bar bkg should be dark gray.
<img width="436" alt="screen shot 2015-12-23 at 1 13 03 am" src="https://cloud.githubusercontent.com/assets/963093/11971898/64b70672-a912-11e5-94fe-8cbf12bfc68a.png">

scottdraves:
tab char should be med-light gray or normally visible.
<img width="172" alt="screen shot 2015-12-23 at 1 15 06 am" src="https://cloud.githubusercontent.com/assets/963093/11971933/c62f9806-a912-11e5-8285-ccdcd770b335.png">

scottdraves:
in OutputContainerCell, GridOutputContainerLayoutManager, lines should be cell border gray.
[...]

PR #3199, opened 2015-12-29
The "offline" indicator is missing from the control panel

alice-kroch:
Issue #855. Please, check. (Note: indicator color in ambiance theme is fixed in issue #3183)


scottdraves:
works, but indicator should be on the right, same as on the notebook page.

alice-kroch:
Fixed. Please, check.

scottdraves:
:+1:

PR #3202, opened 2015-12-29
Spot/torch

scottdraves:
add torch/lua language including basic autotranslation.


ildipo:
it looks good but you should rebase

Issue #3173, opened 2015-12-20
plot colors should depend on theme


scottdraves:
1) the blue, red, and yellow are not right for the dark theme.
the default plot colors (blue red etc) should be part of the theme.
ideally delay selection of RGB values to css or js time.

2) the yellow comes from user code as RGBA value, so no natural way to fix automatically.
therefor should expose the current theme via beaker object API so groovy code can adapt.  so "beaker.theme.name" would evaluate to "default" or "ambiance".

3) groovy API access to the 6 default theme plot colors, so I can pass them into colors of other plot items, and they are displayed correctly to user. "beaker.theme.plotColor" would be an array of 6 colors.

scottdraves:
![screen shot 2015-12-20 at 1 39 54 pm](https://cloud.githubusercontent.com/assets/963093/11919401/c4fa1fb4-a71f-11e5-8090-8a501f8c5951.png)
![screen shot 2015-12-20 at 1 39 20 pm](https://cloud.githubusercontent.com/assets/963093/11919403/cd55dfea-a71f-11e5-9850-79b76591d42d.png)

PR #3204, opened 2015-12-29
plot colors should depend on theme #3173

mnavasiolava:
default plot colors depend on app theme.
theme object ({name: "", plotColors:[]}) was exposed to beaker object


scottdraves:
this is good but is it possible to do better?
the problem is if you change the theme, you have to rerun the cells to get the colors to match.
in particular, on the publication server, running cells is not possible....

mnavasiolava:
if color is not set to plot item (used one of the default theme colors) you shouldn't rerun the cell - new color will be applied automatically to plot after you change the theme
if color is set via beaker object i'm not sure that there is a way to update color without cell rerun, because plot model just contains color and it doesn't know that this color was taken from beaker object

scottdraves:
i thought so... but then i did a careful check of the legend and it looks like just that part is not updated.
the main timeseries are.  i am using the SimpleTimePlot example from the beginning of the interactive charts tutorial.  in default theme, i render, take a snapshot of the legend, then switch to ambiance and then take another snapshot, the color in the legend is unchanged.  the lines did get brighter though:
screen shot 2015-12-29 at 7 35 07 pm
screen shot 2015-12-29 at 7 35 14 pm
screen shot 2015-12-29 at 7 35 37 pm

mnavasiolava:
you're right, legend's colors are cached in state, i'll fix them

scottdraves:
:+1:

PR #3212, opened 2015-12-30
Ambiance bugs

alice-kroch:
Issue #3183.


scottdraves:
ok great, this looks almost all good, except the easyform checkboxes and radio buttons when they are selected are too dark.
screen shot 2015-12-30 at 4 10 09 pm
can you use the brighter blue and std-fg-text checkmark, more like they were?
screen shot 2015-12-30 at 4 17 00 pm

also, looks like the layout of Radio Buttons vs Radio Buttons H got exchanged.

PR #3213, opened 2015-12-31
hack big.js to load without modules (for native beaker)

mnavasiolava:
for electron version Big is not set as global variable.

```
   // Node and other CommonJS-like environments that support module.exports.
   if ( typeof module !== 'undefined' && module.exports ) {
       module.exports = Big

   //AMD.
   } else if ( typeof define == 'function' && define.amd ) {
       define( function () {
           return Big
       })

   //Browser.
   } else {
       global['Big'] = Big
   }
```

that's how export in big.js looks like
for electron the 1st check works instead of the last one

### DIFF
--- a/core/src/main/web/vendor/bower_components/big.js/big.js
+++ b/core/src/main/web/vendor/bower_components/big.js/big.js
@@ -1008,19 +1008,21 @@
     // EXPORT
 
 
-    // Node and other CommonJS-like environments that support module.exports.
-    if ( typeof module !== 'undefined' && module.exports ) {
-        module.exports = Big
-
-    //AMD.
-    } else if ( typeof define == 'function' && define.amd ) {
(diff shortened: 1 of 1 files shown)


scottdraves:
why is this a problem for the electron version but not for the regular version?

mnavasiolava:
in regular beaker version global “module” var is undefined, while electron defines “module” var and big js tries to use model.export instead of global var
image

tomca32:
Makes sense. Thanks @mnavasiolava.

I still hate that we are doing this. This is the second library which needs this hack (first one was D3) so perhaps we should think about solving this problem since it will happen a third time.

Maybe having a build step for electron load dependencies as modules somehow? Or embracing modules for the rest of our JS code?

scottdraves:
:+1:

Issue #3024, opened 2015-12-03
plot tooltip not cropped


scottdraves:
open a tooltip then pan the chart, it can leave the output area.
![screen shot 2015-12-03 at 9 21 21 am](https://cloud.githubusercontent.com/assets/963093/11562986/4fb41fe4-999f-11e5-89da-a72750597da5.png)

PR #3235, opened 2016-01-04
plot tooltip not cropped #3024

kpaharelau:
outsideGrid function for checking position tooltips in the grid area has been added

### DIFF
--- a/core/src/main/web/outputdisplay/bko-plot/plottip.js
+++ b/core/src/main/web/outputdisplay/bko-plot/plottip.js
@@ -97,6 +97,14 @@
     };
 
 
+    /**
+     * This code checks that tip is in the grid area
+     * @param x - x coordinate of the tip
+     * @param y - y coordinate of the tip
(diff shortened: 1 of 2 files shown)


core/src/main/web/outputdisplay/bko-plot/plottip.js
```diff
@@ -96,6 +96,17 @@
...
 
+
+    var outsideGrid = function (scope, x, y, w, h) {
+      var xPadding = 10;
+      var bBox = scope.jqgridg.get(0).getBBox();
+      var W = bBox.width;
+      var H = bBox.height;
+      var X = bBox.x;
+      var Y = bBox.y;
+      return x > W + X - xPadding || x + w - X + xPadding < 0 || y > H + Y || y + h - Y < 0;
```
tomca32, line 115: Could maybe use a comment to explain what are we returning since it's tricky to read this line.

scottdraves:
:+1:

PR #3237, opened 2016-01-04
scroll wheel stops working over charts #3078

kpaharelau:
Handlers for  events "focusin" and "focusout" has been added


core/src/main/web/outputdisplay/bko-plot/plotitems/std/plottreemapnode.js
```diff
@@ -110,19 +110,64 @@
...
+			};
+
+			// set zoom object
+			scope.maing.on("focusin", function() {
+				enableZoom();
+			}).on("focusout", function() {
+				disableZoom();
 			});
 
+      scope.maing.call(zoom)
```
scottdraves, line 163: indentation

Issue #3250, opened 2016-01-05
Tooltip lines not drawn for  some types of plotitems.


kpaharelau:
Tooltip lines not drawn for  some types of plotitems:
1. Bar
2. Steam 
3. Heatmap

and check other

kpaharelau:
Fix within https://github.com/twosigma/beaker-notebook/issues/3025

Issue #3025, opened 2015-12-03
plot hover circle inconsistent


scottdraves:
when you hover over a point in the plot, the tooltip appears, and a little circle appears around the point.
as you move the mouse around the point though, the circle appears and vanishes, but the tooltip stays up.
seems like the zone for the circle is smaller than the zone for the tip.
they should be the same and the one for the tip is the right one.

scottdraves:
also, if you click to bring up to tooltip, if you click when the circle is not visible then you don't get the line connecting the tip to the plotitem.

Issue #3243, opened 2016-01-04
tooltips vanish on scroll


scottdraves:
if you pan or zoom a chart and a tooltip is cropped, then it vanishes and does not come back when the viewport is restored.  the same problem exists if you use the legend to toggle the visibility of a line.

tips should be hidden like this, but they should come back.

PR #3256, opened 2016-01-06
plot hover circle inconsistent #3025

kpaharelau:
1. Removing drawing circle for line. Do the bold line, like when you hover over the legend
2. Change algorithm for getting  coordinates of  target point for the pinned tooltip line
   a. Use center for bars and heatmap
   b. use mouse position for a steam
   This is bug is correlated to https://github.com/twosigma/beaker-notebook/issues/3250.
3. Filter svgfilter did not work - fixed

Use      https://www.dropbox.com/s/yltwp4rk6nrclwa/3025.bkr?dl=0
for testing new behaviour


scottdraves:
OK this is a step in the right direction, i don't get tips without lines anymore :)

but there are some problems.
when you highlight an item on mouseover, it should just highlight the individual point or or line segment, not the whole series.
i am not seeing points getting highlighted.
the highlight does not seem as fat as when i use the legend.

also, it seems inconsistent to sometimes use the clicked location like for stems, and sometimes use the center.  i think always using the actual clicked location is better.

kpaharelau:
example why using the clicked location is not good solution by my opinion

image

you can use https://www.dropbox.com/s/yltwp4rk6nrclwa/3025.bkr?dl=0
for the tests

kpaharelau:
this is also fix for
https://github.com/twosigma/beaker-notebook/issues/3250
https://github.com/twosigma/beaker-notebook/issues/3243

scottdraves:
i don't see what you mean about the problem with using the clicked location.
the image above looks fine.  what's the problem?

also still missing: the item you are hovered over should highlight (get fatter).
similar to the legend, if you move the mouse quickly over the item it should not respond.
this is true of the tooltip in general.

scottdraves:
:+1:

scottdraves:
kick jenkins

Issue #3214, opened 2015-12-31
scroll bars in ambiance theme


scottdraves:
table H and V scrollbars should have dark gray bkg, like the codemirror scrollbar:
![screen shot 2015-12-24 at 12 21 15 pm](https://cloud.githubusercontent.com/assets/963093/11996816/825b2f82-aa39-11e5-833b-4d91dd352531.png)
![screen shot 2015-12-24 at 12 23 43 pm](https://cloud.githubusercontent.com/assets/963093/11996818/853389e8-aa39-11e5-92c3-82275de7514f.png)

and easyform scrollbars should have the same dark bkg as the horizontal codemirror scrollbar.
<img width="157" alt="screen shot 2015-12-23 at 12 58 21 am" src="https://cloud.githubusercontent.com/assets/963093/11971776/66ff09ea-a910-11e5-86dc-31e1ec2f4477.png">

PR #3260, opened 2016-01-06
#3214. scroll bars in ambiance theme

maxyeg:


scottdraves:
looks good for table scrollbars and easyform list scrollbars, but:

this breaks the resize handle on easyform text areas.
instead of the diagonal lines i am just seeing a gray box,
and it becomes detached from the area:
screen shot 2016-01-06 at 11 19 06 am

also i forgot to mention it in the issue but can you please also apply this style to the scrollbars in the language manager:
screen shot 2016-01-06 at 11 22 47 am

scottdraves:
is it possible to style the main scrollbar for the notebook, all the way on the right?
if so, please include it.

scottdraves:
replaced with

Issue #2388, opened 2015-09-18
[Native Beaker] File -> Publish dialog broken


scottdraves:
cannot enter anything.  it should be a native top-level dialog.

mnavasiolava:
relates to #2895

PR #3284, opened 2016-01-08
[Native Beaker] File -> Publish dialog broken #2388

### LINKED ISSUE #2895: Electron HTML form input un-editable
I had previously created HTML form elements to drive other beaker cells, this works well in the browser. When I open the notebooks in the Electron version they render correctly, but the input fields are not editable. Is this a bug or do I need to re-implement the way I am taking user input?

mnavasiolava:
also fixes #2895 (Electron HTML form input un-editable)


scottdraves:
:+1:

Issue #3283, opened 2016-01-08
R crashes on this code


scottdraves:
```
wfact = cut(women$weight,3)
table(wfact)
```

the websocket has this syntax error:

```
[{
  "data" : {
    "type" : "SimpleEvaluationObject",
    "update_id" : "548199c8-6660-4776-b047-ee128e9e7b87",
    "expression" : "wfact = cut(women$weight,3)\ntable(wfact)",
    "status" : "FINISHED",
    "payload" : "wfact\n(115,131] (131,148] (148,164] \n        6         5         4 ",
    "jsonres" : [\\n6, 5, 4\\n],
    "outputdata" : [ ]
  },
  "channel" : "/object_update/548199c8-6660-4776-b047-ee128e9e7b87"
}]
```

which shows up in the console as

```
Uncaught SyntaxError: Unexpected token \
```

scottdraves:
this?
https://github.com/twosigma/beaker-notebook/pull/3156

maxyeg:
I think here is the same situation: https://github.com/duncantl/RJSONIO/issues/18
I've tested it in R shell. The results:

`> rjson::toJSON("\u003")`
`[1] "\"\\u0003\""`
`> RJSONIO::toJSON("\u003")`
`[1] "[ \"\003\" ]"`

In beaker we use the last one.

maxyeg:
I've tried to configure jackson `ObjectMapper` with `ALLOW_UNQUOTED_CONTROL_CHARS` and that helps to avoid parsing exception.
But in the end it fails on the client in `cometd.js` at `org.cometd.JSON.fromJSON(response);`

scottdraves:
thanks for researching.  looks like it is a RJSONIO bug.
fixing it in R is probably too slow though, I think doing it in Java (like you do now) is OK.
but can you reproduce the results of rjson lib so the result is proper unicode?

scottdraves:
my previous comment moved here so it is not lost:

i am thinking the original solution to just add the quotes is wrong: https://pub.beakernotebook.com/#/publications/5693dec2-8e59-479f-b1a9-921d358ea27f

PR #3297, opened 2016-01-11
#3283. R crashes on this code

maxyeg:
Examples to check:
`beaker::set('test','\003')`
and 
`wfact = cut(women$weight,3)`
`table(wfact)`


scottdraves:
i am thinking the original solution to just add the quotes is wrong: https://pub.beakernotebook.com/#/publications/5693dec2-8e59-479f-b1a9-921d358ea27f

scottdraves:
took this fix because it does prevent crashing afaik, though answers for control chars are still wrong.

Issue #2773, opened 2015-10-21
version string appears hardcoded in installer


scottdraves:
![screen shot 2015-10-20 at 8 09 27 pm](https://cloud.githubusercontent.com/assets/963093/10624468/8f6aba40-7766-11e5-9c90-cc75f9f48437.png)

PR #3298, opened 2016-01-11
version string appears hardcoded in installer #2773

kpaharelau:
Property 'productVersion' has been added to the 'beaker-notebook/launcher/electron/gradle.properties'. Current value is changed to the 1.5


scottdraves:
This just moves the constant from one place to another.

it should be computed from git, use the same string as we report in the "about" box.
you should use getDescribe() from core/config/builds/dev/build.gradle

scottdraves:
:+1:

scottdraves:
old version string is still hardcoded in makeWinInstaller.iss

also, please name final build product beaker-notebook-$(version)-electron-windows.exe instead of beaker_setup.exe

scottdraves:
:+1:

Issue #2779, opened 2015-10-21
add option for location of prefs and temp files


scottdraves:
this is especially important on windows in order to make a portable version of beaker that you can run from a thumb drive and does not modify any other volume.

this would be a cmd line, like --plugin-option and --public-server

scottdraves:
--portable
puts what normally goes in ~/.beaker instead into the root of the drive where the application is mounted G:.beaker or whatever drive.

altavir:
I think that it is better to use beaker own directory since it won't contaminate external directory structure. The best solution is to allow user to configure paths via configuration files or add additional path parameter to `beaker.command` since it covers all possibliies.

There is also a problem with portability comming from the fact that JVM shipped with beaker is 64-bit and some systems still have 32-bit operating system.

scottdraves:
OK within the application then.

agreed about the 64/32 issue, this is a doc bug first https://github.com/twosigma/beaker-notebook/issues/1498, will address for this release.

aakinlalu:
I am considering using portable because corporate firewall. Do I have to install languages I need in the portable folder or thumb drive?

scottdraves:
they can be anywhere, on thumb drive or not.

BeakerNotebook.com
ScottDraves.com
twitter.com/Scott_Draves

PR #3299, opened 2016-01-11
add option for location of prefs and temp files #2779

kpaharelau:
--portable
puts what normally goes in ~/.beaker instead into the folder where the application is mounted - {path_to_beaker_application}/.beaker .

### DIFF
--- a/core/src/main/java/com/twosigma/beaker/core/module/config/DefaultBeakerConfig.java
+++ b/core/src/main/java/com/twosigma/beaker/core/module/config/DefaultBeakerConfig.java
@@ -100,8 +100,7 @@ public class DefaultBeakerConfig implements BeakerConfig {
 
     if (pref.getPortable()){
       //get parent for the 'core' folder
-      Path path = Paths.get("").toAbsolutePath();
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/twosigma/beaker/core/module/config/DefaultBeakerConfig.java
```diff
@@ -94,7 +96,18 @@ public DefaultBeakerConfig(BeakerConfigPref pref, GeneralUtils utils)
     this.useKerberos = pref.getUseKerberos();
     this.portBase = pref.getPortBase();
     this.reservedPortCount = 4;
-    this.dotDir = System.getProperty("user.home") + "/.beaker/v1";
+
+
+    if (pref.getPortable()){
+      String path = Paths.get("").toAbsolutePath().toString();
+      boolean endsWithSlash = path.endsWith(File.separator);
+      path = path.substring(0,
```
scottdraves: what does this do?  is this removing the last directory?
if so then better to use subpath and getNameCount?
we should not do so much detailed string manipulation of filenames.

kpaharelau: good point

scottdraves:
:+1:

Issue #3277, opened 2016-01-08
sort the languages


scottdraves:
it's too hard to find languages in the list, so put them in alphabetical order.

scottdraves:
<img width="575" alt="screen shot 2016-01-07 at 10 33 32 pm" src="https://cloud.githubusercontent.com/assets/963093/12189769/017c4c32-b58f-11e5-8e32-260478174e9f.png">

scottdraves:
this is a regression, maybe from lodash upgrade?

maxyeg:
It was introduced in angular update(from 1.2 to 1.4). They changed `ng-repeat` behavior.
You can see it in angular's repo here: https://github.com/angular/angular.js/commit/c260e7386391877625eda086480de73e8a0ba921#diff-f7b537c61e90325996f3c86b13cd53c4L345

They removed `collectionKeys.sort();`

If we are interested only in how buttons look in dialog we can leave my changes(https://github.com/twosigma/beaker-notebook/commit/e60c1a8ed3f2d1494e7b9dabecbbc9701b46947d)

Another solution will be to sort maps with evaluators definitions:
https://github.com/twosigma/beaker-notebook/blob/master/core/src/main/web/app/beaker.js#L394
https://github.com/twosigma/beaker-notebook/blob/master/core/src/main/web/plugin/init/addevalplugins.js
https://github.com/twosigma/beaker-notebook/blob/master/core/src/main/web/plugin/init/embeddedBeakerConfig.js#L37

tomca32:
Why can't we just use 'orderby' filter?

```
ng-repeat="(pluginName, pluginStatus) in evalTabOp.getEvaluatorStatuses() | orderBy: 'pluginName'"
```

https://docs.angularjs.org/api/ng/filter/orderBy

maxyeg:
I think `orderBy` filter works with arrays.
Adding just `| orderBy: 'pluginName'` doesn't work. I have to convert object which looks like `{'HTML' : '...', 'Latex' : '...'}` into array.
But keys and values from original object are used within `ng-repeat` loop.

maxyeg:
I see an example of using orderBy with objects here(added intermediate toArray filter): http://stackoverflow.com/questions/26474920/order-by-object-key-in-ng-repeat
What do you think?

tomca32:
Interesting, that does complicate things. Do we need sorting by object key anywhere else in the project? If so then maybe we can implement that filter.

tomca32:
[...]

PR #3300, opened 2016-01-11
#3277. sort the languages

maxyeg:


tomca32:
I've said this in the issue but I guess I should have mentioned it here instead.

Why can't we just use 'orderby' filter?

    ng-repeat="(pluginName, pluginStatus) in evalTabOp.getEvaluatorStatuses() | orderBy: 'pluginName'"

https://docs.angularjs.org/api/ng/filter/orderBy

scottdraves:
:+1:

Issue #1561, opened 2015-05-06
show progress even when output is hidden


scottdraves:
if you evaluate a cell and its output is hidden, then there is no feedback that the cell is running or that there is a result when it eventually appears.  this is confusing for users.

when the cell is running, the progress indicators (spinner, elapsed, cancel X, running) should appear.

PR #3313, opened 2016-01-12
show progress even when output is hidden #1561

kpaharelau:
Checking for show output has been improved. Progress  is showed  now for the result status "RUNNING"


scottdraves:
:+1:

Issue #2678, opened 2015-10-09
[Native Beaker] On Mac, give documents an icon


scottdraves:
Notebook files should appear with our branded icon: a combo of beaker logo and the generic, standard doc icon.

PR #3314, opened 2016-01-12
[Native Beaker] On Mac, give documents an icon #2678

mnavasiolava:


scottdraves:
:+1: 
not working for me but i think it's because launch services DB is screwed up on my dev machine.
will test this further on other machines, i am pretty sure it's a step in the right direction.

PR #3325, opened 2016-01-13
#3292. Scala on windows broken

maxyeg:
Remove usage of pathing jar as it breaks scala evaluator.
Minor fixes: removed unused imports, added files separator. Set addDynamicDir property to outDir as it's done in groovy in DynamicClassLaoderSimple.


scottdraves:
:+1:

PR #3326, opened 2016-01-13
Ambiance: modularizing css

alice-kroch:
Please, review the code


tomca32:
OK, so no more fixed values in that file, I like that. Could we move all the rules into a separate file called theme.scss or basetheme.scss. Then this file would have only the color definitions and one rule saying something like .ambiance-theme extends .base-theme (or whatever the sass syntax is for extending)

That way all theme files would just have different color definitions and would be really easy to manage. I'm pretty confident that this is possible with sass since it can import and extend rules: http://sass-lang.com/documentation/file.SASS_REFERENCE.html

scottdraves:
right, the goal is to define a new theme with a file that has just 10-15 color definitions.

alice-kroch:
Updated. Please, check this approach

tomca32:
Modularization looks good. We'll have to update it with additional css changes that were added to master in the meantime.

scottdraves:
please bring this up to date with master

tomca32:
There are some totally new properties in the ambiance scss file like cursor image on various elements that need to be included.

alice-kroch:
Updated. Please, check.

scottdraves:
markdown blockquote fix has been lost it seems.

alice-kroch:
Fixed.

scottdraves:
kick jenkins

scottdraves:
:dolphin:

alice-kroch:
Up to date with master. Please, check.

PR #3340, opened 2016-01-14
#3283. R crashes on this code

maxyeg:
Removed double quoting.
Using this example looks fine: https://pub.beakernotebook.com/#/publications/5693dec2-8e59-479f-b1a9-921d358ea27f


scottdraves:
:+1:

Issue #3346, opened 2016-01-14
regression ambiance: language manager tab


scottdraves:
![screen shot 2016-01-14 at 12 14 31 pm](https://cloud.githubusercontent.com/assets/963093/12331712/c34f0f10-bab8-11e5-8d8d-c5219534e489.png)

scottdraves:
there is still a problem, see the PR
https://github.com/twosigma/beaker-notebook/pull/3348
i merged it but there is more :(

ideally every time a plugin becomes ready, and the tab is added, then that tab should come to the front.

scottdraves:
cannot repro

PR #3348, opened 2016-01-14
regression ambiance: language manager tab #3346

mnavasiolava:


scottdraves:
the labels on the background tabs should be yellow not blue.
here's how it was:
screen shot 2016-01-14 at 2 01 13 pm

also when you hover over the background tabs, the mouse should be a finger/hand.

scottdraves:
also, when the tabs first appear, none of them are selected, and it looks pretty funny.
the tabs should be selected as they appear.
or if there are any tabs, then at least one tab should be selected.
screen shot 2016-01-14 at 3 32 46 pm

Issue #3353, opened 2016-01-15
windows electron installer version can be wrong


scottdraves:
if you generate an installer with one version,  and then change to a new branch, and then build another installer, the exe file gets the right name, but the installer in the "Welcome to the Beaker Setup Wizard" page reports the old version.

probably the template file was not run because it already existed.

PR #3366, opened 2016-01-15
windows electron installer version can be wrong #3353

kpaharelau:
> probably the template file was not run because it already existed.

Yes - I have added './templates/makeWinInstaller.iss' to the 'cleanElectron' task. This task should be executed before 'makeInstall' task


scottdraves:
:+1:

Issue #3259, opened 2016-01-06
[Native Beaker] Show stdout/err menu doesn't work


mnavasiolava:
1. open notebook
2. select View-> Show stdout/err menu item
   stdout div is not shown

if you e.g. expand/collapse a section or run a cell stdout div will be shown
1. click "Hide" button on stdout div
2. stdout div is hidden
3. look at View-> Show stdout/err menu item - there is a check next to menu item (but it shouldn't be because div is hidden)

mnavasiolava:
the same for Show Hierarchy, Delete all output cells, Collapse/Open all sections menus

PR #3374, opened 2016-01-15
[Native Beaker] Show stdout/err menu doesn't work #3259. 

mnavasiolava:
partial fix for updating notebook after menu was clicked.
there is still a bug with updating checks in menu if they were changed from the application (like for example when we hide stdoutput div with the "hide" button and menu should be unchecked)


scottdraves:
:+1:

Issue #3397, opened 2016-01-18
add target for portable version


scottdraves:
electron build for windows should have another target, like makeBundle but called
makePortableBundle.  it does the same thing, but then also edits beaker.command.bat
(down in beaker/resources/dist) to give the --portable option to the core.
and then it should zip it up and name it like "beaker-notebook-1.4.2-0-ge55c059-electron-windows.zip"

PR #3402, opened 2016-01-19
add target for portable version #3397

kpaharelau:
task 'makePortableBundle' has been added.

also 'cleanIss' functionality has been restored - was lost while merge


scottdraves:
this should zip up the electron app, not the old app that opens your web browser.

scottdraves:
ok, zips the electron app but now missing --portable :(

Issue #3383, opened 2016-01-17
area plotitems should expand in all 4 directions


scottdraves:
not just up, they should expand down, left and right too.
when you hover over them in the legend.

PR #3408, opened 2016-01-20
area plotitems should expand in all 4 directions #3383

kpaharelau:
I am using svg filter for a highlighting. It is more simple than old code.


scottdraves:
yes, the filter is simpler, that's great, i like it :)

1) but it is too transparent, can you keep the original look, maybe using the dilate operator?
2) the reference to svgAreaFilter would break when beaker is embedded.  see how we had to fix Triangle in plottext.js?  please make a function to create urls like this and call it from each place.
3) use this technique for all hover highlighting.
4) it should animate in and out, like when you hover over lines.

scottdraves:
kick jenkins

scottdraves:
kick jenkins

PR #3415, opened 2016-01-21
Parallelize tests and increase build reliability

tomca32:
Isolating autotranslation test into its own file.

Don't merge this yet. I want to see how is CI performing.


tomca32:
kick jenkins

PR #3416, opened 2016-01-22
fixes #3399, adds language env config option, $GROOVY_JAR_PATH variable

### LINKED ISSUE #3399: language preferences to include location of groovy install to use
something like:

"Groovy" : {
  "options" : "-Xmx8g"
  "path" : "$HOME/groovy-2.4.3"
}

// which would mean $HOME/groovy-2.4.3/bin, etc. would be used for the groovy plugin

dkobylarz:
this PR lets the environment variables from beaker.pref.json be injected when language scripts are executed. The env values are read into com.twosigma.beaker.core.module.config.DefaultBeakerConfig#pluginEnvps

```
{
  "pref-format" : "1",
  "allow-anonymous-usage-tracking" : false,
  "edit-mode" : "default",

  "languages" : {
    "Groovy" : {
      "env" : {
        "GROOVY_JAR_PATH": "/var/lib/groovy/groovy-all-2.5.0-SNAPSHOT.jar",
        "someVar": "someValue"
      }
    }
  }
}
```

the groovy plugin env build script update honors that particular env variable (GROOVY_JAR_PATH) to override default built-in groovy jars

### DIFF
--- a/core/src/main/java/com/twosigma/beaker/core/module/config/DefaultBeakerConfig.java
+++ b/core/src/main/java/com/twosigma/beaker/core/module/config/DefaultBeakerConfig.java
@@ -170,6 +170,8 @@ public class DefaultBeakerConfig implements BeakerConfig {
 
     augmentPluginOptions();
 
+    augmentPluginEnv();
+    
     this.publicServer = pref.getPublicServer();
     this.useHttpsCert = pref.getUseHttpsCert();
     this.useHttpsKey = pref.getUseHttpsKey();
@@ -469,5 +471,34 @@ public class DefaultBeakerConfig implements BeakerConfig {
     } catch (Exception e) {
       // ignore
     }
+  }
+  
+  @SuppressWarnings("unchecked")
+  private void augmentPluginEnv() {
+	  
+	  Map<String, JSONObject> plugins = (Map<String, JSONObject>) this.prefs.get("languages");
+	  for(Map.Entry<String, JSONObject> entry : plugins.entrySet()) {
+		  
+		  String plugin = entry.getKey();
+		  
+		  Object env = entry.getValue().get("env");
+		  
+		  if(env == null) continue;
+		  
+		  if(!(env instanceof Map)) { throw new RuntimeException("'env' value must be a map"); };
+		  
+		  Map<String, String> m = (Map<String, String>) env;
+		  
+		  List<String> envL = new ArrayList<String>();
+		  
+		  for(Map.Entry<String, String> e : m.entrySet() ) {
+			  
+			  envL.add(e.getKey() + "=\"" + e.getValue().replace("\"", "\\\"") + "\"");			  
+		  }
+		  
+		  pluginEnvps.put(plugin, envL.toArray(new String[envL.size()]));
+		  
+		  
+	  }
   }
 }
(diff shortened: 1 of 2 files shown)


scottdraves:
thanks for the PR!  i will review and get back to you.  it seems reasonable from what i see here.
did you sign a CLA?

tomca32:
Java exception on build:

1) Error injecting constructor, java.lang.NullPointerException
  at com.twosigma.beaker.core.module.config.DefaultBeakerConfig.<init>(DefaultBeakerConfig.java:91)
  at com.twosigma.beaker.core.module.config.DefaultBeakerConfig.class(DefaultBeakerConfig.java:91)
  while locating com.twosigma.beaker.core.module.config.DefaultBeakerConfig
  while locating com.twosigma.beaker.core.module.config.BeakerConfig
    for parameter 0 at com.twosigma.beaker.core.rest.PluginServiceLocatorRest.<init>(PluginServiceLocatorRest.java:69)
  at com.twosigma.beaker.core.module.URLConfigModule.configureServlets(URLConfigModule.java:77)
  while locating com.twosigma.beaker.core.rest.PluginServiceLocatorRest

1 error
    at com.google.inject.internal.InjectorImpl$4.get(InjectorImpl.java:987)
    at com.google.inject.internal.InjectorImpl.getInstance(InjectorImpl.java:1013)
    at com.twosigma.beaker.core.Main.main(Main.java:211)
Caused by: java.lang.NullPointerException
    at com.twosigma.beaker.core.module.config.DefaultBeakerConfig.augmentPluginEnv(DefaultBeakerConfig.java:480)

dkobylarz:
@scottdraves a while ago Vital AI  CTO signed a corporate CLA. This is the company I work at.

scottdraves:
thanks for the PR, i like the idea of setting envars in the prefs like this.
but when i run it, i have the same problem as the other one, the init cells of the time series tutorial fail.

scottdraves:
so i like the envar part of this, but the part that allows groovy jars to be replaced is a pretty big hack.
for example, it only works if the replacement version has the same API.
i have to think there is a better way to achieve the same goal.
we have this plan: https://github.com/twosigma/beaker-notebook/wiki/Refactoring
that would allow a separate groovy backends for different versions.
right now we are thinking we'll get there in 3-6 months.
would that work for you?  can you separate out the part that sets the envars only?

hadfield:
hi scott - 

in our case we need to use a particular version of groovy to run cells, and rather than build a new version of Beaker from scratch, having the "groovy home" settable in the preferences seems the way to go.

This follows the same pattern as Python in language preferences here:
https://github.com/twosigma/beaker-notebook/wiki/Language-Preferences

since beaker also uses groovy to launch and manage the groovy evaluator process, it does seem to be a bit of a "hack" and could cause some issues with Beaker's internal version of groovy being different from the evaluator of groovy.  Other than totally separating out groovy as a plugin in the same way python is, I'm not sure how else to handle it.

In our particular case we have a locally build groovy to include certain groovy patches which aren't released yet, but the same scenario would be true for any case in wanting to use a different version of groovy than the version embedded with Beaker.

the refactoring link you point to seems to indicate that groovy would be treated the same as python and be an "external" plugin -- that would still have the same issue of an embedded groovy of a particular version and the plugin groovy of a particular version where they may not match, no?

scottdraves:
once the plugin is external, you can have multiple plugins with different versions of groovy.

Issue #2888, opened 2015-11-07
make plot and table interactions programmable


scottdraves:
from groovy (initially but also js/python/r) you should be able to attach actions to various elements in the plots and tables, so you can have a closure be called or a cell tag be executed when you user clicks on a cell in a table, or clicks on a data point in a plot.

The plotitems classes like Line() and Bars() should have another parameter onClick that takes a closure that gets called when you click on that item.  The closure receives an object with the x/y coordinates of the item you clicked on (values from the xs, ys arrays), as well as the index into the xs/ys arrays, and the line object itself, and general event info (with things like mouse coordinates and modifier keys).  also allow onKey().  also allow a string instead of a closure, then it runs cells with that tag and defines a global with the event object.  see the implementation of EasyForm for an example of how to get a closure called from JS.

scottdraves:
plots done.  tables soon!

Pandaski:
Hi @scottdraves 
Could you let me know if the programmable interactions is implemented in Python based Plots?
I am trying to implement an interactive action in Python BeakerX Plot like the Groovy version done. But i cannot find any document about that, even I cannot find related attribute in the Line object.

I know this branch is closed in 2 years ago. So your (or any other people's) reply will be highly appreciated. Thx!

scottdraves:
Thanks for the report, I opened https://github.com/twosigma/beakerx/issues/7523

PR #3417, opened 2016-01-22
make plot and table interactions programmable #2888

mnavasiolava:
Please find notebook with examples in our tutorial ("Plot actions").
Looking forward to your comments and remarks


scottdraves:
thanks this looks great, need to do some tests...

scottdraves:
kick jenkins

scottdraves:
:+1:

plugin/jvm/src/main/java/com/twosigma/beaker/chart/serializer/ObservableChartSerializer.java
```diff
@@ -0,0 +1,27 @@
...
+import com.twosigma.beaker.chart.ObservableChart;
+import com.twosigma.beaker.chart.actions.ChartObjectManager;
+import com.twosigma.beaker.jvm.updater.UpdateManager;
+import org.codehaus.jackson.JsonGenerator;
+import org.codehaus.jackson.map.JsonSerializer;
+
+import java.io.IOException;
+
+@Singleton
+public abstract class ObservableChartSerializer<T extends ObservableChart> extends JsonSerializer<T> {
```
tomca32, line 15: @mnavasiolava Why is this both a Singleton and abstract? I don't know much about Guice so I might be missing something but I don't see how can an abstract class be a Singleton, or be instantiated at all.

mnavasiolava, line 15: @tomca32 you're right, singleton annotation shouldn't be there.

Issue #3330, opened 2016-01-13
implement clone method


scottdraves:
for all plotitems and plots

PR #3429, opened 2016-01-26
implement clone method #3330

mnavasiolava:
org.apache.commons.lang3.SerializationUtils.clone is used to clone plot objects


scottdraves:
:+1:

Issue #3424, opened 2016-01-25
add eval sequence numbers 


scottdraves:
on the cell output menu, just after the Elapsed time item, add an item "Evaluation Sequence: XXX"
each time we evaluate a cell (in any language) it should increment a counter.
counter starts out at 1.
the value of this counter is shown here if it is defined (if a cell has not been evaluated, then the item does not show.
the value is not saved in the notebook or shared.

this helps users figure out if a cell needs to be re-evaluated because it's old

PR #3435, opened 2016-01-27
add eval sequence numbers #3424

mnavasiolava:


scottdraves:
something's not right, i am seeing the sequence number stuck at 1

scottdraves:
ah.  the sequence number should be global to the notebook, not per cell.

scottdraves:
:+1:

Issue #3426, opened 2016-01-26
Scaling of Electron UI


RhysU:
Running the 1.5 Electron build on a 2560x1440 display produces very small fonts.  In the browser, zooming can scale things up to a readable level.  If possible, please permit UI scaling with Electron.

scottdraves:
you are right, i think this used to work, but it's broken now.
thanks for the report.
should be on same cmd key as in browser,
plus in view menu, add items "zoom in", "zoom out", and "actual size".

scottdraves:
so actually the cmd keys  do work, but only on the notebook windows, not on the "start page".

scottdraves:
the cmd keys do not work if you are editing a text or code cell.

PR #3440, opened 2016-01-28
Scaling of Electron UI #3426

mnavasiolava:


scottdraves:
Like in Chrome, it should take command-= as well as command-+
and in the menu, just show cmd-+

mnavasiolava:
command-= works for me
and about cmd-+ there is an issue in electron https://github.com/atom/electron/issues/1507

Issue #3398, opened 2016-01-18
windows version leaves nginx dir


scottdraves:
every time you run the windows version it creates an nginx directory in ~/.beaker/v1
it should be deleted on application exit.
this is true in the regular application (which you control-c in a terminal),
and also the electron version (when you close all the windows or press control-q).

PR #3442, opened 2016-01-29
windows version leaves nginx dir #3398

kpaharelau:
I found that  org.apache.cxf.helpers.FileUtils.delete method works fine.

Electron application on the Windows platform has big problem. We use 'taskkill' command for a java process stopping. In this case java application doesn't receive any signals and ShootdownHook doesn't work.  I have used communication across console output. Idea is very strange but it is works

### DIFF
--- a/core/src/main/java/com/twosigma/beaker/core/rest/UtilRest.java
+++ b/core/src/main/java/com/twosigma/beaker/core/rest/UtilRest.java
@@ -153,6 +153,14 @@ public class UtilRest {
     return clean(content);
   }
 
+
+  @GET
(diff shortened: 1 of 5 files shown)


scottdraves:
thanks cit but i agree, it seems pretty hacky.
is there a way to send a message to the server to exit instead of task kill?
about a rest call to exit?

kpaharelau:
I think solution is a service that returns list of temporary folders for delete

scottdraves:
kick jenkins

scottdraves:
worked on mac but still left an nginx dir on linux.

Issue #3438, opened 2016-01-27
Nodejs nodePlugin doesn't work on Windows bin


bcaller:
The variable `bin` is not assigned to the value from the config file. Instead it points to the function `bin(x)`. Therefore the `subprocess.call` does not work.

`subprocess.call([bin, "app/app.js", port, "127.0.0.1"], shell=True)`

https://github.com/twosigma/beaker-notebook/blob/master/plugin/node/src/dist/nodePlugin#L27

bcaller:
More permanent link: https://github.com/twosigma/beaker-notebook/blob/ed8fd62cdb6d2207beb3bbb3f990e14311ece337/plugin/node/src/dist/nodePlugin#L27

scottdraves:
thanks for the report, should not be hard.

PR #3444, opened 2016-01-29
#3438. Nodejs nodePlugin doesn't work on Windows bin

maxyeg:
Usage of `bin` here was copy-pasted from other plugins run scripts.

Here you can see that I change folder to nested one. That is because if I execute the script from current folder it tries to run directly `node.js` file from current folder. But we need to run Node starter from the `PATH`.
So to avoid shell confusion I switched folder to `app` where there are no confusing files.
But if someone tries to run: `subprocess.call(["node", "app/app.js", port, "127.0.0.1"], shell=True)` without changing the folder and Node js will run fine, I will consider it to be my local problem and will remove `os.chdir("app")` statement.

### DIFF
--- a/plugin/node/src/dist/nodePlugin
+++ b/plugin/node/src/dist/nodePlugin
@@ -24,6 +24,7 @@ self_path = sys.argv.pop()
 os.chdir(os.path.dirname(self_path))
 
 if os.name == 'nt':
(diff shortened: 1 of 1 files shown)


scottdraves:
:+1:

Issue #3332, opened 2016-01-13
support 64 bit integers in tables and autotranslation


scottdraves:
similar to how we do nanoplots, we should use big.js and strings.
but for general autotranslation, you need to distinguish strings from 64-bit ints.
so they become objects `{'type': 'int64', 'value': '4611686018427387904'}`
for values > 10e53 or whatever the JS max is.

scottdraves:
https://github.com/twosigma/beaker-notebook/pull/3454 provides a good start but leaving this open as we get feedback on that and work towards more general solution.

PR #3454, opened 2016-02-03
support 64 bit integers in tables and autotranslation #3332

kpaharelau:
Support for BigInteger has been added.

Use for testing

https://www.dropbox.com/s/3z7wu8300097bau/3332.bkr?dl=0


tomca32:
kick jenkins

scottdraves:
OK this is a good start but the issue is much larger.  Eventually we are going to need the object format specified in the issue.  But we can get there in some smaller steps.

First step, it should work with Longs in addition to BigIntegers.

Also, add your tutorial as a separate notebook to the tutorials.

scottdraves:
:+1:

scottdraves:
please bring up to date with master.

Issue #3449, opened 2016-02-02
Strings with double slash ("//"), e.g., URLs, broken in Java language cells


expz:
Strings with a double slash break the Java interpreter. The double slash is apparently interpreted as the beginning of a comment?
### Correct Behavior in Terminal

Open a command line with java. Create `Test.java`:

``` java
public class Test {
  public static void main(String[] argv) {
    String url = "http://www.example.com";
    System.out.println("--- Bug not found ---");
  }
}
```

Then it should run correctly:

```
$ javac Test.java
$ java Test
--- Bug not found ---
```
### Bug in Beaker Notebook

Create a Notebook with Java enabled. Create a Java cell with the code:

``` java
String url = "http://www.example.com";
System.out.println("--- Bug not found ---");
```

Running the cell produces the error:

```
ERROR: java.lang.IllegalStateException: ';' expected
n("--- Bug not found';' expected
 not found ---");}
}unclosed string literal
 found ---");}
}
reached end of file while parsing
 ---");}
}
```

expz:
This will break tons of existing Java code that users would copy and paste, but it has an easy workaround for users just writing a little bit of code for themselves:

``` java
String url = "http:/" + "/www.example.com";
System.out.println(url);
```

scottdraves:
Thanks for the report, we'll get right on it.

PR #3457, opened 2016-02-04
Strings with double slash ("//"), e.g., URLs, broken in Java language…

kpaharelau:
… cells #3449

Problem was un the not correct comments removing.

We had after code normalizing

public class Test {
  public static void main(String[] argv) {
    String url = "http:
    System.out.println("--- Bug not found ---");
  }
}
'

I have added function 'removeComments' instead regex processing

### DIFF
--- a/plugin/javash/src/main/java/com/twosigma/beaker/javash/utils/JavaEvaluator.java
+++ b/plugin/javash/src/main/java/com/twosigma/beaker/javash/utils/JavaEvaluator.java
@@ -211,15 +211,6 @@ public class JavaEvaluator {
     return jac.doAutocomplete(sb.toString(), caretPosition);    
   }
 
-  protected enum FilterState{
-    IN_CODE,
-    CAN_BE_COMMENT_START,
-    INSIDE_STRING,
-    IN_COMMENT_BLOCK,
-    ON_COMMENT_LINE,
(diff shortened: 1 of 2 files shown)


scottdraves:
:+1:

tomca32:
Java Enums are pretty much classes, right?

Isn't it better then to have the enums have a method to do all this logic rather then switching the enum from the outside? It seems to me that would be more maintainable and cleaner.

Not a Java guy really so I might be talking crap.

scottdraves:
this code is copy-and-pasted from stackoverflow, changes are likely to just introduce bugs.

scottdraves:
fails on this:

```
String foo = "\\" + // comment
  "\\";
```

would also fail on a unicode escape.

scottdraves:
how about https://pub.beakernotebook.com/#/publications/56bc9b29-2439-437e-ae61-815fdc9d95e7

scottdraves:
the error is actually correct
:+1:

scottdraves:
still, just keeping track of 2 chars of history is inadequate.  what can you learn from this?
https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/java/com/github/javaparser/ast/comments/CommentsParser.java

scottdraves:
:+1:

Issue #3475, opened 2016-02-08
vim edit mode broken


scottdraves:
you should get a block cursor and ESC should toggle insert vs movement modes.

PR #3498, opened 2016-02-09
vim edit mode broken #3475

kpaharelau:
I changed width cursor for a better visualisation

.CodeMirror.cm-fat-cursor div.CodeMirror-cursor {
  width: 0.5em;
}

### DIFF
--- a/core/src/main/web/app/app.scss
+++ b/core/src/main/web/app/app.scss
@@ -325,6 +325,11 @@ background-color: rgb(205, 223, 214);
   display: block;
 }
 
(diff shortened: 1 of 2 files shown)


scottdraves:
i think the real problem is still there: `Uncaught TypeError: Cannot read property 'isVimOp' of null`

scottdraves:
ESC in code cell does not exit insert mode and return to movement mode.
also, the solid cursor blocks the visibility of the character that it covers.
it should be inverse.

scottdraves:
:+1:

Issue #3479, opened 2016-02-08
need API access to language manager settings


scottdraves:
add read-only access that does not show up in files or shares or beaker object (similar to the colors).

beaker.language.Groovy.classPath should get you a string.
should work in groovy and javascript, at least.

PR #3503, opened 2016-02-09
need API access to language manager settings #3479

mnavasiolava:


scottdraves:
ok this partly works, but if you try to use groovy only, it does not work.
see this notebook:
https://pub.beakernotebook.com/#/publications/56ba327b-9b29-4029-8152-535674e9c91b
if you run the 1st cell it fails with the visible error.
if you run the JS cell, then run the Groovy cell, then it works.
if you update the classpath, then evaluate the groovy cell, you get the old value, you need to run the JS to update it.

scottdraves:
also, see the 3rd cell, beaker object should not have this be visible, it should be removed.

core/src/main/web/app/helpers/helper.js
```diff
@@ -100,6 +100,41 @@
...
+        if (beaker) {
+          beaker.language = {};
+        }
+      },
+      setLanguageManagerSettingsToBeakerObject: function (plugin) {
+        var beakerObject = this.getBeakerObject().beakerObj;
+        if (beakerObject && beakerObject.language) {
+          var spec = plugin.spec;
+          var beakerLanguageSettings = {};
+          for (var property in spec){
```
tomca32: This is fine but you might want to use lo-dash sugar for iterating own properties:
https://lodash.com/docs#forOwn

I just think it cuts down on the boilerplate, but its your choice

mnavasiolava: good idea. thanks @tomca32

scottdraves:
:+1:

Issue #3487, opened 2016-02-09
column menu open and then click in code cell


scottdraves:
menu should go down.

PR #3505, opened 2016-02-09
column menu open and then click in code cell #3487

mnavasiolava:


scottdraves:
thanks that fixes the problem with clicking on code, but there are still some issues with how the column menus open and close:
1) if you open the table menu, and then open a column menu, the table menu stays open.  it should close.
2) if you open a column menu, and then click on another column menu, the first one closes which is good.  but the 2nd one should open too.
3) if you open a column menu and then click on the table to select a row, the menu should close.

mnavasiolava:
fixed

scottdraves:
:+1:

Issue #3489, opened 2016-02-09
column menu 


scottdraves:
the blue highlight should be moved down, it blocks the border line above it.  the border to its left is correct.
<img width="99" alt="screen shot 2016-02-08 at 10 04 31 pm" src="https://cloud.githubusercontent.com/assets/963093/12906613/286b2a54-ceb0-11e5-90ad-f91c8c000a17.png">

and the hamburger positioning should be the same on the index column for the whole table as it is on the data columns.  the positioning below is correct, the above is too high.  this one also has the border blocking problem.
<img width="317" alt="screen shot 2016-02-08 at 10 05 10 pm" src="https://cloud.githubusercontent.com/assets/963093/12906644/65820520-ceb0-11e5-85ca-e08d33fce875.png">

also the heading "Index" should be moved the right a little bit, so it does not crowd the menu.

PR #3508, opened 2016-02-10
column menu #3489

mnavasiolava:


scottdraves:
This is good but it needs a few more things:
1) Move "Index" further to the right, enough so that it's just to the right of the blue box when the menu is open.
2) Remove the up/down sort triangle indicator on the right side of the Index column.
3) If the index column is sorted ascending (as it starts out) then no indication of sort.
4) If the index column is sorted descending, then show the down-triangle in place of the menu, like the other columns.

mnavasiolava:
done

scottdraves:
:+1:

Issue #3511, opened 2016-02-11
bignums crash autotranslation


scottdraves:
```
Feb 10, 2016 10:39:46 PM com.twosigma.beaker.jvm.object.TableDisplay$DeSerializer deserialize
SEVERE: exception deserializing TableDisplay 
java.lang.NumberFormatException: For input string: "1130660681277582059829"
    at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
    at java.lang.Long.parseLong(Long.java:592)
    at java.lang.Long.parseLong(Long.java:631)
    at com.twosigma.beaker.jvm.object.TableDisplay$DeSerializer.getValueForDeserializer(TableDisplay.java:145)
    at com.twosigma.beaker.jvm.object.TableDisplay$DeSerializer.deserialize(TableDisplay.java:175)
    at com.twosigma.beaker.jvm.serialization.BasicObjectSerializer.deserialize(BasicObjectSerializer.java:267)
    at com.twosigma.beaker.jvm.serialization.NamespaceBindingDeserializer.deserialize(NamespaceBindingDeserializer.java:53)
    at com.twosigma.beaker.jvm.serialization.NamespaceBindingDeserializer.deserialize(NamespaceBindingDeserializer.java:34)
```

see the last example in the bignums tutorial in branch [spot/3511](https://github.com/twosigma/beaker-notebook/tree/spot/3511).

PR #3512, opened 2016-02-11
bignums crash autotranslation #3511

kpaharelau:
Fix String to BigInteger functionality(use constructor BigInteger(String)).


scottdraves:
:+1:

Issue #3460, opened 2016-02-04
show keyboard shortcuts in menus


scottdraves:
on the right of each menu item that has a keyboard shortcut.

PR #3516, opened 2016-02-11
Keyboard shortcuts in menus

alice-kroch:
Issue #3460 . Please, check.

### DIFF
--- a/core/src/main/web/app/helpers/helper.js
+++ b/core/src/main/web/app/helpers/helper.js
@@ -1077,7 +1077,8 @@
       hideLanguageManagerSpinner: function(error) {
         bkUtils.hideLanguageManagerSpinner(error);
       },
-      isElectron: bkUtils.isElectron
+      isElectron: bkUtils.isElectron,
+      isWindows: bkUtils.isWindows
     };
 
     return bkHelper;
--- a/core/src/main/web/app/helpers/helper.js
+++ b/core/src/main/web/app/helpers/helper.js
@@ -679,9 +679,6 @@
       cancelTimeout: function(promise) {
         return bkUtils.cancelTimeout(promise);
       },
-      trustAsHtml: function(piece) {
-        return bkUtils.trustAsHtml(piece);
-      },
       getHomeDirectory: function() {
         return bkUtils.getHomeDirectory();
       },
--- a/core/src/main/web/app/helpers/helper.js
+++ b/core/src/main/web/app/helpers/helper.js
@@ -679,6 +679,9 @@
       cancelTimeout: function(promise) {
         return bkUtils.cancelTimeout(promise);
       },
+      trustAsHtml: function(piece) {
+        return bkUtils.trustAsHtml(piece);
+      },
       getHomeDirectory: function() {
         return bkUtils.getHomeDirectory();
       },
--- a/core/src/main/web/app/helpers/helper.js
+++ b/core/src/main/web/app/helpers/helper.js
@@ -1078,7 +1078,7 @@
         bkUtils.hideLanguageManagerSpinner(error);
       },
       isElectron: bkUtils.isElectron,
-      isWindows: bkUtils.isWindows
+      isMacOS: bkUtils.isMacOS
     };
 
     return bkHelper;
--- a/core/src/main/web/app/mainapp/components/notebook/cell-directive.js
+++ b/core/src/main/web/app/mainapp/components/notebook/cell-directive.js
@@ -179,20 +179,20 @@
 
         $scope.cellview.menu.addItem({
           name: 'Delete cell',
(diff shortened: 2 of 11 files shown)


core/src/main/web/app/helpers/helper.js
```diff
@@ -679,6 +679,9 @@
       cancelTimeout: function(promise) {
         return bkUtils.cancelTimeout(promise);
       },
+      trustAsHtml: function(piece) {
```
scottdraves: we shoiuldn't need this

core/src/main/web/app/mainapp/components/notebook/cell-directive.js
```diff
@@ -179,17 +179,20 @@
 
         $scope.cellview.menu.addItem({
           name: 'Delete cell',
+          shortcut: bkUtils.isWindows ? bkUtils.trustAsHtml('Ctrl-Alt-D') : bkUtils.trustAsHtml('&#x2318;-Alt-D'),
```
scottdraves: there is a lot of repeated code here.  instead of putting the conditional on windows into every shortcut, the shortcut should have a form with two values for the two platforms (Linux and Windows keybindings are the same).  so how about:

```
shortcut: ['Ctrl-Alt-D', 'Cmd-Alt-D']
```

Then on Mac, Cmd gets replaced with &#x2318; and Alt with &#x2325;

if the shortcut value is just a single string, then it would be the same on all platforms.

core/src/main/web/app/helpers/helper.js
tomca32: Yeah I agree. This stuff is already trusted, it's coming from us. We don't need to clean it.

core/src/main/web/app/utils/basic/commonui.js
```diff
@@ -201,11 +201,34 @@
...
+        };
 
         $scope.getCustomMarkup = function(item) {
           return $sce.trustAsHtml(_.result(item, 'markup') || '');
-        }
+        };
+
+        $scope.getShortcut = function(itemShortcut) {
+         function replace(str) {
+            if (!bkHelper.isWindows) {
```
scottdraves: linux uses the windows keyboard, so this conditional is wrong

core/src/main/web/app/utils/basic/commonui.js
```diff
@@ -201,11 +201,34 @@
...
+              var mapObj = {
+                Cmd: "&#x2318;",
+                Alt: "&#x2325;"
+              };
+              str = str.replace(/Cmd|Alt/gi, function(matched){
```
scottdraves: instead of duplicating the Cmd and Alt, should loop over the keys of the mapObj and apply substitution to each one.

scottdraves:
thanks, much better, but still a few fixes.

scottdraves:
also:

there should be no dash, it should say Cmd N in the shortcut string, but actually in the final menu there should be no space at all between the symbol and the letter, so remove them i guess.

Cmd S is for save not for save-as.

please add shortcut shift cmd S for save as (and shift should display with fat arrow like on mac).

arrow keys should display as thin arrows like on mac.

core/src/main/web/app/utils/basic/commonui.js
```diff
@@ -201,11 +201,38 @@
...
+                Up: "&#x2191;",
+                Down: "&#x2193;"
+              };
+              var regexp = new RegExp(Object.keys(mapObj).join("|"),"gi");
+              str = str.replace(regexp, function(matched) {
+                return mapObj[matched];
+              });
+            }
+
+            return $sce.trustAsHtml(str);
```
scottdraves, line 227: trustAsHtml is not needed!

core/src/main/web/app/mainapp/components/notebook/cell-directive.js
```diff
@@ -179,17 +179,20 @@
 
         $scope.cellview.menu.addItem({
           name: 'Delete cell',
+          shortcut: ['Ctrl-Alt-D', 'AltCmdD'],
```
scottdraves: for mac, the shortcut strings should include a space or dash because this is hard to read.
but remove it in the replace() function.

core/src/main/web/app/utils/basic/commonui.js
alice-kroch, line 227: how can you bind unsafe string value with special symbols without marking it as trusted html? angular doesn't accept such binding. I don't understand why not to use it because it's [written in the documentation](https://github.com/angular/angular.js/blob/v1.5.0/src/ng/directive/ngBind.js#L150). Otherwise - include ngSanitize module, but does it make sense?

scottdraves:
:+1:

Issue #3517, opened 2016-02-11
Ambiance theme not working in full screen mode


tomca32:
Select an ambiance theme for Beaker.

Press Alt+F11 to enter full screen mode.

Observe an awful screen.
![screen shot 2016-02-11 at 10 03 53 am](https://cloud.githubusercontent.com/assets/3384784/12979696/d5d879b0-d0a6-11e5-978b-fccdfa098ad6.png)

PR #3526, opened 2016-02-12
Ambiance theme: full screen mode fix

alice-kroch:
Issue #3517. Please, check

### DIFF
--- a/core/src/main/web/app/styles/theme/base-theme.scss
+++ b/core/src/main/web/app/styles/theme/base-theme.scss
@@ -648,7 +648,7 @@
   }
 
   .CodeMirror-fullscreen {
-    background: $main-theme-panel-background;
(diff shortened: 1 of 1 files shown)


core/src/main/web/app/styles/theme/base-theme.scss
```diff
@@ -648,7 +648,7 @@
   }
 
   .CodeMirror-fullscreen {
-    background: $main-theme-panel-background;
+    background: $main-theme-panel-background !important;
```
tomca32, line 651: Is the style from the default theme overriding this? Is that why we need 'important'?

alice-kroch, line 651: yes, it have 'important'

scottdraves:
:+1:

PR #3528, opened 2016-02-15
write a test to ensure vim mode works. #3476

kpaharelau:
First test for checking insert and movement modes


scottdraves:
please reopen when you think it's ready.

Issue #3502, opened 2016-02-09
running cell before evaluator is ready


scottdraves:
can get stuck, it used to kick in when the backend finished spinning up.

PR #3531, opened 2016-02-16
#3502. running cell before evaluator is ready

maxyeg:
Cause: when I add evaluator by click in Language Manager, evaluator object has empty `name` field. And when evaluator is ready and it's time to execute waiting cell, manager can't find evaluator by name.

P.S. The same scenario is fine in Tutorials, because all evaluators are loaded from .bkr file and there `name` field contains evaluator name.

Solution: if `name` field is empty, then compare with `plugin` field which contains actual evalutorId.


scottdraves:
:+1:

PR #3546, opened 2016-02-18
write a test to ensure vim mode works. #3476

kpaharelau:
This is tests for a main vi functions. I am ready for new ideas :-)


test/tests/edit-mode.js
```diff
@@ -0,0 +1,121 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+
+var BeakerPageObject = require('./beaker.po.js');
+var path = require('path');
+var beakerPO;
+
+describe('Edit Mode', function() {
```
tomca32: It's called Edit mode but we're testing only Vim here. So lets call it "Vim mode"

test/tests/edit-mode.js
```diff
@@ -0,0 +1,121 @@
...
+    browser.waitForAngular();
+
+    beakerPO.newEmptyNotebook.click();
+  });
+
+  afterEach(function() {
+    beakerPO.closeNotebook();
+  });
+
+  it('test Vim toggle between insert and movement modes.', function() {
```
tomca32: No need to say 'test' in the beginning, of course it's a test :)

The "it" label will form sentences with the describe label so if we call it "toggles between normal and insert modes" then when the test runs it will say: "Vim mode toggles between normal and insert modes" which gives us a nice sentence and describes accurately what we are testing

test/tests/edit-mode.js
```diff
@@ -0,0 +1,121 @@
...
+
+    //click 'Esc' character - switch to movement mode
+    browser.actions().sendKeys('\uE00C').perform();
+    expect(element(by.css('.CodeMirror.cm-fat-cursor div.CodeMirror-cursors')).isPresent()).toBe(true);
+
+    beakerPO.setNormalEditMode();
+    expect(element(by.css('.CodeMirror.cm-fat-cursor div.CodeMirror-cursors')).isPresent()).toBe(false);
+  });
+
+  it('test Vim insert and delete.', function() {
```
tomca32: it's called insert and delete but there is no delete in this test.

test/tests/edit-mode.js
```diff
@@ -0,0 +1,121 @@
...
+
+    browser.actions().sendKeys('test').perform();
+    beakerPO.getCellInput().then(function (result) {
+      expect(result).toBe('test')
+    });
+
+    beakerPO.setNormalEditMode();
+  });
+
+  it('test Vim overwrite mode.', function() {
```
tomca32: How about: "R command replaces characters"

test/tests/edit-mode.js
```diff
@@ -0,0 +1,121 @@
...
+
+    beakerPO.getCellInput().then(function (result) {
+      expect(result).toBe('session')
+    });
+
+    beakerPO.setNormalEditMode();
+  });
+
+
+  it('test Vim uppercase.', function() {
```
tomca32: ~ command switches character case

test/tests/edit-mode.js
```diff
@@ -0,0 +1,121 @@
...
+    browser.actions().sendKeys('~~~~~~~').perform();
+
+    beakerPO.getCellInput().then(function (result) {
+      expect(result).toBe('MEETING');
+    });
+
+    beakerPO.setNormalEditMode();
+  });
+
+  it('test Vim cursor commands.', function() {
```
tomca32: $ and 0 commands go to the beginning and end of line, respectively.

tomca32:
Nice. Just a few suggestions. I think the test code looks good I would just improve a bit on some minor details.

Can we rename the test file to vim-mode.js (or something similar) ? Currently it's called edit-mode.js and that really isn't an accurate description if we are only testing vim inside it.

The idea of the description function in the tests is that it forms sentences with individual "it" tests. So for example the describe here is "Edit mode" and then the first test is called "test Vim insert and delete.". So when the test runs it's going to be called:

"Edit mode test Vim insert and delete." - Which doesn't really mean anything.

I've entered some suggested test labels in the code comments.

test/tests/edit-mode.js
```diff
@@ -0,0 +1,121 @@
...
+    browser.actions().sendKeys('0').perform();
+    beakerPO.getCellInputCursor().then(function (pos) {
+      expect(pos.ch).toBe(0);
+    });
+
+    beakerPO.setNormalEditMode();
+  });
+
+
+});
```
tomca32: Put end of line here. Git complains about it a lot.

test/tests/beaker.po.js
```diff
@@ -207,10 +229,37 @@ var BeakerPageObject = function() {
     return element(by.css('.code-cell-area .' + language + '-menuitem'));
   };
   this.cellEvaluatorDisplay = element(by.css('.code-cell-area .cell-evaluator-menu b'));
+
+  //CodeMirror API. See for information https://sharpkit.net/help/SharpKit.CodeMirror/SharpKit.CodeMirror/CodeMirror/
+
   this.setCellInput = function(code) {
-    browser.executeScript('$(".CodeMirror")[0].CodeMirror.setValue("' + code + '")');
+    return browser.executeScript('$(".CodeMirror")[0].CodeMirror.setValue("' + code + '")');
```
tomca32, line 236: Nice catch. I guess that's why it didn't work properly.

kpaharelau:
Ok. Thanks for comments. Will be done ASAP

scottdraves:
:+1:

PR #3557, opened 2016-02-19
Tom/2831 - Added Groovy autocompletion endpoint

tomca32:
- Added Groovy autocompletion endpoint
- Get matched token from CodeMirror if not passed from autocompletion backend
- Do not show autocompletion documentation window if documentation is empty


plugin/groovy/src/dist/groovy.js
```diff
@@ -149,6 +149,16 @@ define(function(require, exports, module) {
         cb(x, undefined, true);
       });
     },
+    getAutocompleteDocumentation: function(matchedWord, callback) {
+      $.ajax({
```
scottdraves: use bkHelper.httpPost

plugin/groovy/src/main/java/com/twosigma/beaker/groovy/rest/GroovyShellRest.java
```diff
@@ -20,10 +20,7 @@
 import com.twosigma.beaker.jvm.object.SimpleEvaluationObject;
 import java.io.IOException;
 import java.net.MalformedURLException;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.UUID;
+import java.util.*;
```
scottdraves: - is not allowed in imports

Issue #3527, opened 2016-02-14
I'd like globals in each cell to be automatically global in all cells


csells:
The beaker object as a universal hub for objects in all languages is PFM, but I find that I'm mostly using a single language and it quickly becomes tiresome to type "beaker." for access to all objects shared between cells, including cells of the same language (I primarily work in JS and HTML). I'd really love it if objects declare in global scope in each cell become global to the notebook and for each cell's scope resolution algorithm to include Beaker globals. That would make Beaker a joy to use.

scottdraves:
thanks. agreed https://github.com/twosigma/beaker-notebook/issues/797
unclear how to fix though maybe ES6 will help.

scottdraves:
@tomca32 got any ideas?

tomca32:
I will take a look at the JS evaluator to see what's possible.

csells:
I'd love this to work for other languages, too, but this one change made Beaker + HTML + JavaScript a joy!

scottdraves:
not sure what you mean, afaik the only language that had a problem sharing a regular variable between cells was JS.  Here are some examples: https://pub.beakernotebook.com/#/publications/56c7c197-f2b8-4233-bb6f-b6c0c1463d45

csells:
I didn’t know that. Every better!

On Feb 19, 2016, at 5:28 PM, Scott Draves <notifications@github.com<mailto:notifications@github.com>> wrote:

not sure what you mean, afaik the only language that had a problem sharing a regular variable between cells was JS. Here are some examples: https://pub.beakernotebook.com/#/publications/56c7c197-f2b8-4233-bb6f-b6c0c1463d45

Issue #797, opened 2014-10-23
javascript variables not shared between cells


scottdraves:
something about the JS namespace is broken: 
http://sharing.beakernotebook.com/gist/anonymous/de9a1c7dd49517aef353

also, looksl like user code is subject to "strict" but should not be.

as a workaround you can use the notebook namespace.

PR #3558, opened 2016-02-19
#3527 - Eval JS code indirectly so that the evaluated code will be in…

tomca32:
… non-strict mode

For those curious about what am I talking about: http://stackoverflow.com/questions/19357978/indirect-eval-call-in-strict-mode

This solves #3527 as well as #797


scottdraves:
please add a comment with a link to that SO page.

scottdraves:
:dolphin:

Issue #3490, opened 2016-02-09
add to column menu


scottdraves:
item: Filter...

it opens a text field field above the column, absolutely positioned above the input side of the cell above.  that search term applies to just this column.

it can be an expression like >10000

ie if it begins with > < >= <= or = then does numerical comparison.
otherwise does string match.

scottdraves:
so if the string is "> 1000" then it's a numerical comparison as described above.
please support another syntax as well, &&, so they can write:
">1000 && < 1010"

so to parse, split string on &&, then each of the terms is interpreted as above.

PR #3572, opened 2016-02-23
Filters in column menu #3490

mnavasiolava:


core/src/main/web/outputdisplay/bko-tabledisplay/bko-tabledisplay.js
```diff
@@ -77,6 +77,39 @@
...
+          var numValue = parseFloat(cellValue);
+          searchString = searchString.replace(/=/g, '==');
+          searchString = searchString.replace(/&&/g, '&& numValue');
+
+          try {
+            match = eval("numValue" + searchString);
+            if (!match) {
+              break;
+            }
+          } catch (e) {
```
tomca32, line 102: We don't want to do anything with this error? Not even log it?

mnavasiolava, line 102: this error occurs often enough, e.g. user start typing in filter field ">" or "> 0 && <", all these expressions cannot be evaluated and we'll get an error. idk if it makes sense to log it.

scottdraves:
this is great.  a few glitches:
1) when i drag a column to reorder it, there are a bunch of extra hamburger menus in it.
2) the keyboard bindings don't work over the header of the column if another element has had the focus (try clicking on a code cell, then hover mouse over column header vs body cell).

scottdraves:
i can't reproduce the problem with extra hamburger menus.  testing more...

scottdraves:
:+1:

PR #3575, opened 2016-02-24
Spot/tut

scottdraves:
we need to merge ES6 first then merge this.


tomca32:
OK, will do that after ES6 is merged

Issue #3563, opened 2016-02-21
highlighted menu item hides checkmark


scottdraves:
<img width="274" alt="screen shot 2016-02-21 at 2 15 56 pm" src="https://cloud.githubusercontent.com/assets/963093/13204748/b81cfc78-d8a5-11e5-8972-e2f31f687869.png">
<img width="278" alt="screen shot 2016-02-21 at 2 16 01 pm" src="https://cloud.githubusercontent.com/assets/963093/13204749/b8246c7e-d8a5-11e5-975d-1e250cbc2ca9.png">

PR #3576, opened 2016-02-24
Highlighted menu item hides checkmark

alice-kroch:
Issue #3563 . Please, check


scottdraves:
checkmark should change color the same as the item text.

scottdraves:
also needs to be fixed in ambiance mode.

scottdraves:
:+1:

PR #3584, opened 2016-02-24
Tom/1625 - ECMAScript 6 support

tomca32:
There is a checkbox for ES6 support in language manager. It defaults to true.


scottdraves:
cool.  but please use radio buttons instead of a checkbox,
name the group "JavaScript Version" with two options:
"Browser Native" and "ECMAScript 2015"

tomca32:
This PR includes the #3588 PR for fixing checkbox display in default theme.

PR #3594, opened 2016-02-26
fix embedded version

scottdraves:
fix runembed mode


scottdraves:
looks like the text test is flaking out...
bring up to date with master to pick up the fix.

scottdraves:
:+1:

Issue #3567, opened 2016-02-22
add to start page


scottdraves:
1) big button for opening a notebook.
2) recent notebooks.

scottdraves:
combine with https://github.com/twosigma/beaker-notebook/issues/3549

scottdraves:
1) the big button should say "Open notebook..." and it should be the same as the open file->open bkr notebook menu item.  the button should be the same size as the "new empty" and "new default" buttons, and placed in the row below them, taking up half of the drag zone button now.  the drag zone button becomes the same size as the other 3 buttons.  so there are 4 buttons in 2 rows of 2, all the same size.

scottdraves:
2) the list of recent notebooks should go below the 4 buttons.  if there are no recent notebooks, then nothing should show at all.  if there are some then the title "Recent Notebooks" appears, followed by a a list in the same style as the "open notebooks" above.  if you click on one, then it opens.  there is no date/time or close button.

PR #3597, opened 2016-02-29
Start page improvements

alice-kroch:
Issue #3567. Please, check.


scottdraves:
if there are no recent notebooks i get the following:
screen shot 2016-02-29 at 11 16 08 am
but it should just not show the "Recent Notebooks" section at all in this case.

scottdraves:
test suite fails.

scottdraves:
:+1:

Issue #3494, opened 2016-02-09
eliminate table options window


scottdraves:
add to column menu Items "Fix Column Left" and "Fix Column Right".
They have checkboxes you can toggle on and off.
Any column to the left of a fixed-left column is also fixed.
Any column to the right of a fixed-right column is also fixed.

move the "pagination" check box to the table menu (an item that has a check you can toggle).
when turning on scrolling (by unchecking hat item), start out with the same number of rows as are currently displayed by pages.

if pagination is off, then we get the scrollbar.  change the number of rows displayed by clicking and dragging on the LR corner of the table, between the scrollbars.  put the drag icon here and change
the mouse hover to match.

if possible, when pagination is enabled and those controls are visible at the bottom of the table, add a "pagination" check box which is checked.  if the user unchecks it, then the controls vanish and we enter scroll bar mode

PR #3600, opened 2016-02-29
eliminate table options window #3494

mnavasiolava:
besides the changes described in #3494 this pr also contains:
1) fixes for header menu (for fixed columns)
2) fixes for column reordering (fixed columns are blocked from reordering)
3) fixes for key bindings and fixed columns


scottdraves:
1) if i reorder the columns, then the menus do not move with them.
2) the "use pagination" menu item should be a regular checkmark, like the "Lock" in the notebook menu.

mnavasiolava:
done

Issue #3540, opened 2016-02-17
undo/redo keys should work in easyform fields


scottdraves:
text fields and areas

PR #3606, opened 2016-03-01
undo/redo keys should work in easyform fields #3540

kpaharelau:
Undo/Redo worked fine for TextArea. Problem was only for a TextField. Solution is very simple :-)


kpaharelau:
kick jenkins

scottdraves:
:+1:

Issue #3550, opened 2016-02-18
bogus logging


scottdraves:
output like this make real errors hard to find and distracts users.
what's the best way to eliminate?
it should be an option (default on) because in some circumstances we want to see everything like this.

```
nginx-stderr>2016/02/18 15:12:43 [error] 62091#0: *48 "/Users/spot/.beaker/v1/nginx2662179024838529286/html/fa06eb6/groovy.726569710906/cometd/index.html" is not found (2: No such file or directory), client: 127.0.0.1, server: , request: "GET /fa06eb6/groovy.726569710906/cometd/ HTTP/1.1", host: "127.0.0.1:8801"
nginx-stderr>2016/02/18 15:12:45 [error] 62091#0: *49 "/Users/spot/.beaker/v1/nginx2662179024838529286/html/7415844/java.040809700761/cometd/index.html" is not found (2: No such file or directory), client: 127.0.0.1, server: , request: "GET /7415844/java.040809700761/cometd/ HTTP/1.1", host: "127.0.0.1:8801"
nginx-stderr>2016/02/18 15:13:14 [error] 62091#0: *50 "/Users/spot/.beaker/v1/nginx2662179024838529286/html/7415844/java.040809700761/cometd/index.html" is not found (2: No such file or directory), client: 127.0.0.1, server: , request: "GET /7415844/java.040809700761/cometd/ HTTP/1.1", host: "127.0.0.1:8801"
```

scottdraves:
seems to be back

scottdraves:
could not reproduce, maybe i switched branches which would confuse this.

PR #3610, opened 2016-03-01
Hide Zombie Logging

maxyeg:
For review.
Not sure about the names. 
And I can't pass 'nginx-bogus-logging' option to beaker(maybe because of windows).

Intended to fix Issue #3550


core/src/main/java/com/twosigma/beaker/core/Main.java
```diff
@@ -96,6 +96,7 @@ private static CommandLine parseCommandLine(String[] args)
     opts.addOption(null, "require-password", false, "Ask for password when connecting");
     opts.addOption(null, "listen-interface", true, "Interface to listen on - requires ip address or '*'");
     opts.addOption(null, "portable", false, "Configuration and runtime files located in application instead of user home directory.");
+    opts.addOption(null, "nginx-bogus-logging", false, "Switchs off logging unsuccessfull requests from outdated beaker pages.");
```
scottdraves: nginx-bogus-logging -> hide-zombie-logging

default should be true.

Switchs off logging unsuccessfull requests from outdated beaker pages.
->
Prevent distracting logging by Beaker clients of previous server instances.

scottdraves:
i tried bypassing the option and just forced that conditional to true, and then opened chartingTutorial.bkr, and it does not load.  i think the rule is catching too much.

core/src/main/java/com/twosigma/beaker/core/rest/PluginServiceLocatorRest.java
```diff
@@ -727,6 +737,12 @@ else if (nginxRule.equals("ipython2"))
     }
     nginxConfig = nginxConfig.replace("%(plugin_section)s", pluginSection.toString());
     nginxConfig = nginxConfig.replace("%(extra_rules)s", this.nginxExtraRules);
+    if (this.hideZombieLogging) {
+      nginxConfig = nginxConfig.replace("%(catch_outdated_requests_rule)s",
+          this.CATCH_OUTDATED_REQUESTS_RULE);
+    } else {
+      nginxConfig = nginxConfig.replace("%(catch_outdated_requests_rule)s", "");
+    }
```
scottdraves: better style:

nginxConfig = nginxConfig.replace("%(catch_outdated_requests_rule)s", this.hideZombieLogging ? this.CATCH_OUTDATED_REQUESTS_RULE : "");

scottdraves:
OK the rule is good. :)

but it should be enabled by default.  so the cmd line flag should be --show-zombie-logging to disable this feature.

maxyeg:
I've passed "show-zombie-logging" and it was treated by parser as an argument, but not an option. So hasOption was always false.
I've debugged GnuParser a bit and it expects options to be prefixed with '-' or '--'. In both cases it will be treated as option. And so hasOption will work.

scottdraves:
:+1:

Issue #3591, opened 2016-02-25
improve filtering and searching


scottdraves:
after playing with the filters for a while i think they could still use some refinement, to make them more powerful and simpler.

1) there be two menu items, replace Filter with Filter and Search (this goes for columns and the table-wide menu)
2) searching always does substring matching.
3) filtering calls eval() on an expression from the field in an environment where there is a variable defined for each column.  in column filters, there is also a variable defined "x" with the current column.
4) the left side of the text field should have either a filter icon or a search icon, depending on which it is.
5) only one or the other can be active at a time, so if you have a filter going, and you select the search menu item, the filter is cleared and the icon switches to search.

PR #3611, opened 2016-03-02
improve filtering and searching #3591

mnavasiolava:


scottdraves:
this is great but needs some more work:
1) add the search and filter icons to the menu items, on the left.
2) when filtering, i am not seeing the columns bound to variables.  for example, in the mpg table in the R examples tutorial, i should be able to filter on an expression like "cyl>=6&&cty<15".  if a column name is not a legal javascript variable, then it does not get defined.  these should get defined on all filter evals.
3) instead of using "x" as the variable, use "$".
4) there should be a table-wide filter also.  this eval does not have $ defined but $$ and the column names would be.
5) in addtion to $, also define $$ which is a object with fields, one for each column.  this allows you to access columns with names that are not legal field names like $$['has space'].

mnavasiolava:
done

scottdraves:
:+1:

Issue #3532, opened 2016-02-16
add menu to each value in beaker object interface


scottdraves:
to the right of each variable name, put a little drop-down triangle (like next to the langauge menu for cells).
change the color of the text for the variable name to yellow, like an anchor of a link in markdown, but it changes to blue on hover.
when you click to open the menu, the items are "Delete" and "Rename".
Delete just does it.
Rename brings up a modal dialog where they can enter a new name, or cancel.

scottdraves:
partial implementation: https://github.com/twosigma/beaker-notebook/pull/3641

scottdraves:
closed by mistake before, PR is still under development

PR #3641, opened 2016-03-09
Menu in beaker object interface

alice-kroch:
Issue #3532. Please, check.


scottdraves:
1) the triangle for the menu should point down, not right.
2) the display should update after you make a change.
3) i tried delete, and it deleted all the properties, not just the one i selected.
4) i tried rename, and it renamed correctly, but it deleted all the other properties.

scottdraves:
1 and 2 seem fixed, thanks
but 3 and 4 are still a problem...

scottdraves:
3+4 seem fine now.

still needs to focus on text field when the rename dialog opens, and fix color of the menu in default theme (should be green matching the language menu in cells in nonadvanced mode).

kpaharelau:
spot [5:21 PM] 
ah

[5:21] 
these changes for the beaker object should not apply the the output container

[5:21] 
i did not intend for output containers to have menus

[5:21] 
only the beaker object

[5:21] 
i guess it should be another kind of container

scottdraves:
great.  a few remaining details:
1) remove the "beaker object menu" tooltip
2) bring up to date with master
3) the triangle is wrong. it is narrower than the one in the language menu:
screen shot 2016-03-21 at 11 32 25 am
screen shot 2016-03-21 at 11 32 13 am
4) the text does not highlight and the cursor does not change to the finger in a large enough area.  if you point between the triangle and the text, or below the triangle, then it does not activate.  it should be active in the whole area around.
untitled 2

Issue #3632, opened 2016-03-06
command N


scottdraves:
the new notebook keyboard commands on mac are control-n, not command-n.
the documentation & menu items need to be fixed.
and it doesn't work when in a notebook, only on the start page.

PR #3644, opened 2016-03-10
Replace cmd commands for new notebook commands

alice-kroch:
Issue #3632. Please, check.


kpaharelau:
kick jenkins

scottdraves:
OK this looks good but I realized it's also broken on Windows, but it has the opposite problem.
There, control-N just opens a new browser, so we can't use it for new notebook.
On Windows it should be Alt-N and Shift-Alt-N

alice-kroch:
Updated. Please, check.

scottdraves:
:+1:

kpaharelau:
kick jenkins

kpaharelau:
kick jenkins

Issue #3582, opened 2016-02-24
make table columns resizable by dragging


scottdraves:

scottdraves:
https://github.com/Silvacom/colResize ?

scottdraves:
still to do: i can't make a column narrower than the content in it.
if i drag it small, then it should truncate the contents.

PR #3647, opened 2016-03-10
make table columns resizable by dragging #3582

mnavasiolava:
at this point fixed columns resizing is not available


scottdraves:
cool but i can't make a column narrower than the content in it.
if i drag it small, then it should truncate the contents.

scottdraves:
:+1:

Issue #3612, opened 2016-03-02
class loader should be parent-first


scottdraves:

scottdraves:
http://frankkieviet.blogspot.in/2009/03/javalanglinkageerror-loader-constraint.html

KrzysztofKowalczyk:
Hi, I think I hit this issue when trying to use Selenium from beaker. To reproduce you can download selenium, set it on classpath and create new instance of firefox driver. Attaching beaker notebook with that, be aware it downloads stuff and then one need to manually adds its output to Java or Groovy classpath.
[Selenium.bkr.zip](https://github.com/twosigma/beaker-notebook/files/163235/Selenium.bkr.zip)

The actuall error in my case
java.lang.LinkageError: loader constraint violation: when resolving overridden method "com.sun.org.apache.xerces.internal.jaxp.DocumentBuilderImpl.setErrorHandler(Lorg/xml/sax/ErrorHandler;)V" the class loader (instance of com/twosigma/beaker/jvm/classloader/DynamicClassLoaderSimple) of the current class, com/sun/org/apache/xerces/internal/jaxp/DocumentBuilderImpl, and its superclass loader (instance of ), have different Class objects for the type org/xml/sax/ErrorHandler used in the signature

PR #3650, opened 2016-03-10
changed classloader to parent first

ildipo:
should fix issue https://github.com/twosigma/beaker-notebook/issues/3612


scottdraves:
tested running scala notebook with spark jar.
@dkobylarz would this cause any problems for you?

Issue #3651, opened 2016-03-11
table keyboard cmd focus


scottdraves:
if you are editing a code cell, and then hover the mouse over the table in its output, and type a character eg "1", then it both is entered into the code cell, and also it changes the formatting of that column.
it shouldn't go into the table unless you have clicked on it somewhere, and are already interacting with it.
this is part of https://github.com/twosigma/beaker-notebook/issues/3232

PR #3653, opened 2016-03-11
table keyboard cmd focus #3651

mnavasiolava:
table becomes interactive only if you click on its container.

i had to make changes in colReorder plugin because otherwise if we e.g. had focus in code cell and then clicked on a table header focus wasn't removed from the code cell because event was prevented and we got both typing in code cell and table interactions


scottdraves:
:+1:

Issue #3683, opened 2016-03-15
Wrong chart size in Firefox browser


kpaharelau:
![image](https://cloud.githubusercontent.com/assets/13516511/13780492/c3e15f08-ead0-11e5-8ae0-500ebc42624d.png)

PR #3684, opened 2016-03-15
Wrong chart size in Firefox browser #3683

kpaharelau:
use simple width() and height() functions in a firefox browser.

### DIFF
--- a/core/src/main/web/app/helpers/globals.js
+++ b/core/src/main/web/app/helpers/globals.js
@@ -50,7 +50,22 @@
       THEMES: {
         DEFAULT: 'default',
         AMBIANCE: 'ambiance'
-      }
+      },
+      //see http://stackoverflow.com/questions/9847580/how-to-detect-safari-chrome-ie-firefox-and-opera-browser
+      // Opera 8.0+
+      IS_OPERA: (!!window.opr && !!opr.addons) || !!window.opera || navigator.userAgent.indexOf(' OPR/') >= 0,
+      // Firefox 1.0+
+      IS_FIREFOX: typeof InstallTrigger !== 'undefined',
+      // At least Safari 3+: "[object HTMLElementConstructor]"
+      IS_SAFARI: Object.prototype.toString.call(window.HTMLElement).indexOf('Constructor') > 0,
+      // Internet Explorer 6-11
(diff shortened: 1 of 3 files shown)


core/src/main/web/outputdisplay/bko-plot/plotutils.js
```diff
@@ -54,11 +54,24 @@
         222	: "SINGLE_QUOTE"
       };
     return {
+
+      isFirefox: function ()
```
scottdraves: these should be application-wide helper functions, not just here.
in fact i am surprised we don't have something like this already, please check.

core/src/main/web/outputdisplay/bko-plot/plotutils.js
```diff
@@ -54,11 +54,24 @@
...
+      {
+        return typeof InstallTrigger !== 'undefined';
+      },
+      isSafari: function ()
```
scottdraves: should be formatted

```
function () {
 return ...
}
```

core/src/main/web/app/helpers/globals.js
```diff
@@ -50,7 +50,22 @@
       THEMES: {
         DEFAULT: 'default',
         AMBIANCE: 'ambiance'
-      }
+      },
+      //see http://stackoverflow.com/questions/9847580/how-to-detect-safari-chrome-ie-firefox-and-opera-browser
```
scottdraves: should go in helper.js, global.js is just for constants not for functions.

also, please only define firefox, safari, and chrome as we support only them.

core/src/main/web/app/helpers/helper.js
```diff
@@ -73,6 +73,14 @@
 
     var bkHelper = {
 
+      //see http://stackoverflow.com/questions/9847580/how-to-detect-safari-chrome-ie-firefox-and-opera-browser
+      // Firefox 1.0+
+      isFirefox: typeof InstallTrigger !== 'undefined',
+      // At least Safari 3+: "[object HTMLElementConstructor]"
+      isSafary: Object.prototype.toString.call(window.HTMLElement).indexOf('Constructor') > 0,
```
scottdraves: Safari not Safary

PR #3698, opened 2016-03-17
Resizable columns. Truncate cell text. #3582

mnavasiolava:


scottdraves:
doesn't work for me -- no change in behavior from master.

mnavasiolava:
@scottdraves which table you've tried?
i tried on beaker.rates data and it works for me
image

Issue #3658, opened 2016-03-11
disconnected dialogs still stack up


scottdraves:
if one is open, then we don't need more.

PR #3700, opened 2016-03-17
#3658. disconnected dialogs still stack up

maxyeg:


scottdraves:
kick jenkins

core/src/main/web/app/mainapp/mainapp.js
```diff
@@ -1160,34 +1160,30 @@
...
+            "Disconnected", function() {
+              // "Save", save the notebook as a file on the client side
+              bkSessionManager.dumpDisplayStatus();
+              var timeoutPromise = $timeout(function() {
+                bkUtils.saveAsClientFile(
+                    bkSessionManager.getSaveData().notebookModelAsString,
+                "notebook.bkr");
+              }, 1);
+              timeoutPromise.then(function() {
+                $scope.disconnectedDialog = null;
```
tomca32: Can we not use null? JavaScript's null really shouldn't exist and is super weird. For example try doing this in the console:

```
typeof null; // it's "object"
```

The best way to handle this is to set that value to undefined but without using the literal "undefined" since that's just a variable name. This is done by using void:

```
$scope.disconnectedDialog = void 0;
```

This will give us 100% actual undefined primitive. 

http://stackoverflow.com/questions/1291942/what-does-javascriptvoid0-mean

core/src/main/web/app/mainapp/mainapp.js
```diff
@@ -1160,34 +1160,30 @@
...
+              })
+            }, function() {
+                $scope.disconnectedDialog = null;
```
tomca32: Same here

scottdraves:
this is the kind of problem that typescript would avoid right?

tomca32:
I know typescript had proposals for @NotNullable but I don't know it they implemented it or not. If not, typescript wouldn't help us here. We would still have nulls.

The problem with nulls in JS is that we already have undefined. So if you want to null check you have to check for null and undefined. This is easy to forget and often leads to bugs. Easier to just avoid using null altogether.

There is a counter argument to this. When dealing with object properties, like this:

```
obj = {
  a: void 0
}
obj.a; // this gives us undefined
obj.b; // this is also undefined
```

So in both cases we get undefined. Even though one property exists and the other one doesn't. An argument could be made here that if you want to be sure the property exists but it's value is unknown, we should use null instead of undefined. I still think it's not worth it for all the confusion and additional null checks you'll need to have later in the code.

tomca32:
kick jenkins

scottdraves:
kick jenkins

Issue #3695, opened 2016-03-17
Cursor position is unintuitive when switching cells


dela3499:
To reproduce: 
1. Create two multi-line cells
2. Click somewhere in the middle of the text on the second cell
3. Click somewhere on the first cell
4. Move the cursor down with keyboard arrows until you enter second cell
5. Now cursor is not at beginning of text in second cell, it is in the last place

It becomes a bit weird when I'm editing one cell, use the arrow keys to move down, accidentally enter a new cell, and then I have to press up arrow 10 times, say, to get back to the other cell.

scottdraves:
thanks for the report.  agreed.  when you use the down-arrow key from the last line of one code cell, and it moves to the next code cell, it should move to the first line.

ditto for moving up.

PR #3701, opened 2016-03-17
#3695. Cursor position is unintuitive when switching cells

sshkalikov:


core/src/main/web/app/helpers/core.js
```diff
@@ -502,7 +502,16 @@
             return;
           }
           if (cm.getCursor().line === 0) {
-            moveFocusUp();
+            var prevCell = moveFocusUp();
+            if(prevCell) {
```
scottdraves, line 506: style guide says there should be a space like `if (`

core/src/main/web/app/helpers/core.js
```diff
@@ -517,7 +526,16 @@
             return;
           }
           if (cm.getCursor().line === cm.doc.size - 1) {
-            moveFocusDown();
+            var nextCell = moveFocusDown();
+            if(nextCell) {
```
scottdraves, line 530: ditto

core/src/main/web/app/helpers/core.js
```diff
@@ -517,7 +526,16 @@
...
+              var nextCm = scope.bkNotebook.getCM(nextCell.id);
+              if (nextCm){
```
scottdraves, line 532: similarly should be `) {`

scottdraves:
works great but a few style fixes

PR #3718, opened 2016-03-18
Separated value from label in language manager radio buttons and chan…

tomca32:
…ged js language version format in bkr files

UI will now display human readable version names such as: "ECMAScript 2015" while in the .bkr files it will be "ES2015"

### DIFF
--- a/core/config/tutorials/chartingTutorial.bkr
+++ b/core/config/tutorials/chartingTutorial.bkr
@@ -30,7 +30,7 @@
                     "background": "#FFE0F0"
                 }
             },
-            "languageVersion": "ECMAScript 2015"
+            "languageVersion": "ES2015"
         },
         {
             "name": "Groovy",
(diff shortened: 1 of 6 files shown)


scottdraves:
```
$ grep ECMAScript core/config/tutorials/*
core/config/tutorials/chartingTutorial.bkr:            "languageVersion": "ECMAScript 2015"
core/config/tutorials/easyform.bkr:            "languageVersion": "ECMAScript 2015"
core/config/tutorials/output_container_cell.bkr:            "languageVersion": "ECMAScript 2015"
core/config/tutorials/sql-examples.bkr:            "languageVersion": "ECMAScript 2015"
```

tomca32:
kick jenkins

tomca32:
Updated. All the tutorials were set to ES2015.

Also, now the plugin will use a proper default version even when the information in loaded .bkr files is invalid.

Issue #3720, opened 2016-03-18
define reset kernel key cmd


scottdraves:
alt-R on pc, control-R on mac should
1) reset each kernel that has a reset cmd and then
2) run all init cells

the trick here is identifying the "reset" command, since it has different names in different languages.
we should probably normalize it in the UI and that will make it easy for this code to find as well.

scottdraves:
this was closed by mistake before

PR #3735, opened 2016-03-21
#3720. define reset kernel key cmd

sshkalikov:


scottdraves:
with this test notebook: https://pub.beakernotebook.com/#/publications/56f05fac-0616-4724-bc98-234918a1051a
if i hit control-R i get this weird error:
screen shot 2016-03-21 at 4 53 52 pm

scottdraves:
also, bring up to date with master

scottdraves:
kick jenkins

scottdraves:
great, this is much better.  but still, if i try control-R on the charting tutorial, it seems every other time i get an error from one of the init cells.

scottdraves:
and that reminds me, if one of the init cells does fail, the notebook should scroll to show it.
i think this happens during the normal init cell process.

Issue #3705, opened 2016-03-17
sometimes all tests fail


scottdraves:
eg http://ec2-54-175-192-115.compute-1.amazonaws.com:8080/job/Beaker%20PR/624/consoleFull

```
[firefox #1-1]   beaker landing load
[firefox #1-1]     ✗ can load
[firefox #1-1]       - Failed: Angular could not be found on the page http://localhost:8801/ : retries looking for angular exceeded
```

scottdraves:
:dolphin:

PR #3738, opened 2016-03-21
fix sometimes all test failed #3705

EfimovVladimir:


test/tests/autotranslation.js
```diff
@@ -18,11 +18,16 @@ var BeakerPageObject = require('./beaker.po.js');
...
     jasmine.DEFAULT_TIMEOUT_INTERVAL = 120000; // Slow initialization cells, CI box might need a long time to initialize these
     beakerPO = new BeakerPageObject();
-    browser.get(beakerPO.baseURL);
-    browser.waitForAngular();
+    browser.get(beakerPO.baseURL).then(
+        function(){done();},
+        function(reason){ console.log('error occurred while browser load page : ' + reason);}
+    );
+    browser.waitForAngular().then(function() {
+      done();
```
tomca32: So what happens when we call done() the second time? I thought the way this worked is that when the done is called, the tests will start. Seems a bit weird to call done() twice.
I thought that we want to get the baseURL 'then' wait for angular and 'then' call the done() callback:

```
browser.get(beakerPO.baseURL)
  .then(browser.waitForAngular)
  .then(done);
```

test/tests/category-plot-tutorial.js
```diff
@@ -21,11 +21,15 @@ var beakerPO;
...
     beakerPO = new BeakerPageObject();
-    browser.get(beakerPO.baseURL + "beaker/#/open?uri=file:config%2Ftutorials%2FCategoryPlot.bkr&readOnly=true");
-
-    browser.waitForAngular();
+    browser.get(beakerPO.baseURL + "beaker/#/open?uri=file:config%2Ftutorials%2FCategoryPlot.bkr&readOnly=true").then(
+        function(){done();},
+        function(reason){ console.log('error occurred while browser load page : ' + reason);}
+    );
+    browser.waitForAngular().then(function() {
+      done();
```
tomca32: No need to wrap the done() call in a function. We can just reference 'done' directly:

```
browser.waitForAngular().then(done);
```

test/tests/autotranslation.js
EfimovVladimir: Anyway Protractor automatically applies browser.waitForAngular() before every WebDriver action.
And sometimes we get exeption "Angular could not be found on the page" when execute  browser.get(beakerPO.baseURL). And after that error Protractor fail all tests.
I give second chance to load Angular.
I can write like this:
browser.get(beakerPO.baseURL)
.then(done, browser.waitForAngular())
.then(done);
Or/And we can set time to wait for Angular to start for browser.get(url, opt_timeout). Default 10 sec. For example we can set 20 sec.
What do you think?

tomca32: Didn't know this that Protractor does waitForAngular automatically, thanks.

If that's the case, can we just remove the waitForAngular call completely? I don't believe in second chances in the code. I would hope that just making that beforeEach block async would help us to avoid this error.

I wouldn't go with the timeouts, we already have far too many of those. If you think that we still need the "second chance", I would definitely go with the second example of that promise code:

``` javascript
browser.get(beakerPO.baseURL)
  .then(done, browser.waitForAngular)
  .then(done);
```

But I would much rather first try it without waitForAngular and just let the beforeEach wait on browser.get to complete.

test/tests/autotranslation.js
```diff
@@ -18,16 +18,10 @@ var BeakerPageObject = require('./beaker.po.js');
...
     jasmine.DEFAULT_TIMEOUT_INTERVAL = 120000; // Slow initialization cells, CI box might need a long time to initialize these
     beakerPO = new BeakerPageObject();
-    browser.get(beakerPO.baseURL).then(
-        function(){done();},
-        function(reason){ console.log('error occurred while browser load page : ' + reason);}
-    );
-    browser.waitForAngular().then(function() {
-      done();
-    });
+    browser.get(beakerPO.baseURL);
```
tomca32: I still think we should try doing it async, without it we're pretty much back where we started. So if you put the done callback in and add this, we should be good:

``` javascript
browser.get(beakerPO.baseURL).then(done);
```

Also, maybe there's a way to write that beforeEach only once? Tests are running in a node environment so it should be easy to isolate it in a module and import.

Issue #3740, opened 2016-03-21
If the cell outputs "false" the output is hidden


tomca32:
![screen shot 2016-03-21 at 2 01 33 pm](https://cloud.githubusercontent.com/assets/3384784/13928773/76d0469c-ef6d-11e5-9e8f-bd35b516ca50.png)
![screen shot 2016-03-21 at 2 02 05 pm](https://cloud.githubusercontent.com/assets/3384784/13928783/82dc69de-ef6d-11e5-9587-3925dbc6e5c5.png)

tomca32:
Working as intended

PR #3741, opened 2016-03-21
#3740 - Accounting for cell output being 'false'

tomca32:


tomca32:
#3740 - that is working as intended. Closing

PR #3748, opened 2016-03-22
add menu to each value in beaker object interface #3532

kpaharelau:
great. a few remaining details:
1) remove the "beaker object menu" tooltip
2) bring up to date with master
3) the triangle is wrong. it is narrower than the one in the language menu:
4) the text does not highlight and the cursor does not change to the finger in a large enough area. if you point between the triangle and the text, or below the triangle, then it does not activate. it should be active in the whole area around.

DONE


scottdraves:
:+1:

Issue #3671, opened 2016-03-14
tables flash "Processing..."


scottdraves:
when i click on the column headers to change the sort order,
i see "Processing..." flashing in the middle of the cell.

it should only show that if the delay is greater than 0.5s

PR #3750, opened 2016-03-22
tables flash "Processing..." #3671

mnavasiolava:
add Processing indicator only if table has more than 500 items


scottdraves:
:+1:

PR #3765, opened 2016-03-23
output container hidden tabs with table not formatted correctly for h…

kpaharelau:
I have fixed problem but found very strange issue.
Some time ago the code for a table redraw worked fine

 _.defer(function(){ scope.table.draw(false);});

 It doesn't work for unhided cell now.
1. Open 'Output Containers and Layout Managers' tutorial
2. Hide output cell for Stacked example
3. Evaluate cell
4. Unhide output cell

I found only one way to fix it

$timeout(function () {
              _.defer(function(){ scope.table.draw(false);});
              tableChanged = false;
            }, 0);

I ready for any new ideas


scottdraves:
even if the output is open the whole time i don't see the body of the table:

screen shot 2016-03-23 at 1 54 02 pm

Issue #3760, opened 2016-03-22
java errors garbled


scottdraves:
https://pub.beakernotebook.com/#/publications/56f1a654-ed98-4714-85b3-2ab96fcac9a1

open the error to see:

```
ception {
syntax}
}
';' expected
n {
syntax}
}
```

other errors are also broken eg https://pub.beakernotebook.com/#/publications/56f1acb9-9532-436d-bd91-518554941451

```
java.lang.IllegalStateException: method getDateTxt() is already defined in class test.beaker.BeakerTest
ng();  }  public String getDateTxt() {    return _date.toString();  }  public S
```

sshkalikov:
Compilation error messages will be better now. 
But they may be **much** better if we won't remove all line breaks in code (now we're compiling very long line of code (here's an example https://i.gyazo.com/d5558b3429023cfdf957780d518e8074.png))

PR #3775, opened 2016-03-23
#3760. java errors garbled

sshkalikov:


scottdraves:
OK good it's not garbled, that's fixed.  but
it's missing the 1st line with the most important info.
in the syntax error case, the 1st line has the line number and then
"error: class, interface, or enum expected" or "error: <identifier> expected"

also, the new source file needs a copyright/license banner.

also, for the name of the new file, just call it "JavaSourceCompiler.java".  "Beaker" is redundant.

scottdraves:
:+1:

Issue #3795, opened 2016-03-24
In some cases Run button doesn't work


mnavasiolava:
- open charting tutorial
- expand code cell for Custom Plot Example
- set cursor to the 1st line
- scroll down so Run button becomes visible
- click Run

**result:**
cursor was set to the last line
the code wasn't executed

scottdraves:
i get the cursor going to the last line which is a bug, but the code is executed fine.
clicking in the bkg of the code cell below the cm should move the cursor to the end.
i think this is misfiring if user clicks on run button.

PR #3796, opened 2016-03-24
In some cases Run button doesn't work #3795

kpaharelau:
use setSelection function instead goDocStart

### DIFF
--- a/core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
+++ b/core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
@@ -417,7 +417,8 @@
               //codecomplete is up, skip
               return;
             }
-            scope.cm.execCommand('goDocStart');
+
+            scope.cm.setSelection({line: 0, ch: 0 }, {line: 0, ch: 0 }, {scroll: false});
           });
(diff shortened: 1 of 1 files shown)


core/src/main/web/app/mainapp/components/notebook/codecell-directive.js
```diff
@@ -417,7 +417,8 @@
               //codecomplete is up, skip
               return;
             }
-            scope.cm.execCommand('goDocStart');
+
+            scope.cm.setSelection({line: 0, ch: 0 }, {line: 0, ch: 0 }, {scroll: false});
```
tomca32, line 421: Why does the cursor position affect the run button?

kpaharelau, line 421: This command  scope.cm.execCommand('goDocStart'); calls  scope.cm.setSelection. See documentation about this function

doc.setSelection(anchor: {line, ch}, ?head: {line, ch}, ?options: object)
Set a single selection range. anchor and head should be {line, ch} objects. head defaults to anchor when not given. These options are supported:
scroll: boolean
Determines whether the selection head should be scrolled into view. Defaults to true.

We get scrolling on a big code cells and problem with focusing as result

Issue #3785, opened 2016-03-24
add table-of-contents command


scottdraves:
in the notebook menu, use an hrule to make a section of the existing commands, and add "show table of contents"

show table of contents
collapse all sections
open all sections

the ToC view opens all sections except closes the leaves.

scottdraves:
just doing it with sections doesn't work (see https://github.com/twosigma/beaker-notebook/pull/3807).
need something stronger, maybe some kind of floating element with just the section tree.

PR #3807, opened 2016-03-25
show table of contents menu #3785

mnavasiolava:


scottdraves:
a "leaf" section is one that does not contain a section, not one that is "level 1". for example the ToC for the charting tutorial should be like:
screen shot 2016-03-30 at 12 34 37 am

mnavasiolava:
ok, if we should check if a section contains child sections i get the following result for the charting tutorial

image

sections 1 and 2 are not collapsed because they have subsections

scottdraves:
you are right this isn't working out as intended. canceling it, sorry.

Issue #3710, opened 2016-03-18
filter field layered over index columns


scottdraves:
in a wide table with a scrollbar, open the filters and start to write in it, then scroll until until it crosses the index&fixed columns and goes over them instead of under.

scottdraves:
<img width="465" alt="screen shot 2016-03-17 at 10 18 48 pm" src="https://cloud.githubusercontent.com/assets/963093/13866713/af5c323c-ec8e-11e5-94d0-c449a83c0245.png">

PR #3817, opened 2016-03-28
filter field layered over index columns #3710

mnavasiolava:
filter input is truncated and fit into the column if table is scrolled horizontally


scottdraves:
:+1:

Issue #3707, opened 2016-03-18
addevalplugins.js should generated


scottdraves:
during build core/src/main/web/plugin/init/addevalplugins.js

PR #3825, opened 2016-03-30
addevalplugins.js should generated #3707

kpaharelau:
1. task 'addevalplugins' has been added to the core project
2. copyVendorStuff.dependsOn addevalplugins
3. use gradle realclean for deletion generated file


scottdraves:
:+1: but should wait for release

Issue #3833, opened 2016-03-31
Unnecessary horizontal scroll in table


mnavasiolava:
in python2&3 tutorial
![image](https://cloud.githubusercontent.com/assets/12935591/14173859/4774184a-f749-11e5-8de7-cbf705f994c4.png)

PR #3834, opened 2016-03-31
Unnecessary horizontal scroll in table #3833

mnavasiolava:


scottdraves:
good catch, thanks :+1:

Issue #3832, opened 2016-03-30
table missing controls


scottdraves:
https://pub.beakernotebook.com/#/publications/56fc4ef2-003d-4191-a06f-e2b0a8302343

you can reproduce this notebook with the python2&3 tutorial or just download the above to view.
the problem is the pagination controls are missing, ditto for the scrollbar.  but if you click on the
value column, or you filter, you can see there are many more rows.
<img width="318" alt="screen shot 2016-03-30 at 6 10 10 pm" src="https://cloud.githubusercontent.com/assets/963093/14159244/c0e3521a-f6a2-11e5-9bde-bca5d37751ae.png">

mnavasiolava:
we show pagination controls only if there are more than 25 items in a table. otherwise we fix table height to 350px and add scroll. this table has 23 items and there is a scroll

mnavasiolava:
about scrollbar. if you use chrome on mac and you have property 'Show scroll bars' set to 'Automatically based on mouse or trackpad' you may not see scrollbar thumb until you start scrolling. but in any case you are able to scroll the table

scottdraves:
then please change the limit from 25 to 15 so you can either see all the elements, or the regular nav UI shows.

scottdraves:
wait cancel that, let me experiment with this....

scottdraves:
yea we should not have special behavior, if it exceeds the height then the full controls should appear.

PR #3861, opened 2016-04-04
table missing controls #3832

mnavasiolava:
the limit of items to show pagination controls was changed to 15. and they are shown without a scrollbar

### DIFF
--- a/core/src/main/web/outputdisplay/bko-tabledisplay/bko-tabledisplay.js
+++ b/core/src/main/web/outputdisplay/bko-tabledisplay/bko-tabledisplay.js
@@ -278,7 +278,7 @@
     var CELL_TYPE = 'bko-tabledisplay';
     var ROW_HEIGHT = 27;
     var ROW_HEIGHT_ADVANCED_MODE = 22;
-    var MIN_ROWS_FOR_PAGING = 15;
+    var MIN_ROWS_FOR_PAGING = 25;
     return {
       template: JST['bko-tabledisplay/output-table'],
       controller: function($scope, $uibModal) {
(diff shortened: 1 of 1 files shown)


tomca32:
Could we maybe put these values in globals? As they stand now they are pretty much magic numbers and it's not obvious why they have those values.

scottdraves:
disagree that they should be globals since they are only used in this one place.
but a comment that describes their relationship would be good.
plus, does doesn't this assume 30 pixels per row?  in advanced mode that's not true anymore...

tomca32:
Fair point about globals. How about constants defined at the beginning of the file? Or a comment describing the relationship, I like to avoid comments if we can simply name the variable something like 
MIN_ITEMS_FOR_PAGINATION

mnavasiolava:
added constants for number of rows for paging and row height
also took into account different row heights in normal and advanced modes

scottdraves:
kick jenkins

core/src/main/web/outputdisplay/bko-tabledisplay/bko-tabledisplay.js
```diff
@@ -1589,20 +1592,20 @@
           };
 
           var domCommon = '<"bko-table"Z' + (scope.data.length > 500 ? 'r' : '') + 't';
+          var rowHeight = bkHelper.getBkNotebookViewModel().isAdvancedMode() ? ROW_HEIGHT_ADVANCED_MODE : ROW_HEIGHT;
           if (!scope.pagination.use) {
             init.paging = false;
-            init.scrollY = scope.pagination.rowsToDisplay * 27 + 2;
+            init.scrollY = scope.pagination.rowsToDisplay * rowHeight + 2;
```
tomca32, line 1598: did you mean ROW_HEIGHT here?

scottdraves:
i am seeing the controls even though there are only 23 rows, and one page.

core/src/main/web/outputdisplay/bko-tabledisplay/bko-tabledisplay.js
mnavasiolava, line 1598: no, i meant rowHeight which is different for advanced and normal mode

tomca32, line 1598: Oh, I see. Cool, thx

mnavasiolava:
@scottdraves but you asked to change the limit of rows to show pagination controls to 15 or there is smth i misunderstood?

scottdraves:
i meant it should do the obvious thing:
if there is 1 page then do not show pagination controls.
if there is > 1 page then do show.

Issue #3857, opened 2016-04-03
text cell stuck in TeX mode  


Chris1221:
I'm having some unexpected results when trying to render some TeX inline. For example, the following code

``` tex
For a population of $n$ oscillators $\Omega$, with members $N_i, i \in N, 1 \leq i \leq n$. If $x, x\subset\Omega$ projects onto $N_i$ then the gap junctional current to $N_i$ is $\sum_{x} G_{x,i} H( 0.3 - V_x)$
```

Renders as follows:

![image](https://cloud.githubusercontent.com/assets/8516981/14230129/94144d92-f919-11e5-9873-fc4c649079c8.png)

This has been done on the most recent version of Beaker on OSX 10.10.4. The expression renders correctly in LaTeX and MathJax, though I cannot find a way to validate it in KaTex.

![image](https://cloud.githubusercontent.com/assets/8516981/14230137/0427f480-f91a-11e5-9fff-8b511a890bb4.png)

Any assistance would be appreciated.

scottdraves:
thanks, confirmed, we'll look into this.  seems like a bug.

scottdraves:
https://pub.beakernotebook.com/#/publications/57007c03-503e-4732-8e9a-1949ecf6c86c

Chris1221:
Thanks!

Chris1221:
This is probably a related issue but the following is also causing issues:

``` tex
For cell $i$ of $n$, define pro-necrotic burden as $\nu_i$, with total burden of the system equal to $\frac{1}{n} \sum^n_{i=1} \nu_i$. Note that $\nu_{1}, ..., \nu_{n} \in [0,1]$ and is a monotonically increasing function with respect to time. 
```

renders as 

![image](https://cloud.githubusercontent.com/assets/8516981/14232718/3b724f50-f981-11e5-8bbf-228023bf6233.png)

Though it renders correctly in MathJax:

![image](https://cloud.githubusercontent.com/assets/8516981/14232720/4b114308-f981-11e5-8164-37ecbf2fe4ec.png)

Same specs as above.

Chris1221:
And another one:

``` tex
[...]

PR #3862, opened 2016-04-04
text cell stuck in TeX mode #3857

kpaharelau:
only text block scanned by markdown

use for a testing
https://www.dropbox.com/s/i9of3t051hb5bd9/3857.bkr?dl=0


core/src/main/web/app/helpers/helper.js
```diff
@@ -547,13 +547,62 @@
...
-            renderer: bkRenderer
-          });
-          markupDeferred.resolve(result);
+
+          function traverseChildNodes(node) {
+            var next;
+
+            var toDOM=function(str){
+              var d=document
+                ,i
```
scottdraves: style requires var on every tline.

scottdraves:
works on that example well but try it on the text tutorial, look for these problems:
1) in the 1st cell i see a stray "<"
2) the text now flows into the area under the buttons in the top-right of the cell.
3)  the $e^{i\pi}+1=0$ example is being put on a line by itself.
4) &dagger;  should be written out once and rendered once, not rendered twice...

scottdraves:
screen shot 2016-04-04 at 9 59 39 pm
screen shot 2016-04-04 at 9 59 52 pm

vs

screen shot 2016-04-04 at 10 07 49 pm
screen shot 2016-04-04 at 10 07 55 pm

scottdraves:
this works great but:
1) if we are replacing markd, then replace it everywhere and remove it from bower.json & our src tree, and index_template.html
2) exclude the lib subdir of markdown-it

scottdraves:
:+1:

scottdraves:
holding for a couple of days

PR #3865, opened 2016-04-04
Tom/3837 - Added option for autocompleting parameters

tomca32:
in beaker.pref.json

"autocomplete-parameters": ["true"|"false"]

Defaults to "true", set to "false" to stop autocompleting parameters (normal ipython behavior)

Also refactored various calls to getPreference into a getBeakerPreference on bkUtils.


tomca32:
kick jenkins

Issue #3840, opened 2016-03-31
cmd click on recent notebook


scottdraves:
should open it in a new window.  current behavior is to keep using the current window.
this is the list of recent notebooks on the start page.

PR #3875, opened 2016-04-05
cmd click on recent notebook #3840

kpaharelau:
Use window.open for recent items from control page

### DIFF
--- a/core/src/main/web/app/helpers/helper.js
+++ b/core/src/main/web/app/helpers/helper.js
@@ -30,7 +30,7 @@
    *   plugins dynamically
    * - it mostly should just be a subset of bkUtil
    */
-  module.factory('bkHelper', function(bkUtils, bkCoreManager, bkDebug, bkElectron, bkPublicationAuth, GLOBALS) {
+  module.factory('bkHelper', function($location, bkUtils, bkCoreManager, bkDebug, bkElectron, bkPublicationAuth, GLOBALS) {
     var getCurrentApp = function() {
       return bkCoreManager.getBkApp();
     };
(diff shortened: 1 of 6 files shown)


core/src/main/web/app/helpers/helper.js
```diff
@@ -205,6 +205,28 @@
       getBaseUrl: function() {
         return bkUtils.getBaseUrl();
       },
+      openNotebookInNewWindow: function (notebookUri, uriType, readOnly, format) {
```
scottdraves: this can't be right we need a clean way.

scottdraves:
code is clean now, but the new window functionality should only apply if the cmd key is down (ctrl on pc), just like for normal links.

core/src/main/web/app/helpers/helper.js
```diff
@@ -205,6 +205,22 @@
...
+
+        var path = "/open?uri=" + window.encodeURIComponent(notebookUri);
+        if (uriType) {
+          path += "&type=" + uriType;
```
tomca32: No need to do this manually, you can use $httpParamSerializer.

https://docs.angularjs.org/api/ng/service/$httpParamSerializer

Docs are not great, here's an SO example:

http://stackoverflow.com/a/34039330/2341467

core/src/main/web/app/helpers/helper.js
```diff
@@ -205,6 +205,22 @@
       getBaseUrl: function() {
         return bkUtils.getBaseUrl();
       },
+      openNotebookInNewWindow: function (notebookUri, uriType, readOnly, format) {
+
+        var path = "/open?uri=" + window.encodeURIComponent(notebookUri);
```
tomca32: Since $httpParamSerializer will do the encoding, you can use it for uri param also.

scottdraves:
:+1:

scottdraves:
kick jenkins

Issue #3904, opened 2016-04-08
update publish dialog


scottdraves:
remove "unlisted" checkbox, instead show a notice like this:
<img width="343" alt="screen shot 2016-04-07 at 9 28 44 pm" src="https://cloud.githubusercontent.com/assets/963093/14371714/6c5b0b72-fd08-11e5-927f-5249188dcfcd.png">

PR #3909, opened 2016-04-08
update publish dialog #3904

mnavasiolava:
the checkbox was removed and new label was added instead
but i have 2 questions:
1. do we still have to send 'unlisted' property when category is not selected or it will be up to pub server to determine that publication is private? at this point when i don't send 'unlisted' and don't select a category my publication is searchable
2. do we have to hide 'unlisted' label if the category is selected.

### DIFF
--- a/core/src/main/web/app/mainapp/components/publication/publication.js
+++ b/core/src/main/web/app/mainapp/components/publication/publication.js
@@ -153,9 +153,17 @@
         $scope.model.contents = bkHelper.sanitizeNotebookModel(notebook);
       }
 
+      function getPublicationModel() {
+        var pubModel = _.clone($scope.model);
+        if (_.isEmpty(pubModel.name)) {
+          pubModel.name = ' '; //to avoid error on pub server
+        }
+        return pubModel;
+      }
+
       function createPublication() {
         setModelContents();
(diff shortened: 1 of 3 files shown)


tomca32:
kick jenkins

scottdraves:
i can answer 2: never show the "unlisted" checkbox.
@ildipo can answer 1 for sure, but assume that the server determines it.

mnavasiolava:
@scottdraves i meant new label ('If you do not select a category your publication will be 'unlisted''), not checkbox. should it always be visible?

scottdraves:
it should only be visible if the category has not been set.

scottdraves:
confirmed: unlisted flag should be removed from the REST call too.

mnavasiolava:
ok, it is not sent now

scottdraves:
if category is not set then allow blank title
screen shot 2016-04-11 at 11 49 13 am

mnavasiolava:
done.
please note that now pub server doesn't allow blank title. i get `422 Unprocessable Entity: {name: ["name must be present"]}` response

scottdraves:
just send a space to avoid the error

mnavasiolava:
done

scottdraves:
:+1:

Issue #3881, opened 2016-04-05
test new tables and plots for memory leaks


scottdraves:
add tables and plots to the memory tests,
run the memory tests, inspect the code.

scottdraves:
https://github.com/twosigma/beaker-notebook/issues/3925

PR #3920, opened 2016-04-11
test new tables and plots for memory leaks #3881

kpaharelau:
memory tests has been changed for testing tables and plots


kpaharelau:
kick jenkins

kpaharelau:
kick jenkins

PR #3921, opened 2016-04-11
reduce run time category-plot-tutorial spec

EfimovVladimir:


test/tests/beaker.po.js
```diff
@@ -415,6 +415,7 @@ var BeakerPageObject = function() {
     if (!containerIdx)
       containerIdx = 0;
     this.scrollToCodeCellOutput(codeCellOutputIdx);
+    browser.sleep(1000);
```
scottdraves: what are these for?

EfimovVladimir: wait for loading cell output.
I can change it on browser.wait()

scottdraves: yea please use that.

PR #3926, opened 2016-04-11
run tests for tutorial(clojure, java, python, sql) without restart browser

EfimovVladimir:


scottdraves:
bring up to date with master please (and keep python out).

EfimovVladimir:
done

Issue #3839, opened 2016-03-31
reset languages of notebook with python2&3 fails


scottdraves:
create an empty notebook, add IPython, add Python3, then reset.
application hangs with the following on the JS console:
http://127.0.0.1:8801/fe1f0d9/ipython.805349462604/api/kernels/a69985c1-d435-4c35-92b3-2a0d09ab9816/restart Failed to load resource: the server responded with a status of 403 (Forbidden)

Python2 and Python3 seem to work if either is present but not both, perhaps a shared name?

scottdraves:
The same error happens if you reset using the language manager instead of the menu item.

sshkalikov:
In ipython.js and python3.js we have [the code](https://github.com/twosigma/beaker-notebook/blob/a62d06d28c44577759b787271d833c2e1fbc29d8/plugin/ipythonPlugins/src/dist/ipython/ipython.js#L63) that 'expires' old python cookies before starting of plugin.
When we start ipython for example, it sets a cookie with "username-127-0-0-1-8804" (because it starts on port 8804). Then, when we start python3, it kills all cookies with text "username-127-0-0-1-", because it thinks these cookies belongs to him. And when two plugins work, we have only one cookie with text like "username-127-0-0-1-8085" (last port, where some python started)
And seems like this is the reason why we get an error trying to reset **first started python plugin** (last started plugin can be resetted without any problems). Reset command works good without that cookies-killing-code (when we have cookies for each started plugin) but removing of this code don't looks like good approach.

PR #3930, opened 2016-04-12
#3839. reset languages of notebook with python2&3 fails

sshkalikov:


core/src/main/web/app/helpers/helper.js
```diff
@@ -673,6 +673,9 @@
           return [];
         }
       },
+      hasIPythonEvaluatorsLoaded: function () {
+        return _.intersection(this.getEvaluators(), ['IPython', 'IRuby', 'Julia', 'Python3', 'Torch']).length > 0;
```
scottdraves: there has to be a better way...
how about these plugins set a global, then they can just check if the global is set?

sshkalikov:
this way?

scottdraves:
:+1:

Issue #3925, opened 2016-04-11
memory and dom leaks in new tables


scottdraves:
memory test reports:

Heap Size Delta:      9.45 MB
Node Count Delta:     18400
Event Listener Delta: 0

PR #3932, opened 2016-04-12
memory and dom leaks in new tables #3925

mnavasiolava:
test results:

```
Heap Size Delta:      1.83 MB
Node Count Delta:     0
Event Listener Delta: 0
```


scottdraves:
:+1:

Issue #3929, opened 2016-04-12
Shebang at top of ipythonPlugin should be /usr/bin/env python


hxu:
I had an issue where I couldn't start any IPython shells because the Beaker was using the wrong Python interpreter.  Here's the stack trace from the server:

```
[qtp1727361096-20] WARN org.eclipse.jetty.servlet.ServletHandler - /rest/plugin-services/IPython
java.lang.RuntimeException: unable to get IPython hash
    at com.twosigma.beaker.core.rest.PluginServiceLocatorRest.hashIPythonPassword(PluginServiceLocatorRest.java:614)
    at com.twosigma.beaker.core.rest.PluginServiceLocatorRest.generateIPythonConfig(PluginServiceLocatorRest.java:632)
    at com.twosigma.beaker.core.rest.PluginServiceLocatorRest.locatePluginService(PluginServiceLocatorRest.java:432)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:497)
    at com.sun.jersey.spi.container.JavaMethodInvokerFactory$1.invoke(JavaMethodInvokerFactory.java:60)
```

I had set the interpreter in beaker.pref.json, but it wasn't being respected, and I could launch the ipython notebook separately without a problem.

The fix for me was to change the shebang at the top of `config/plugins/eval/ipythonPlugins/ipython/ipythonPlugin` from `/usr/bin/python` to `/usr/bin/env python`, while having the virtualenv that contained ipython active.

Note that I did not use `install-dependencies`, and python was installed into a separate conda environment.

scottdraves:
thanks for the report, i think you are onto something, we'll have to look at that.

scottdraves:
https://github.com/twosigma/beaker-notebook/issues/3927

PR #3939, opened 2016-04-13
Shebang at top of ipythonPlugin should be /usr/bin/env python #3929

kpaharelau:
I think that using env (https://en.wikipedia.org/wiki/Env) command is a good idea. I have changed all start scripts for python plugins

### DIFF
--- a/plugin/ipythonPlugins/src/dist/ipython/ipythonPlugin
+++ b/plugin/ipythonPlugins/src/dist/ipython/ipythonPlugin
@@ -45,4 +45,4 @@ os.environ["PYTHONPATH"] = plugin_dir + old_path
 if os.name == 'nt':
     subprocess.call(["python", plugin_dir+"/ipythonPluginStart.py"], shell=True)
 else:
-    os.execlp("python", "python", plugin_dir+"/ipythonPluginStart.py")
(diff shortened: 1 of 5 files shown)


scottdraves:
didn't they mean this? https://github.com/twosigma/beaker-notebook/blob/master/plugin/ipythonPlugins/src/dist/ipython/ipythonPlugin#L1

plugin/ipythonPlugins/src/dist/ipython/ipythonPlugin
```diff
@@ -45,4 +45,4 @@ os.environ["PYTHONPATH"] = plugin_dir + old_path
 if os.name == 'nt':
     subprocess.call(["python", plugin_dir+"/ipythonPluginStart.py"], shell=True)
 else:
-    os.execlp("python", "python", plugin_dir+"/ipythonPluginStart.py")
+    os.execlp("/usr/bin/env", "python", "python", plugin_dir+"/ipythonPluginStart.py")
```
kpaharelau: yes

scottdraves: that's not the shebang i don't see that it would do anything.
it's the 1st line of the file: https://en.wikipedia.org/wiki/Shebang_(Unix)

Issue #3955, opened 2016-04-17
file dialog checkCallback broken


scottdraves:
The fix for https://github.com/twosigma/beaker-notebook/issues/3808 introduced a bug where we can only open bkr files now.  others such as ipynb fail.

```
 if (result && result.indexOf('.bkr') === -1){
```

this has at least two problems:
1) bkr is hardcoded, instead it should use whatever extension was passed into the dialog
2) it should check that the extension is at the end of the string

but really it's worse than that, what if a directory name happens to end in .bkr?  there is no way to answer this question by looking at the name, you have to ask the server to check it.

scottdraves:
https://github.com/twosigma/beaker-notebook/blob/a3f52f70c3b705285ca1fe47b1ad21832a0a983b/core/src/main/web/app/helpers/core.js#L79

PR #3958, opened 2016-04-18
file dialog checkCallback broken #3955

kpaharelau:
Using server function for check path is directory

### DIFF
--- a/core/src/main/java/com/twosigma/beaker/core/rest/FileIORest.java
+++ b/core/src/main/java/com/twosigma/beaker/core/rest/FileIORest.java
@@ -30,12 +30,7 @@ import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Paths;
-import java.util.ArrayList;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Arrays;
+import java.util.*;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/twosigma/beaker/core/rest/FileIORest.java
```diff
@@ -30,12 +30,7 @@
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Paths;
-import java.util.ArrayList;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Arrays;
+import java.util.*;
```
scottdraves: import \* is against style

core/src/main/web/app/helpers/core.js
```diff
@@ -75,15 +75,32 @@
...
-            path: result
+      newStrategy.checkCallback = function (result) {
+        var deferred = bkUtils.newDeferred();
+        if (!result){
+          deferred.resolve({
+            result: true
           });
-          return false;
+        }else {
+          bkHelper.httpGet(bkHelper.serverUrl("beaker/rest/file-io/isDirectory"),
```
scottdraves: we can't make a rest call directly here.  the file dialog is created using a strategy object to access the filesystem.
this strategy object might access a different filesystem than the normal unix one on this server.

scottdraves:
👍

Issue #3956, opened 2016-04-17
access denied error has TMI


scottdraves:
if you try to open a file that is not accessible, it fails and shows the whole stack trace, which is confusing for a user because it looks like beaker has crashed.  it should just show the 1st line
![screen shot 2016-04-16 at 8 30 55 pm](https://cloud.githubusercontent.com/assets/963093/14584387/4e13baea-0412-11e6-8dc7-2b4b9824f122.png)
and it should not say "Open failed" twice.

PR #3959, opened 2016-04-18
access denied error has TMI #3956

kpaharelau:
New exception handler has been added

> and it should not say "Open failed" twice.

In this case first message will be 'Oops!'


scottdraves:
👍

Issue #3983, opened 2016-04-20
just hitting return should close this alert


scottdraves:
![screen shot 2016-04-20 at 9 54 46 am](https://cloud.githubusercontent.com/assets/963093/14677070/fe7c467c-06dd-11e6-97db-335f1c3b3f7b.png)

PR #3990, opened 2016-04-21
just hitting return should close this alert #3983

kpaharelau:
class 'modal-submit' has been added to the close button


scottdraves:
:+1:

Issue #3973, opened 2016-04-19
ipython kernelspec deprecated


scottdraves:
https://github.com/twosigma/beaker-notebook/blob/master/launcher/ubuntu/install-dependencies.sh#L66

we should switch to jupyter kernelspec.

sshkalikov:
'jupyter kernelspec install-self' is deprecated in ipython 4.x, too) discussion:  https://github.com/jupyter/jupyter/issues/23
and looks like replacement is 'ipython kernel install' (https://github.com/ipython/ipykernel/pull/43) which does the same and isn't deprecated **but** it doesn't work in ipython 3.x at all, so i'm going to use it in install-dependencies.sh, but for Docker images we should keep using 'ipython kernelspec install-self' (it isn't deprecated in ipython 3.x which we're using in docker)

scottdraves:
still need to remove it from the Dockerfile

PR #4014, opened 2016-04-25
#3973. ipython kernelspec deprecated

sshkalikov:


scottdraves:
also needed in Dockerfile

sshkalikov:
sorry i'm not sure, there's [comment on the issue ](https://github.com/twosigma/beaker-notebook/issues/3973#issuecomment-214242130)
in two words: in dockerfile we use old ipython 3.x and this change will break it. and 'ipython kernelspec install-self' is not deprecated in ipython3.

Issue #4003, opened 2016-04-22
groovy output control fails


scottdraves:
```
// But if you set the followin preference, then output is directed to a panel that opens up
// and collects all ouput like a log, and is not saved in the notebook.
// The preference itself is not saved in the notebook.
beaker.prefs.useOutputPanel = false
```

this cell in the groovy tutorial does not work.

scottdraves:
actually it works fine, but the documentation (groovy-examples.bkr) is wrong it should say ...useOutputPanel = true

PR #4024, opened 2016-04-26
groovy output control fails #4003 

kpaharelau:
> actually it works fine, but the documentation (groovy-examples.bkr) is wrong it should say ...useOutputPanel = true  

FIXED


scottdraves:
do not also change the output display logic.

PR #4025, opened 2016-04-26
addevalplugins.js should generated #3707

kpaharelau:
> spot [4:40 AM]
>  this file isn't used anymore is it?
> [4:40]
>  core/src/main/web/plugin/init/addevalplugins.js
> [4:40]
>  should be removed from the repo afaik

 DONE


scottdraves:
generated JS should go in core/src/main/web/app/genweb/

Issue #3856, opened 2016-04-02
add keyboard navigation of language manager


scottdraves:
up/down/left/right should move focus around the grid of buttons, and enter should start the one under the cursor.

PR #4028, opened 2016-04-26
add keyboard navigation of language manager #3856

mnavasiolava:


scottdraves:
user needs to be able to start grid navigation with an arrow key, not just tab.

mnavasiolava:
done

scottdraves:
👍

Issue #4031, opened 2016-04-26
update dockerfile to use latest ipython


scottdraves:
the widgets are broken.

PR #4037, opened 2016-04-27
#4031. update dockerfile to use latest ipython

sshkalikov:


scottdraves:
i wasn't clear in the issue: need to fix the widgets!

scottdraves:
i have ipython 4.1.2 and when i run the cell for the factorial interaction i get no slider.
screen shot 2016-05-06 at 11 43 37 am

sshkalikov:
there must be a lot of output in js console in this case. could you attach it to this PR please?

sshkalikov:
i made some experiments with different versions and looks like all this fix is useless because it works only when **ipywidgets** version is between 5.0 and 5.1.1 (i had ipywidgets 5.0 when i was doing this task, and now i have 5.1.3 and it works not in all cases) :disappointed: 
and now i have no idea how to make it work with any of the possible package versions
ps. what ipywidgets version is installed on your system?

scottdraves:
i have ipywidgets 4.1.1.

seems like we are getting too many versions of this.

the JS files must exist in their python, can't we get them delivered by the server and loaded into the client instead of copying them into our server?

scottdraves:
impressive but i am getting:

require.js:141 Uncaught Error: Module name "ipython3_namespace" has not been loaded yet for context: _
http://requirejs.org/docs/errors.html#notloaded

```
$ ipython --version
3.2.0
```

scottdraves:
OK works in the docker container.  please bring up to date with master.
then merge with nginx-free branch.

PR #4038, opened 2016-04-27
Tom/429 - Search for and import JS libraries

tomca32:
We are using cdnjs.com to look for libraries. They have a pretty simple API documented here: https://github.com/cdnjs/cdnjs/blob/master/documents/api.md


scottdraves:
🐬

Issue #2097, opened 2015-08-12
provide access to beaker version via beaker object


ildipo:
something like

beaker.getVersion() should return the version string e.g. "beaker (c) ... v 1.4.3 Max OS/X released on xx/yy/zzzz"
beaker.getVersionNumber() should return the version number e.g. 1.4.3

PR #4045, opened 2016-04-28
#2097. provide access to beaker version via beaker object

sshkalikov:


scottdraves:
https://pub.beakernotebook.com/#/publications/02454ddc-0e3e-11e6-ab25-b769593b900f
i get:

```
sergey/2097-0-ge6a8a9d
Beaker e6a8a9d sergey/2097-0-ge6a8a9d
```

but first one should be sergey/2097, second one should have that plus hash and the build time/date.

Issue #4052, opened 2016-04-29
Returning a javax.swing.ImageIcon as a code cell results generates an exception.


Zavster:
I'm trying to return an Image as the result of a code cell using Java. A simple example as per below yields this exception (obviously replace path with an image that exists).

import javax.swing.ImageIcon;
return new javax.swing.ImageIcon("c:\Temp\chart.png");

The exception is as follows...

Apr 29, 2016 1:53:50 PM com.twosigma.beaker.jvm.serialization.BasicObjectSerializer writeObject
SEVERE: exception in serialization
org.codehaus.jackson.map.JsonMappingException: sun.awt.image.ToolkitImage cannot be cast to java.awt.image.BufferedImage
    at org.codehaus.jackson.map.ser.StdSerializerProvider._serializeValue(StdSerializerProvider.java:625)
    at org.codehaus.jackson.map.ser.StdSerializerProvider.serializeValue(StdSerializerProvider.java:256)
    at org.codehaus.jackson.map.ObjectMapper.writeValue(ObjectMapper.java:1613)
    at org.codehaus.jackson.impl.JsonGeneratorBase.writeObject(JsonGeneratorBase.java:314)
    at com.twosigma.beaker.jvm.serialization.BasicObjectSerializer.writeObject(BasicObjectSerializer.java:199)
    at com.twosigma.beaker.jvm.serialization.PlotObjectSerializer.writeObject(PlotObjectSerializer.java:45)
    at com.twosigma.beaker.jvm.object.EvaluationResult$Serializer.serialize(EvaluationResult.java:59)
    at com.twosigma.beaker.jvm.object.EvaluationResult$Serializer.serialize(EvaluationResult.java:41)
    at org.codehaus.jackson.map.ser.StdSerializerProvider._serializeValue(StdSerializerProvider.java:610)
    at org.codehaus.jackson.map.ser.StdSerializerProvider.serializeValue(StdSerializerProvider.java:256)
    at org.codehaus.jackson.map.ObjectMapper.writeValue(ObjectMapper.java:1613)
    at org.codehaus.jackson.impl.JsonGeneratorBase.writeObject(JsonGeneratorBase.java:314)

scottdraves:
thanks good catch, will fix asap

scottdraves:
btw in the meantime, BufferedImage may work for you instead.

Zavster:
Indeed, BufferedImage works as expected, so should be a fairly easy fix. Thanks...

[...]

PR #4057, opened 2016-05-02
Returning a javax.swing.ImageIcon as a code cell results generates an…

kpaharelau:
… exception. #4052

I have used idea from http://stackoverflow.com/questions/12533543/converting-a-imageicon-array-to-a-bufferedimage


plugin/jvm/src/main/java/com/twosigma/beaker/jvm/serialization/ImageIconSerializer.java
```diff
@@ -15,6 +15,7 @@
  */
 package com.twosigma.beaker.jvm.serialization;
 
+import java.awt.*;
```
scottdraves: import \* violates style.

scottdraves:
also, it causes the java icon to pop up.
you fixed this problem for R a long time ago, can you fix it for Java too?

otherwise it works.

Issue #3234, opened 2016-01-03
drag image to html or text cell should insert base64 representation


scottdraves:
we should instead the base64 encoding with an wrapper, like

```
<img src="data:image/png;base64,iVBORw0K....">
```

similarly handle jpg.

PR #4070, opened 2016-05-04
#3234. drag image to html or text cell should insert base64 represent…

sshkalikov:
…ation

### DIFF
--- a/core/src/main/web/app/beaker.js
+++ b/core/src/main/web/app/beaker.js
@@ -301,7 +301,7 @@
       }
     });
 
-    beakerModule.run(function($rootScope, $location, $route, $document, bkUtils, bkCoreManager, bkHelper, bkElectron) {
+    beakerModule.run(function($rootScope, $location, $route, $document, bkUtils, bkCoreManager, bkHelper, bkElectron, bkDragAndDropHelper) {
       var user;
       var lastAction = new Date();
       var beakerRootOp = {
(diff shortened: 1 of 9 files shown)


scottdraves:
i dragged a PNG into a notebook and got this error on the JS console:

```
SyntaxError: Unexpected token � in JSON at position 0
    at Object.parse (native)
    at fromJson (http://127.0.0.1:8801/7e0c357/beaker/app/dist/beakerVendor.js:50772:14)
    at defaultHttpResponseTransform (http://127.0.0.1:8801/7e0c357/beaker/app/dist/beakerVendor.js:59623:16)
    at http://127.0.0.1:8801/7e0c357/beaker/app/dist/beakerVendor.js:59714:12
```

scottdraves:
ditto for jpg

sshkalikov:
i made some changes and it will work better, i think, but the fact you get this error means that you drag image on the notebook body, but inserting images into cells works only by dragging images directly into codemirror (and image tag will be inserted right in the place where you drop the image)

scottdraves:
yes it works if i drag to a HTML cell but not to a text cell, I just get binary text.

i thought before you said it could tell if i was dragging in a notebook or an image.

if the user does drag to the notebook body, can you just insert a text cell with the this in it?

sshkalikov:
> but not to a text cell

must be fixed by last commit

> just insert a text cell with the this in it

nice solution, i'll try to do it so

scottdraves:
thanks yup that's better now.

yea please do the one that makes a new cell too...

sshkalikov:
Made it to add new cell when image is dropped on the body. But for the moment i'm just adding new cell to the end of notebook, because i've not found an easy way to find the closest cell to the drop point. Do we need to add cell at the mouse position or it's ok now?

scottdraves:
cool this is great, but when it completes inserting the text cell, it should format it so the image is visible, not leave it open as if they were editing the text.

Issue #3688, opened 2016-03-15
API to load another notebook


scottdraves:
resulting from discussion with @kyotov and @ildipo:

users often have a problem where they want to share code between notebooks in a less formal way than moving the code into a regular module.  an easy way to get this is to define a function on the beaker object that imports another notebook, and then evaluates its init cells.  ie:

```
beaker.loadLibrary("......").then(.......)
```

library notebooks can be developed to define useful functions in their init cells, and then document and demonstrate them in regular cells.
the loadLibrary function should be able to take a list.
it should only load each lib once (keep a table as it goes).
it should work on a URI that beaker can open.

before implementing this, should check how it interacts with the module systems of the underlying languages.  it it possible to define a proper python or R module from inside a beaker notebook?
maybe not necessary but should not be uninformed on this point.

altavir:
In order to do so properly, you will need cell dependencies an cell references, because initialization cells are rather bad way to indicate which part of the notebook is loaded. To top it all, user can not control when initialization cells are loaded (If cell loads slowly, it is very inconvenient).
On the other hand, if you have cell dependencies, you can reference not the whole notebook, but a specific cell, and then this specific cell and its dependencies are loaded when needed.

scottdraves:
we don't want to reference a specific cell.  perhaps someday we'll have dependencies between cells and then this API can be extended to include a cell tag in addition to a notebook URI.  as far as i can tell what you are asking for is a superset of this plan, so this is a good place to start.

altavir:
> as far as i can tell what you are asking for is a superset of this plan

This is not quite correct. In fact, you should make an ideological decision about whether you want to use notebook or cell as your main structural element and adhere to this decision. The example of bad similar decision is java packages. Sun intended them to be major structural element, but in practice they are used only as name spaces and package private visibility modifiers always never used.

In my opinion, notebook is not very good structural element, because it does not have definite way to know which of its elements are calculated and which of calculated ones are up-to-date. Additionally, notebook can and usually does contain some additional cells, which does not matter for particular calculation. How to decide which cells to run? Currently you want to use initialization cells only. But what if you want to run non-initialization cell or if initialization cells have conflicting global state change?

[...]

PR #4079, opened 2016-05-06
#3688. API to load another notebook

sshkalikov:
added loadLibrary(url) method returning promise to beaker object **only for js**
can add it for other languages, too, but it won't be reliable


scottdraves:
didn't work for me.  is there a demo?  maybe i did it wrong.
i created a notebook with a groovy cell that just had one cell

```
x = 10
```

and saved it as ~/lib.bkr.

then i made a new notebook with a JS cell that said

```
beaker.loadLibrary("file:/Users/spot/lib.bkr")
```

and evaluated it and got the following console errors:

```
beakerVendor.js:62726 TypeError: Cannot read property 'type' of null
    at Object.bkHelper.receiveEvaluationUpdate (beakerApp.js:3949)
    at output.then.JavascriptCancelFunction (:8801/3c7a203/beaker/plugin/evaluator/javaScript.js:300)
    at processQueue (beakerVendor.js:65042)
    at beakerVendor.js:65058
    at Scope.$eval (beakerVendor.js:66310)
    at Scope.$digest (beakerVendor.js:66126)
    at Scope.$apply (beakerVendor.js:66418)
    at done (beakerVendor.js:60756)
    at completeRequest (beakerVendor.js:60954)
    at XMLHttpRequest.requestLoaded (beakerVendor.js:60895)(anonymous function) @ beakerVendor.js:62726
beakerVendor.js:62726 TypeError: Cannot read property 'update_id' of null
    at beakerApp.js:9859
    at Scope.$digest (beakerVendor.js:66154)
    at Scope.$apply (beakerVendor.js:66418)
    at done (beakerVendor.js:60756)
    at completeRequest (beakerVendor.js:60954)
    at XMLHttpRequest.requestLoaded (beakerVendor.js:60895)
```

sshkalikov:
Sounds like you're doing everything as designed, so there's my fail. Maybe
this one cell of lib.bkr is not init cell?

scottdraves:
right i forgot to set that.
so if no such cells should do nothing not crash.

scottdraves:
cool, when I set the init cell, it worked! 🐬 

but: looks like it returns an array of the values from the cells.
instead, it should just print the names of the libraries as they are loaded recursively,
and ideally have a progress meter for evaluating the cells within each library.
return value should be just be a string that reports the total number of cells evaluated in each language.

also, if i didn't already have the language loaded and the notebook has cells in another language, then i got an error.  there are two options to fix this:
either start that language, or skip the cells.
i can see arguments for both but i think starting the language is the simpler for users to understand, so let's start with that.

scottdraves:
more serious problem: the 2nd time I run a cell with loadLibrary(), it crashes:

```
beakerVendor.js:62726 TypeError: Cannot read property 'type' of null
    at Object.bkHelper.receiveEvaluationUpdate (beakerApp.js:3949)
    at output.then.JavascriptCancelFunction (javaScript.js:300)
    at processQueue (beakerVendor.js:65042)
    at beakerVendor.js:65058
    at Scope.$eval (beakerVendor.js:66310)
    at Scope.$digest (beakerVendor.js:66126)
    at Scope.$apply (beakerVendor.js:66418)
    at beakerVendor.js:68243
    at completeOutstandingRequest (beakerVendor.js:55294)
    at beakerVendor.js:55571(anonymous function) @ beakerVendor.js:62726(anonymous function) @ beakerVendor.js:59455processQueue @ beakerVendor.js:65050(anonymous function) @ beakerVendor.js:65058Scope.$eval @ beakerVendor.js:66310Scope.$digest @ beakerVendor.js:66126Scope.$apply @ beakerVendor.js:66418(anonymous function) @ beakerVendor.js:68243completeOutstandingRequest @ beakerVendor.js:55294(anonymous function) @ beakerVendor.js:55571
beakerVendor.js:62726 TypeError: Cannot read property 'update_id' of null
    at beakerApp.js:9859
    at Scope.$digest (beakerVendor.js:66154)
    at Scope.$apply (beakerVendor.js:66418)
    at beakerVendor.js:68243
    at completeOutstandingRequest (beakerVendor.js:55294)
    at beakerVendor.js:55571
```

sshkalikov:
now it prints path to library that being loaded, starts missing evaluators and returns number of executed cells
just in case, [there're my test notebooks ](https://www.dropbox.com/sh/9g7t4glf62atl1o/AAB_2nFvQ1pCrfRCr5kytcqxa?dl=0)
also fixed bug with second run and bug with loading notebook without initialization cells

scottdraves:
calling it repeatedly like this:

```
beaker.loadLibrary("/Users/spot/libxasdf.bkr");
beaker.loadLibrary("/Users/spot/lib2.bkr");
```

doesn't really work because the 2nd load does not wait for the 1st to finish.  in fact, it should probably give an error.  we could have users .then everything but i think we should just allow this function to take a list:

```
beaker.loadLibrary(["/Users/spot/libxasdf.bkr",
                               "/Users/spot/lib2.bkr"]);
```

also, there seems to be a caching problem.  if you load a notebook, then edit and save that notebook, and then load it again, you get the old version.  this would make development really hard.

another weirder problem:
say you have a notebook with just JS in it.  you use this JS cell to load a notebook with groovy.
groovy is loaded.  you then create a groovy cell.  then save the notebook & open.
the groovy cells never resolve because groovy was not added to the notebook model.

scottdraves:
and please bring up to date with master.

Issue #319, opened 2014-06-08
Allow user to select password for public server


scottdraves:
Currently long secure passwords are generated by Beaker, and the user cut-and pastes them into their browser.  It would be better to allow users to choose their own password, and store a hash of it in their ~/.beaker config.  The salt should then be generated randomly, see https://github.com/twosigma/beaker-notebook/blob/master/core/src/main/java/com/twosigma/beaker/core/module/config/DefaultBeakerConfig.java#L263

This is a leftover from the previous Issue https://github.com/twosigma/beaker-notebook/issues/212

PR #4082, opened 2016-05-08
Expose public-server password to environment variable.

KillerGoldFisch:
I edited the Dockerfile and DefaultBeakerConfig to add the opportunity to start the server with a static password. If you set the `BEAKERPASSWD` environment variable to `<random>` (which is the default in the Dockerfile) or delete it, the password will be generated randomly as before.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,7 +54,7 @@ RUN chown -R beaker:beaker /home/beaker/.beaker
 ###################
 
 # Password for public-server authentication.
-# Set to "<random>" for randomy generated password.
+# Set to "<random>" for randomly generated password.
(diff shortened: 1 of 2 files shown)


scottdraves:
Hi thanks, yea that looks useful, I believe it has been requested.
To merge we would need a signed CLA, can see the "Contributing" section of the top-level readme.

scottdraves:
https://github.com/twosigma/beaker-notebook/issues/319

KillerGoldFisch:
Just in case... :)

2016-05-08 20:54 GMT+02:00 Scott Draves notifications@github.com:

Issue #3159, opened 2015-12-17
show summary of hidden cell


scottdraves:
if there is a value, then we should show a one-line summary, analogous to how when the input is hidden it still shows the language and the number of lines of code.

the summary would be as follows:
report number of lines of stdout and stderr ("20 lines of stdout")
If a table, then "a table with XX rows and YY columns",
If a chart then "a plot with ZZ items".
If another type then just report its name.

If it's a code cell then show the 1st line, and the total number of lines.

scottdraves:
2nd part of https://github.com/twosigma/beaker-notebook/issues/1561

PR #4104, opened 2016-05-13
#3159 show summary of hidden cell

sshkalikov:


sshkalikov:
done it the easiest way so far. should i change the layout of input cell summary?
example:

scottdraves:
this is great.  that layout is fine, but it should truncate the line if it's too long, so it stops before the run button.

also, if it's an output container, it should say "Container with X items" instead of just "OutputContainer".

and for an output with stdout/stderr like this:
https://pub.beakernotebook.com/#/publications/008e6efa-1943-11e6-9e47-7f6b95ec0fc2

it should also summarize the regular output, whatever it is.

scottdraves:
:+1:

Issue #4113, opened 2016-05-16
y axis label and tick labels can overlap


scottdraves:
https://pub.beakernotebook.com/#/publications/3ce13330-1b80-11e6-88a8-9b8eb184b815

PR #4123, opened 2016-05-17
y axis label and tick labels can overlap #4113

kpaharelau:
I calculate leftLayoutMargin and rightLayoutMargin using labels' length


core/src/main/web/outputdisplay/bko-plot/bko-plot.js
```diff
@@ -236,6 +236,45 @@
           });
         };
 
+        function measureText(pText, pFontSize, pStyle) {
```
scottdraves, line 239: seems like this should come from a std lib like jquery or angular but i can't find anything.

Issue #4109, opened 2016-05-16
improve default plot colors


scottdraves:
the yellow for #3 is ugly and the blues for #2 and #4 are too similar.
![screen shot 2016-05-16 at 10 38 24 am](https://cloud.githubusercontent.com/assets/963093/15292722/6e13227a-1b52-11e6-8773-d25128cd52c6.png)

scottdraves:
F15854 (red)
5DA5DA (blue)
60BD68 (green)
FAA43A (orange)
B2912F (brown)
B276B2 (purple)
DECF3F (yellow)
4D4D4D (gray)
F17CB0 (pink)

PR #4127, opened 2016-05-18
improve default plot colors #4109

kpaharelau:
It is ready for a base theme. I need list of colurs for a ambiance theme

example with new colors

image


scottdraves:
please apply them to ambiance mode as well and then i can adjust.

Issue #4115, opened 2016-05-16
upperMargin ignored


scottdraves:
the upperMargin (and lowerMargin) optional parameters to YAxis class seems to be ignored.
they help control the relative size and position of the two Y axes.
if you just set the Y axis be default to be the range of the data on it, then that has size 1.
the top and bottom margins are added to this to set the total height visible.
both of them default to 0.

so for example if you set the upperMargin to 1.0, then the data on this axis will occupy the lower 50% of the plot area.

or if upperMargin is 1.0 and lowerMargin is 1.0 then the data will occupy the middle 33% of the vertical plot area.

PR #4133, opened 2016-05-20
#4115. upperMargin ignored

sshkalikov:


scottdraves:
1) lowerMargin is not a parameter to CategoryPlot or Plot, it is for YAxis only.  same for upperMargin.
2) the ruler lines and tick labels should still fill the entire plot.  it's only the region taken by the data points that moves.
3) actually I cannot understand lowerMargin.  for now, ignore it.
4) this test fails:

```
✗ Nanosecond Resolution
      - Failed: Index out of bound. Trying to access element at index: 0, but there are only 0 elements that match locator By.cssSelector("#maing > g")
```

sshkalikov:
after

plugin/jvm/src/main/java/com/twosigma/beaker/chart/AbstractChart.java
```diff
@@ -119,6 +121,24 @@ public double getXUpperMargin() {
...
+
+  public double getUpperMargin() {
+    return upperMargin;
+  }
+
+  public AbstractChart setUpperMargin(double upperMargin) {
+    this.upperMargin = upperMargin;
+    return this;
+  }
+
```
mnavasiolava: do we need 'upperMargin' for AbstractChart? it's YAxis parameter, right?

Issue #4112, opened 2016-05-16
JVM languages leak threads


scottdraves:
every time you run a groovy cell, the number of live threads increases by 2-3.
the treads are named pool-X-thread-Y where X and Y are numbers.

1) fix the thread leak.
2) give the threads a descriptive name.
3) scala & clojure have the same problem. other JVM languages probably also do, check them and fix as needed.

note: i found it easy to debug this with jvisualvm on ubuntu.

PR #4135, opened 2016-05-20
#4112. JVM languages leak threads

sshkalikov:
cometd thread pool still gives ugly names to his threads (usually starts with pool-1-... [example](https://i.gyazo.com/1fc2302452abb2e7bfab81b908ac5546.png) ) but i can't change it


scottdraves:
ok i see that some of the threads now have the name ObservableUpdater-X but they still leak and many are still named pool-1-thread-Y.

I just created an empty notebook, added groovy, and then ran a cell with 1+1 several times:
screen shot 2016-05-20 at 11 40 06 am

altavir:
I encountered the thread naming problem in my own project. Java thread pool executor is very convenient tool, but it does not allow to define thread naming policy. Even if one passes a name to a thread using custom thread factory, it soon becomes obsolete due to thread reusing,
One needs to use something like `Thread.currentThread().setName(...)` to keep thread names up to date with running processes. Also one can keep separate Thread - name mapping.

Issue #4139, opened 2016-05-20
windows may differ between client and server


scottdraves:
https://github.com/twosigma/beaker-notebook/blob/a3f52f70c3b705285ca1fe47b1ad21832a0a983b/core/src/main/web/app/utils/basic/treeview.js#L78

this use of `bkUtils.isWindows` in several places in the file dialog processing is incorrect because it tests if the client is running on windows, but the strings in the file dialog come from the server, which might be linux (for example).

PR #4140, opened 2016-05-23
windows may differ between client and server #4139

kpaharelau:
I have created serverOS object for getting information about server OS type. It uses "beaker/rest/util/version" rest service that returns needed information. Also I have changed 'isWindows',  'isMacOS' and 'osName' from variables to a functions. It is simpler than using promises


kpaharelau:
Tested on a MacOS and Windows

Issue #4089, opened 2016-05-10
System shell commands should output with colors


wllmtrng:
Under Mac OS X, using shell commands within the ipython cell results in improperly colored output. Characters are shown instead of the proper colors. Below is an example with using the `ls` command.

<img width="715" alt="screen shot 2016-05-09 at 7 49 29 pm" src="https://cloud.githubusercontent.com/assets/1335855/15134318/d88f8c70-161f-11e6-9849-879f45a09adc.png">

scottdraves:
thanks for the bug report

PR #4181, opened 2016-06-01
#4089. System shell commands should output with colors

sshkalikov:


core/bower.json
```diff
@@ -37,6 +37,7 @@
     "jquery-throttle-debounce": "v1.1",
     "angular-ui-select": "0.16.1",
     "vis": "4.16.1",
-    "vivagraphjs": "latest"
+    "vivagraphjs": "latest",
+    "ansi_up": "1.3.0"
```
mnavasiolava: hi, we already have ansi_up in dependencies https://github.com/twosigma/beaker-notebook/blob/master/core/bower.json#L33
it is also included in index_template.html https://github.com/twosigma/beaker-notebook/blob/master/core/src/main/web/app/template/index_template.html#L109

sshkalikov: Thanks)

mnavasiolava: no problem :)

Issue #4177, opened 2016-05-31
failed open added to history


scottdraves:
if you open a file and it fails, it should not be added to the history.

to reproduce:  click the open button.
enter a non existing string and hit return.
click away the error dialog.
observe the bad file now appears in the history.

PR #4182, opened 2016-06-01
#4177. failed open added to history

sshkalikov:


core/src/main/web/app/utils/recentmenu.js
```diff
@@ -104,6 +104,19 @@
             _server.addItem(item, refreshMenu);
           }
         },
+        updateRecendDocument: function (oldUrl, item) {
```
scottdraves: Recent not Recend

Issue #3818, opened 2016-03-28
research replacing file browser with a std lib


scottdraves:
https://github.com/Kloudless/file-explorer

scottdraves:
i could have sworn we had an issue for this already but i can't find it.

scottdraves:
looks like kloudless support for local filesystems is an upsell, so we are looking at https://github.com/Studio-42/elFinder instead (but using a Java server instead of PHP).

djaddison:
This seems like a good option if we can skin / restyle it.

PR #4183, opened 2016-06-01
research replacing file browser with a std lib #3818

kpaharelau:
ready for review


scottdraves:
this is great.  we need it to show .bkr files with our document icon (marina made it for the mac).
also, the menu text is too small, and a lot of the menus are disabled.  if they are always disabled we shoulid just not show them.

also we need a way to copy-and-paste in a whole file path.

scottdraves:
also since it has a 3clause BSD license please add it to the credits dialog http://127.0.0.1:8801/ab70fa1/beaker/credits.html (from the "about" dialog, click the link about other open source projects)

scottdraves:
also, how about ambiance mode?

scottdraves:
nice.  needs a checkbox "show hidden files" to the right of the "show .bkr files only" one.

and when it first opens up currently it has the root dir open:
screen shot 2016-06-07 at 10 22 02 pm
it should be open on home, root should be closed, below.

scottdraves:
if you have a file selected, and the Open button is highlighted, and you hit the return key, it should open.

icon should be beaker document, not the beaker.  this one i think: https://github.com/twosigma/beaker-notebook/blob/master/launcher/electron/bkrfiles.icns
also it should be at retina resolution.  the icon should match the text:

screen shot 2016-06-07 at 10 35 59 pm

also, add another row between the bottom of the browser and the checkboxes.
this would have a text field that tracks the current file, like our current file open dialog.
user should be able to click on it, paste, and then hit open or type return and get either that file,
or the dialog would open to that location.

scottdraves:
please merge master to fix tests.

scottdraves:
awesome, almost there!

a few more fixes:
if i create a new notebook and then hit save, the save dialog comes up.
the cursor should be in the text box, ready for me to type and hit return.
the ".bkr" extension should be added automatically, like it is now.

scottdraves:
thanks but if i click save, then start typing a name, it is missing the "/" between my home directory name and the filename.  also it should append ".bkr" if it's not already there.

if i click save and then ESC, the dialog closes (good) but the "saving in progress" spinner is left running.

the analysePath REST is getting called on every keystroke, that's too much.
It should only call it when you hit return or click open.

if you press TAB while typing a file name, it should autocomplete. it should add chars until there is a choice, and it should filter the display below (the list or icon view) to just those choices. If you type to add chars, or if you backspace, then the filter updates when you stop typing (0.3s wait).

scottdraves:
if i click save and then ESC, the dialog closes (good) but the "saving in progress" spinner is left running.

scottdraves:
the analysePath REST is getting called on every keystroke, that's too much.
It should only call it when you hit return or click open.

scottdraves:
if you press TAB while typing a file name, it should autocomplete.  it should add chars until there is a choice, and it should filter the display below (the list or icon view) to just those choices.  If you type to add chars, or if you backspace, then the filter updates when you stop typing (0.3s wait).

scottdraves:
This is great!  There are a few glitches though.
1) i click open, the dialog appears.  i type one char and the list filters.  good.  but then i type backspace and nothing happens.
2) after i tab to complete and then i type backspace it removes an entire component from the path instead of just 1 character
3) if i type a slash it should open what i just typed.
4) on tab, if there are multiple possible completions it should still complete as much as possible.  for example, if i type "sp" the filter shows me the following:

screen shot 2016-06-21 at 3 39 55 pm

if i type tab it should complete out to /Users/spot/spark

scottdraves:
1 and 2 are still a problem.
3 and 4 look good

scottdraves:
confirmed 1 and 2 are fixed.  👍 
continuing to test...

scottdraves:
looks good.  can you please merge with master?

core/build.gradle
```diff
@@ -119,6 +119,13 @@ dependencies {
   compile group: 'org.jvnet.winp', name: 'winp', version: '1.18'
   compile group: 'org.samba.jcifs', name: 'jcifs', version: '1.3.3'
   runtime group: 'org.slf4j', name: 'slf4j-simple', version: '1.7.6'
+
+  compile group: 'org.json', name: 'json', version: '20090211'
+  compile group: 'commons-fileupload', name: 'commons-fileupload', version: '1.3.1'
+  compile group: 'commons-io', name: 'commons-io', version: '2.4'
+  compile group: 'com.mortennobel', name: 'java-image-scaling', version: '0.8.6'
```
scottdraves: is java-image-scaling really used?
i don't see any thumbs...
i would prefer this feature were disabled.

kpaharelau:
I temporary disabled thumbs functionality. It did not work correctly in our application. But it can make sense in the future

scottdraves:
approved 🎆 
waiting for release to settle...

scottdraves:
kick jenkins

Issue #3596, opened 2016-02-26
make table format API


scottdraves:
everything that you set with the GUI you should be able to set with an API when you make the table.

scottdraves:
https://github.com/twosigma/beaker-notebook/pull/4184

PR #4184, opened 2016-06-01
Table format API #3596

mnavasiolava:
please find usage examples in table-api.bkr


scottdraves:
looks great.  needs another example for custom actions that run a cell tag.
also needs to pass test suite.

Issue #4092, opened 2016-05-11
How do I load an npm module when using a Nodejs cell?


ashleydavis:
I'm trying to use [data-forge](https://www.npmjs.com/package/data-forge) in Beaker Notebook.

I've tried installing it globally:

`npm install -g data-forge`

Loading the module produces an error:

![image](https://cloud.githubusercontent.com/assets/625126/15171548/b70e5d3c-1792-11e6-9c20-2c009e6db827.png)

Apparently this is just the way Nodejs works which seems odd. I can't seem to do this from the command line REPL either:

![image](https://cloud.githubusercontent.com/assets/625126/15171565/ca0f7fb0-1792-11e6-9e83-526f52e20afc.png)

So the question is, where should I install npm modules in order to use them with Beaker Notebook?

scottdraves:
thanks for the report.  I don't know if this can be done but we'll look into it and if it's not too hard, we'll fit it in.

ashleydavis:
It would be useful. I'm not sure if Nodejs is useful without npm modules. Unless you are providing builtin Nodejs API for data manipulation?

I'm happy to help out if anyone can help me get started on how this sort of thing should fit in.

sshkalikov:
[this stackoverflow thread](http://stackoverflow.com/questions/15636367/nodejs-require-a-global-module-package) has helped me
just try to set environment variable NODE_PATH to point to the location where npm saves its global packages (you can see this location while installing new module)

scottdraves:
we should document this in that tutorial notebook.  sergey: please add a new section to the node tutorial that explains the NODE_PATH envar and has an example of running an express server, and then loads a page from that server and shows it in an iframe.

ashleydavis:
Thanks @sshkalikov this actually works pretty well.

I created a temporary directory `c:\temp\my_beaker_project`.

Then I set NODE_PATH to this directory:

`set NODE_PATH=c:\temp\my_beaker_project`

Then I installed various npm modules.

From the same command line I then changed to the Beaker Notebook directory and invoked `beaker.command.bat`.

I was then able to import the node modules into a Beaker cell.

It would be better however if this didn't have to be set up manually. For example I'd like to be able to install npm modules directly into the Beaker Notebook directory and have Beaker automatically be able to find them. 

This makes more sense to me, what do you think @scottdraves?

sshkalikov:
[...]

PR #4191, opened 2016-06-02
#4092. new section for node tutorial about 'require'

sshkalikov:


scottdraves:
ha, i didn't know we had express in there already.  so this doesn't make a good demo.
instead, do the tweetstream demo from the front page of http://socket.io/

also, update the versions of of the packages we already have installed.

Issue #4190, opened 2016-06-02
Position of initialization cell indicator is broken if a code cell is collapsed


mnavasiolava:
![image](https://cloud.githubusercontent.com/assets/12935591/15744458/b8951b20-28d4-11e6-9da6-70a3702ec387.png)

PR #4192, opened 2016-06-02
Position of initialization cell indicator is broken #4190

mnavasiolava:


scottdraves:
still broken in advanced mode

mnavasiolava:
it's not displayed in advanced mode.

PR #4193, opened 2016-06-02
add tests for docker container#4146

EfimovVladimir:


EfimovVladimir:
To use this on Jenkins should be the build command:
docker build -f ./test/Dockerfile -t beaker/tests .
docker run beaker/tests

Issue #4198, opened 2016-06-03
GridOutputContainerLayoutManager


scottdraves:
should have a constructor with no args that defaults to 2 columns.

PR #4211, opened 2016-06-06
GridOutputContainerLayoutManager #4198

kpaharelau:
> should have a constructor with no args that defaults to 2 columns.

done


core/config/tutorials/output_container_cell.bkr
```diff
@@ -14364,7 +14364,7 @@
             "input": {
                 "body": [
                     "def o = new OutputContainer()",
-                    "o.setLayoutManager(new GridOutputContainerLayoutManager(2)) // 2 columns",
```
scottdraves, line 14367: please don't change the documentation

Issue #4215, opened 2016-06-07
hidden output with stdout shows json


scottdraves:
https://pub.beakernotebook.com/publications/4ab74978-2c64-11e6-aea7-7f532067268b

scottdraves:
<img width="253" alt="screen shot 2016-06-07 at 12 01 30 am" src="https://cloud.githubusercontent.com/assets/963093/15845643/fc013d8c-2c42-11e6-9c7c-cb9636066454.png">

PR #4217, opened 2016-06-07
#4215. hidden output with stdout shows json

sshkalikov:


scottdraves:
only the JSON should be hidden, not the stdout.

Issue #4214, opened 2016-06-06
Time axis labelling faulty in CombinedPlot of TimePlots


eric-heiden:
When creating a CombinedPlots of TimePlot(s) (e.g. SimpleTimePlot) in a Python3 cell, the ticks of the time axis are labelled wrongly. I.e. instead of year numbers, the timestamp integers are displayed.

This error can be seen in the Tutorial "pythonChartingAPI.bkr" under the section "Combined Plots".

PR #4218, opened 2016-06-07
Time axis labelling faulty in CombinedPlot of TimePlots #4214

kpaharelau:
problem has been fixed for python and js plugins


scottdraves:
the fix is good, need to rerun the tutorials to fix the documentation.

Issue #4180, opened 2016-05-31
Funny Plot


scottdraves:
try zooming in and out, and changing the LoD state by clicking on the red words.
https://pub.beakernotebook.com/publications/8769d536-276c-11e6-a580-27c95165e3bd
the bars move strangely and if you zoom in a lot you can see a broken view:
<img width="388" alt="screen shot 2016-06-01 at 12 16 30 am" src="https://cloud.githubusercontent.com/assets/963093/15697997/d30ff594-278d-11e6-8a11-3ce6565a7e70.png">
actually i think part of the cause is the dashes are boxes with 1 item.
maybe draw a regular point instead of a box if there is only 1 item in it.
and in the hover-text for boxes, show "count: X" in the text with the number
of points filtered into that box.

i think there are two problems here.  one is the controls are confusing, the way there are two red words to click on.  instead there should be a pop-up menu with 2 values "group into boxes" and "no grouping".  For a line plot like the Level of Detail tutorial, the 3 values should be "group into boxes", "group into river", and "no grouping".  

2nd, the level-of-detail feature should not engage if the data is not monotonically increasing.
check the data in the link above, are they monotonic?

in the levels-of-detail tutorial you can also change the code to this:

```
p = new Plot(title: "Drunken Sailor Walks", showLegend: true, lodThreshold: 2000)
p << new Points(x:beer.x, y:beer.y, displayName:"Beer Walk")
p << new Points(x:whiskey.x, y:whiskey.y, displayName:"Whiskey Walk")
```

and generate a Plot that is monotonic.

PR #4220, opened 2016-06-07
Funny Plot #4180

mnavasiolava:


scottdraves:
this works well!

the pop-up menus should be right justified against the border of legend with just a few pixels gap.
screen shot 2016-06-07 at 1 52 45 pm

scottdraves:
i opened up this example: https://pub.beakernotebook.com/publications/8769d536-276c-11e6-a580-27c95165e3bd
and it still showed them as boxes.  it should have been disabled because it is not monotonic.

scottdraves:
looks good except still needs this:

and in the hover-text for boxes, show "count: X" in the text with the number
of points filtered into that box (or river).

if the count is not easily accessible, then we can skip this, it is not very important.

Issue #4221, opened 2016-06-07
notebook with bad languages hangs on load


scottdraves:
if you have a notebook that uses languages that are unknown to beaker, then beaker hangs when you open it.  instead it should just warn you and ignore.

PR #4232, opened 2016-06-08
fixed bug #4221: error handling for unknown evaluators

eric-heiden:


scottdraves:
please merge master to fix tests.

scottdraves:
kick jenkins

scottdraves:
kick jenkins

scottdraves:
looks good but in advanced mode the error is not visible.  the gray triangle should go into the place where the language indicator is, and the error text should be a tooltip for it.

Issue #4242, opened 2016-06-09
insert cell bar flashes on cell insert


scottdraves:
when you insert cells with cmd-shift-a, the blue buttons flash briefly.

PR #4243, opened 2016-06-10
insert cell bar flashes on cell insert #4242

mnavasiolava:
we had fade in/out effect for changing buttons bar opacity on hover. but this animation was also applied when we wanted just hide buttons bar (e.g. when a cell becomes not the last one).
that's why i moved the logic from css to js to be able to determine whether to apply animation or not

### DIFF
--- a/core/src/main/web/app/mainapp/components/notebook/newcellmenu-directive.js
+++ b/core/src/main/web/app/mainapp/components/notebook/newcellmenu-directive.js
@@ -37,7 +37,7 @@
         var menu = $(element).find('.new-cell');
         var fade = function (opacity) {
           if (!$scope.isLarge) {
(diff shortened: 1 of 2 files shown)


scottdraves:
this fixes the problem but i think it must be faster or something too, it seems to trigger a little more often, even when the mouse moves quickly over the buttons without pausing.

is it possible to add a small delay, just 20ms or so before the animation starts?

scottdraves:
Or maybe if the mouse leaves before the animation is complete, then it should be canceled and return to 0 instead of going all the way.  i am not sure what exactly the difference is but i am seeing a lot more blue with this implementation.

mnavasiolava:
it was necessary to stop the current animation before starting a new one

Issue #4245, opened 2016-06-10
Table copy to clipboard is broken


mnavasiolava:
try to copy to clipboard the 1st table from table-api tutorial
![image](https://cloud.githubusercontent.com/assets/12935591/15964675/5847d880-2f22-11e6-8d79-d6c7ac255a65.png)

scottdraves:
i got this:

```
beakerVendor.js:63040 TypeError: Cannot read property 'settings' of undefined
    at Object.$scope.allConverters.3 (beakerOutputDisplay.js:1073)
    at Scope.$scope.exportTo (beakerOutputDisplay.js:673)
    at getTableData (beakerOutputDisplay.js:767)
    at Scope.$scope.doCopyToClipboard (beakerOutputDisplay.js:778)
```

but yea thanks good catch

PR #4248, opened 2016-06-10
Table copy to clipboard is broken #4245

mnavasiolava:


scottdraves:
it doesn't crash but i don't get anything in the clipboard either...

Issue #4253, opened 2016-06-13
Grey out run button during execution of a cell


eric-heiden:
The run button should be disabled and greyed out while a cell is being executed.

PR #4254, opened 2016-06-14
#4253. Grey out run button during execution of a cell

sshkalikov:


scottdraves:
i see the  class getting set but it doesn't seem to have any visual impact.
the button should be visibly disabled.

Issue #4202, opened 2016-06-05
split up charting tutorial


scottdraves:
put sections 1 & 2 (introductory examples plus navigation) into their own notebook.
put the rest into a new notebook called "Catalog of Plot APIs and Features".

PR #4261, opened 2016-06-15
split up charting tutorial #4202

mnavasiolava:


scottdraves:
need to change tests to match.

Issue #4259, opened 2016-06-15
closed output text summary


scottdraves:
it should show the 1st line of the actual output, not "Text". end with ellipsis if it overflows.
optimization, is this important: avoid the css property should copy the first 1000 chars?
unclear if a long hidden string is expensive or not.
eg: the init cell of the LoD tutorial:
<img width="315" alt="screen shot 2016-06-14 at 11 14 36 pm" src="https://cloud.githubusercontent.com/assets/963093/16066996/382cde6c-3286-11e6-9ba2-e1d13a9dfe33.png">

For Html, extract the first 1000 chars of text (not tags) and show it as a single line in the same font,
again truncated with ellipsis. eg closing the output of the big HTML cell in the main tutorial:
<img width="305" alt="screen shot 2016-06-14 at 11 09 41 pm" src="https://cloud.githubusercontent.com/assets/963093/16066868/f4fd8336-3284-11e6-9a00-48c3e74b7182.png">

PR #4262, opened 2016-06-15
closed output text summary #4259

kpaharelau:
done

screen for text output. Show first line with ellipsis if need.
image

screen for html output. Show first 1000 symbols with ellipsis if need.
image

### DIFF
--- a/core/src/main/web/app/mainapp/components/notebook/codecelloutput-directive.js
+++ b/core/src/main/web/app/mainapp/components/notebook/codecelloutput-directive.js
@@ -226,7 +226,7 @@
                 return null;
                 break;
               case 'Text':
-                return firstString('' + result);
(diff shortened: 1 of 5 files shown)


scottdraves:
HTML handling is good but Text is removing the brackets:
screen shot 2016-06-15 at 11 16 37 am

kpaharelau:
Also I have did fix for https://github.com/twosigma/beaker-notebook/issues/4266 in this PR
see
image

image

scottdraves:
good the brackets show up now.  but this isn't what I mean for 4266.  you just added a margin above the output to avoid the overlap.  that margin should not change.  the idea is that _like text cells_, the 1st line wraps early.  Here's an example:
screen shot 2016-06-16 at 3 10 09 pm

Issue #4265, opened 2016-06-15
text cells should have a show/hide toggle


scottdraves:
same as regular input/output cells.

PR #4273, opened 2016-06-16
text cells should have a show/hide toggle #4265

mnavasiolava:
the 1st line of raw text is shown in the cell summary


scottdraves:
when it's collapsed it should not show "Text" in addition to the contents.

please also provide a show/hide toggle for sections.  they would work just the same as the main toggle on the left side.

mnavasiolava:
done

scottdraves:
thanks looks great just needs to pass tests

Issue #4281, opened 2016-06-16
var contents permissions


scottdraves:
~/.beaker/v1/var/sessionBackups/
should be 700 not 755

and recentNotebooks should be 600, and the var directory itself should be 700.

scottdraves:
https://github.com/twosigma/beaker-notebook/issues/4281

scottdraves:
i just realized ~/.beaker/v1/web should also be 700.

PR #4286, opened 2016-06-17
var contents permissions #4281

mnavasiolava:


scottdraves:
this looks good.  i just realized ~/.beaker/v1/web should also be 700.

Issue #4283, opened 2016-06-16
vim mode cmd box partially obscured by line numbers


scottdraves:
![screen shot 2016-06-16 at 4 25 22 pm](https://cloud.githubusercontent.com/assets/963093/16131947/f434311a-33de-11e6-8223-db5d18b09038.png)
enter vim mode, type ":"

PR #4307, opened 2016-06-20
vim mode cmd box partially obscured by line numbers #4283

mnavasiolava:
i included CodeMirror's css used in vim mode. so now search box looks like:

image


scottdraves:
looks good in regular mode but comes out wrong in advanced mode.
screen shot 2016-06-20 at 10 21 46 am

mnavasiolava:
fixed

scottdraves:
the editor mode should not change the layout of the cells.
it looks like this is introducing a vertical margin for the vim cmd box.
that is not needed, the box can sit over the other UI.  it just should not be under anything.

scottdraves:
please position the cmd box below the last line of code.

Issue #1766, opened 2015-06-17
Can open same notebook multiple times


samccone:
#### How to reproduce:
- Create a new notebook
- Save it 
- Open it from Recent notebooks
- Return to the dashboard, reopen the same notebook from open recent

(repeat the last two steps)

![screen shot 2015-06-17 at 9 40 56 am](https://cloud.githubusercontent.com/assets/883126/8208327/0fe4b814-14d5-11e5-983d-b9669775a371.png)

---
#### What is expected

The same notebook is not open multiple times

scottdraves:
This is a general problem that you can open the same notebook multiple times, and you end up editing the same file in different windows.

while it might be hard to exactly always determine if two paths resolve to the same file, we should be able to easily check if we already have open the same exact path.

and on unix we can compare inodes.

scottdraves:
In the electron version if you try to open the same file twice it should just bring the window with the already-open notebook to the front.  in the web version, that's not possible, so instead can issue a warning dialog.

scottdraves:
so if we ever open a notebook that is already open we should warn the user "This notebook is already open in another window.  Do you want to deactivate the other window?" and if they say "no" then just open the start page instead.  if they say yes, then open as normal but server should disconnect the other window.  The other window then gets a popup that says "This notebook was opened in another window, and has been disconnected."  Then you can reconnect or download a copy.  
the server already pushes all open/close events to the start page at least, so any page could subscribe.

if we can actually force the other window to close then we should consider it.
i am not aware that you can force a tab to pop to the front but if possible we should consider it.

sshkalikov:
we can close browser window with simple `window.close()` call
forcing a tab to pop to the front is not so easy.. actually the only solution i've found so far is to call `alert()` - this makes the tab active, but alert window is too ugly.

scottdraves:
if i call window.close() i get: `Scripts may close only the windows that were opened by it.`
i think users might be surprised also, so just disconnecting them and having a popup is my preferred plan.

PR #4308, opened 2016-06-20
#1766. Can open same notebook multiple times

sshkalikov:


core/src/main/java/com/twosigma/beaker/core/rest/ClientWindowsService.java
```diff
@@ -0,0 +1,107 @@
+package com.twosigma.beaker.core.rest;
```
scottdraves, line 16: needs banner

scottdraves:
this is great, basically it worked.  one glitch: i opened a file, then opened a new start page, and opened the same file.  i was correctly notified it was already open, i opened it anyway.  then i went back to the other page and saw i was correctly notified my window was deactivated.  i declined to download a copy, choosing to close this copy.  it went to the start page (fine).  but here's the problem: there i saw in the open documents list, the notebook was listed twice instead of once.

scottdraves:
please merge master in

scottdraves:
1) help->tutorial
2) go back to the tab with the start page
3) click "close" next to the line for the tutorial
4) return to the tab with the tutorial
5) hit reload

the following stack trace appears an empty notebook with spinner is left on the screen:

```
beakerVendor.js:63042 TypeError: Cannot read property '_beaker_model_output_result' of undefined
    at Object.set (beakerApp.js:8454)
    at Object.setup (beakerApp.js:8624)
    at _loadNotebookModelAndResetSession (beakerApp.js:5059)
    at loadNotebookModelAndResetSession (beakerApp.js:4978)
    at beakerApp.js:5191
    at beakerVendor.js:68649
    at completeOutstandingRequest (beakerVendor.js:55361)
    at beakerVendor.js:55637
```

i guess this is really a slightly different bug but the solution is very similar.  the difference is the box that pops up should say "this notebook was closed from the start page. do you want to save a copy?" instead of "this notebook has been opened in another window..."

scottdraves:
This is a more direct bug with this PR:
1) click "new notebook"
2) File -> New Beaker Window
3) click on "New notebook" in the list of open sessions

you now have the same notebook open in two tabs.  at step 3 you should get the warning that the notebook is already open etc.

scottdraves:
i talked this over with davide, and he pointed out that there are really two problems:
1) having the same session open in two windows.  this is very bad, it can corrupt the server.
2) having the same file open in two sessions.  this is only really a problem if/when they hit save, though it can be an annoyance to users if it was unintentional.  this sounds unlikely but it actually happens often with the main tutorial, since every time you use the menu item it opens it again.

in case 1 which can only happen if they copy-and-paste a URL, we should prompt as above.

case 2 we could determine by keeping track of when the various windows have saved the document but actually this could also happen if a program completely external to beaker changes the file under us while we are editing it.  so instead we should do what emacs does, which is when the file is opened, we note the time, and before saving it we check the mod time and if it has changed then we prompt the user and ask them if they want to overwrite, or save-as elsewhere.

scottdraves:
Thanks, small change: instead of "Override" the dialog should say "Overwrite".

And a bug: if you close a session, then the server should be notified.  if i open a noteobok, copy the URL into the clipboard, close the browser tab, open a new tab and paste the URL, then I get the warning even though there is only 1 tab open.

scottdraves:
awesome 👍

but on hold while release settles.

scottdraves:
merge master please

Issue #550, opened 2014-09-04
Add a “last edited” line to the notebooks on control pane


samccone:
Add a “last edited” line to the notebooks on control panel? and the text should read Opened on

![image005](https://cloud.githubusercontent.com/assets/883126/4154314/8f68e9d0-3460-11e4-9a46-bc1a3f82aaf0.jpg)

samccone:
Currently we do not store the last Edited time in the app

@scottdraves @anglee what do you guys think?

melmac48:
if we can't do this easily then lets push to v2

PR #4359, opened 2016-06-28
issue #550: Added “last edited” date to control pane and to the "com.…

konstantinTarletski:
…twosigma.beaker.core.rest.SessionBackupRest"


scottdraves:
if the notebook has not been edited and the time is not defined then do not show the line at all:
screen shot 2016-06-28 at 10 11 32 am

scottdraves:
the value should update as other documents are edited in other browser tabs.
the start page can get push notifications by comet of when the edit times change.

konstantinTarletski:
1) "if the notebook has not been edited and the time is not defined then do not show the line at all:" -- FIXED
2) "the value should update as other documents are edited in other browser tabs.
the start page can get push notifications by comet of when the edit times change." -- ALREADY WAS DONE (it works on my PC, and "due to code" -- implemented ). If not working, it will be great if You can provide me with exact "step to reproduce"

scottdraves:
to reproduce:
1) run beaker, open the start page in 2 side-by-side windows.  in one of them open a notebook.
2) correctly, in the remaining start page, the notebook appears in the list.
3) edit the notebook.  correctly the "Last edited" line appears.
4) continue to edit the notebook.  the time in the Last edited line is never updated.

scottdraves:
i guess the problem is when the notebook is first edited, this is sent to the server, but only that once.
it should continue to send updates, but not on every single edit because that would be too much network activity.  it should send updates at most once per minute (use a timer).  we have a similar function here: https://github.com/twosigma/beaker-notebook/blob/1c9c7d641d01303d894bb41d7d542a8644df4438/core/src/main/web/app/beakerEmbed.js#L228

scottdraves:
note that it took about 30s for the Last edited line to appear, this is too long IMO, should be more like when the notebook is opened, it is almost instant.

konstantinTarletski:
For me it is updating immediately:

Opened on Jul 1, 2016 10:51:43 AM 
Last edited Jul 1, 2016 11:13:49 AM

Opened on Jul 1, 2016 10:51:43 AM 
Last edited Jul 1, 2016 11:13:57 AM

Even in different browsers:
Save in "Chrome" , watch in "Firefox":
I opened one BKR in "Chrome" , open aditional "Tab"  in "Chrome", open controlpane.
Also Open "Firefox" with controlpane.
Then go back to the "Chrome" with "Tab" 1 (where I have opened BKR) then push "Ctrl-S".
Then I see that "Last edited" changed in "Chrome" "Tab" 2 and "Firefox".

Then I repeted this about 5-7 times.
It all working, even after 15 minutes.
And it happent immediately.

So I can`t reproduse this bug.

scottdraves:
by "edit" i just mean typing into a code cell, for example.
when you edit, a \* appears next to the file name, and the "last edited" time should change.
i see it update quickly when i hit save, but it should happen sooner, when the docoument model changes, not the disk file.

Issue #4345, opened 2016-06-24
autocomplete for scala


scottdraves:
does not seem very effective or maybe is not working at all.

PR #4391, opened 2016-07-04
#4345. autocomplete for scala

sshkalikov:
now it must work like in scala REPL: it doesn't look for classes in CP, but it searches in imported classes, so for example if you type
`import java.util._`
`new Ha`
it will suggest HashMap, HashSet, Hashtable
of course, it also searches among _imported_ packages from classpath (not from std lib)


scottdraves:
👍  waiting for internal scala version upgrade.

Issue #4397, opened 2016-07-05
save dialog moves cursor


scottdraves:
1) create a new notebook
2) add javascript cell
3) type into the cell.  cursor is at end of the cell.
4) hit ctrl-S, dialog comes up
5) hit ESC, dialog goes away

expected: the notebook is exactly as it was
observed: the cursor moves to the beginning of the cell.

PR #4403, opened 2016-07-06
#4397. save dialog moves cursor

sshkalikov:


scottdraves:
well that does fix the problem as stated, but I think something deeper is wrong because now if there is a selection then the problem comes back.  Hitting cmd-S and then ESC should not change the codemirror at all right?  I think that setSelection() call needs to be rethought.

Issue #4338, opened 2016-06-24
keybindings to raise and lower section heading level


scottdraves:
cmd-shift < and >

and maybe cmd-shift 1-4 to insert-after

help page too.

mnavasiolava:
cmd-shift 3-4 are used for taking screenshots on Mac, so we need other shortcuts to insert after.
maybe ctrl/alt 1-4?

scottdraves:
raise and lower done, still need insert-after.
cmd-option (on mac)  sounds good (ctrl-alt on PC).

sluhtoja:
cmd-n opens a tab with the specified index. Example, you have 3 tabs open, you click cmd-2, second tab opens. Browsers tried: Safari, Chrome.  Is ctrl/cmd-alt-n acceptable?

n is the section number.

scottdraves:
yes

PR #4413, opened 2016-07-07
#4338 keybindings to insert new section after, when changed/added dirty

sluhtoja:
#4376 raise/lower section should mark notebook as dirty
#4338 cmd-alt 1-4 to insert-after

### DIFF
--- a/core/src/main/web/app/template/keyboardShortcuts.html
+++ b/core/src/main/web/app/template/keyboardShortcuts.html
@@ -120,6 +120,26 @@
             </tr>
             <tr>
                 <td></td>
+                <td>Ctrl-alt-1</td>
+                <td>Insert new section with style 1 after the current one</td>
+            </tr>
+            <tr>
+                <td></td>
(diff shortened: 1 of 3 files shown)


scottdraves:
this looks good except i find the cmd-alt-1 impossible to reach with my fingers.
any other ideas?

i am inclined to use cmd-shift-1 and 2 despite the fact that 3 and 4 are taken.
the smaller sections can either be reached with following cmd-< or we can go with 7 and 8 or something.

sluhtoja:
This is a good idea, main section levels will be available with shortcut, but lower ones you have to lower the level manually.

core/src/main/web/app/template/keyboardShortcuts.html
```diff
@@ -120,6 +120,16 @@
             </tr>
             <tr>
                 <td></td>
+                <td>Ctrl-shift-1</td>
```
scottdraves, line 123: shift -> Shift

Issue #4394, opened 2016-07-05
locked cell menus should fade in/out


scottdraves:
if you lock a cell and then move the mouse over the menus, they appear instantly.

the menus should fade in and out the same as the blue insert-cell buttons.

PR #4414, opened 2016-07-07
Luhtoja/4394

sluhtoja:
#4394 locked cell menus fade


scottdraves:
one issue per PR please.
also, put the descriptive text like "locked cell menus fade" in the title.

sluhtoja:
Will,  do

PR #4443, opened 2016-07-12
add test for 'Notebook Control API' tutorial

EfimovVladimir:


scottdraves:
when cell codezSGiYV, runs it should check the outputs of 4 cells:
itself (as you do), then also the 2 cells that get run, and also check that the other cell was not run (has no output)

Issue #3643, opened 2016-03-10
System notification / popup alert when long-running cell completes


dela3499:
Similar to #569, it would be nice to change tabs and work on something else without having to babysit a long-running cell. Don't run into this often, but might be nice to have.

scottdraves:
yea if there were something i could click in the spinner that would be cool.

scottdraves:
there is a web notification API https://developer.mozilla.org/en-US/docs/Web/API/notification
we just need a way to enable it:

put a dropdown menu indicator (gray triangle pointing down, like on the beaker object) just to the right of the spinner.  if you hover over the spinner or the triangle, the mouse turns to the finger.  if you click, the menu comes up with one item "notify when done".  select it enables it, and a check mark appears on it.

the popup should say evaluation completed and give the summary view (as if hidden) of the result.

PR #4447, opened 2016-07-12
#3643. System notification / popup alert when long-running cell compl…

sshkalikov:
…etes


scottdraves:
the notification menu does not pop up when i try this in the charting tutorial.
also, i am not hearing a sound with the notification.
Play this one: https://www.freesound.org/people/cabled_mess/sounds/349504/

PR #4451, opened 2016-07-12
Revert "dmtrius/4416"

scottdraves:
Reverts twosigma/beaker-notebook#4448


scottdraves:
this broke the cell menus, the z index caused a problem.

PR #4457, opened 2016-07-12
Added keybinding to cancel code cell evaluation execution

samdbrice:


core/src/main/web/app/helpers/core.js
```diff
@@ -711,6 +716,8 @@
             "Cmd-Alt-Down": moveCellDown,
             "Ctrl-Alt-D": deleteCell,
             "Cmd-Alt-Backspace": deleteCell,
+            "Ctrl-C": cancel,
+            "Cmd-C": cancel,
```
scottdraves, line 720: cmd-c is copy so don't change it

Issue #4464, opened 2016-07-13
progress display has too much bottom margin


scottdraves:
![screen shot 2016-07-13 at 9 08 06 am](https://cloud.githubusercontent.com/assets/963093/16804350/6a313412-48d9-11e6-9e62-dcc02ace03bb.png)

PR #4466, opened 2016-07-13
#4464 progress display has too much bottom margin

sluhtoja:


scottdraves:
not right for regular results:
screen shot 2016-07-13 at 3 04 00 pm

scottdraves:
and does not fix the problem in advanced mode:
screen shot 2016-07-13 at 3 06 08 pm

Issue #4467, opened 2016-07-13
move implementation of output summaries


scottdraves:
from Code Cell Output to Output Display.

this new dependency causes a problem for published notebooks: https://github.com/twosigma/beaker-notebook/blame/master/core/src/main/web/app/mainapp/components/notebook/outputdisplay/outputdisplay-directive.js#L36

PR #4470, opened 2016-07-13
#4467. move implementation of output summaries

sshkalikov:


scottdraves:
summary of output of 1st HTML cell of the main tutorial shows up as "undefined":
screen shot 2016-07-13 at 6 30 34 pm

Issue #3312, opened 2016-01-12
jdbc password


scottdraves:
right now you have to put the password  in the notebook text like this:

```
jdbc:postgresql://localhost/test?user=fred&password=secret&ssl=true
```

to fix this there should be a checkbox to prompt the user with a dialog for login when the db connects.
the current format should continue to be supported.  Anything else in the dialog like SSL?

PR #4489, opened 2016-07-15
issue #3312 (SQL login) implemented on new branch

konstantinTarletski:


scottdraves:
👍 looks good pls bring up to date with master.

scottdraves:
does not log in on the 1st try.  i made a cell that should run in s particular DB, and set up its jdbc string like:

```
jdbc:sqlserver://zzzz.twosigma.com:1433;DatabaseName=xxxxxx;user=yyyyyy;password=<prompt>
```

and then ran the cell, it failed with an error:

```
java.sql.SQLException: Cannot create PoolableConnectionFactory (Login failed for user 'yyyyyyy'. ClientConnectionId:c7c2be9d-5f21-48da-be4b-caa8fe88503c)
```

then i run it and and the 2nd time it prompts for password.  i enter it and the user and get the same error again.

scottdraves:
cool, the login procedure works!  some remaining issues:
1) if you press enter after typing the password, that should be the same as pressing OK
2) if you press ESC or click cancel, then it should fail the execution with an error "no password provided".  right now, if you close the dialog the evaluation gets stuck "pending" and cannot be canceled.

scottdraves:
the connection string should wrap sooner so that a scrollbar does not appear.
"Default data source" should go across the dialog, not be restricted to the left column.
delete "connection string" and start the jdbc string on the left column and let it flow all the way across.

keep only one hrule, the one between the jdbc string and the name field.  delete the other 3 lines.

sql-dialog

konstantinTarletski:
All PR notes are done
Also find out that it was a bug with "h2database", fixed it : write workarond for connection.

scottdraves:
👍 
will merge after the spark changes settle down.

Issue #3786, opened 2016-03-24
add JS API for native plotting library


scottdraves:
currently only easily callable from JVM languages.

create functions that create the data structures for interactive plots.  the functions are really simple, just wrappers.

scottdraves:
time series done.  next up: categories.

PR #4492, opened 2016-07-15
add JS API for native plotting library #3786.

mnavasiolava:
hi
missing these plots :)
js api for category plots


scottdraves:
thank you very much!

Issue #4298, opened 2016-06-19
upgrade printing to logging


scottdraves:
use slf4j systematically.
there are a bunch of places in the java code where we just System.out.println and System.err.println.
these should be replace with proper logging.

PR #4494, opened 2016-07-16
Issue #4298 - Upgrade printing to logging. - Added/Updated slf4j logger.

dmtrius:


plugin/cpp/src/main/java/com/twosigma/beaker/cpp/utils/CppEvaluator.java
```diff
@@ -56,6 +56,8 @@
 import com.twosigma.beaker.cpp.autocomplete.CPP14Lexer;
 import com.twosigma.beaker.cpp.autocomplete.CPP14Parser;
 import com.twosigma.beaker.cpp.autocomplete.CPP14Listener;
+import org.slf4j.Logger;
```
scottdraves, line 59: please don't import if not used.
this applies to many files not just this one.

scottdraves:
this is a step in the right direction -- but still needs to replace System.*.println() and (for example the output from nginx and the plugins) with logging.

scottdraves:
kick jenkins

scottdraves:
the problem is here:
https://github.com/twosigma/beaker-notebook/blob/121bbbddf2bea9c3fe5b2aa99150b51836a55778/test/runner#L35
there is a similar one in runner_embed iirc

scottdraves:
still broken because slf4j uses stderr but the python runner reads stdout.

scottdraves:
cool 🍰 

please update to master

Issue #4485, opened 2016-07-15
metrics should include version


scottdraves:

scottdraves:
https://github.com/twosigma/beaker-notebook/blob/master/core/src/main/web/app/utils/utils.js#L151

it should add the beaker version (version from getVersionInfo) to every logged event. (tag it with "version").
version should be computed once at application start, and stored.

PR #4502, opened 2016-07-18
issue #4485 added "version" to the log

konstantinTarletski:
added "version" to the log, like :
event = evaluate
obj = {"plugin":"JavaScript","length":40,"version":"kostja/4485"}


scottdraves:
kick jenkins

Issue #4513, opened 2016-07-19
add margin


scottdraves:
between the two checkbox/label items, to the right of "only".
about 8mm
<img width="447" alt="screen shot 2016-07-18 at 11 23 42 pm" src="https://cloud.githubusercontent.com/assets/963093/16937574/31a420a0-4d3e-11e6-93f3-507517a80abd.png">

PR #4527, opened 2016-07-20
#4513 add margin to file dialog footer checkboxes

sluhtoja:


core/src/main/web/app/app.scss
```diff
@@ -342,6 +342,9 @@ background-color: rgb(205, 223, 214);
   height: auto;
 }
 
+.modal-footer.fixed > span:nth-of-type(2){
```
scottdraves, line 345: why so fancy?
isn't it better to add a style to the element and then access it by name?

PR #4544, opened 2016-07-22
PR for issue 4508 "notifications for run-all progrress"

konstantinTarletski:


scottdraves:
cool.  but if there is an error, and evaluation stops, the notification should still go out (and report the error).

scottdraves:
also, i set

```
beakerRegister.evaluationFinishedNotificationUrl = "/mailme"
```

but did not see the corresponding menu item.

konstantinTarletski:
I move code written by Sergei from 
outputdisplay-directive.js

to the

 /core/src/main/web/app/helpers/core.js
 function initAvailableNotificationMethods()
line 1780

so if it works , then for both porgres bars
if not , then either for both

So menu should be similar.

scottdraves:
I see different menus when I actually run it:
screen shot 2016-07-25 at 11 42 52 am
screen shot 2016-07-25 at 11 43 24 am

PR #4568, opened 2016-07-28
PR for issue 4516 "run-all progress meter never goes to 100%"

konstantinTarletski:


core/src/main/web/app/mainapp/mainapp.js
```diff
@@ -863,7 +869,7 @@
               }
               
               $scope.completedCells = 0;
-              $scope.runnAllRuning = false;
```
scottdraves, line 866: should be spelled:

runAllRunning

scottdraves:
👍

Issue #4552, opened 2016-07-22
file permissions dialog broken in ambiance theme


scottdraves:
checkboxes are invisible and don't work afaik
![screen shot 2016-07-22 at 4 05 32 pm](https://cloud.githubusercontent.com/assets/963093/17069721/27d1ce16-5026-11e6-9137-46a716a595e0.png)

PR #4569, opened 2016-07-28
Issue #4552 - File permissions dialog broken in ambiance theme

dmtrius:
Fixed checkbox visibility


scottdraves:
this isn't the right fix.  this shows the normal-theme check boxes:
screen shot 2016-07-28 at 1 15 46 pm

we should show the ambiance-theme check boxes.
for example, from the EasyForm tutorial:

screen shot 2016-07-28 at 1 24 20 pm

Issue #4571, opened 2016-07-28
quoting of SQL values


scottdraves:
https://github.com/twosigma/beaker-notebook/blob/c619e1fe3e4a5eb94c3362187b8431cf7d2980f6/plugin/sqlsh/src/main/java/com/twosigma/beaker/sqlsh/utils/QueryExecutor.java#L156

what if there is a ' inside the value?

scottdraves:
this is just part of https://github.com/twosigma/beaker-notebook/issues/4396
which is open now, so closing.

PR #4591, opened 2016-08-03
PR for issue #4571 "quoting of SQL values"

konstantinTarletski:


plugin/sqlsh/src/main/java/com/twosigma/beaker/sqlsh/utils/QueryExecutor.java
```diff
@@ -153,7 +153,7 @@ protected static String setObject(String sql, Object value){
       String begin = sql.substring(0, index);
       String end = sql.substring(index + 1, sql.length());
       if(value instanceof String){
-        ret = begin + "\'" + value + "\'" + end;
+        ret = begin + "'" + replaseString(((String) value), "'", "''", 0) + "'" + end;
```
scottdraves, line 156: doesn't Java have replaceAll() method?

dmtrius, line 156: What about `U+0027` or `&#39;` ?

scottdraves, line 156: that's a good question, what encoding are the queries?

konstantinTarletski, line 156: Yes there is replaceAll method in Java. But I tried different escape
characters. And for some it was difficult to write regular expressions.
That's why I wrote custom method. If for some reason we change escape
character  (for some DB), it will be easy. Let me know if I need to rewrite
this part.

3 Авг 2016 г. 7:03 PM пользователь "Scott Draves" notifications@github.com
написал:

konstantinTarletski, line 156: How can I change encoding of database to test? Does we support all
encodings?

3 Авг 2016 г. 7:30 PM пользователь "Scott Draves" notifications@github.com
написал:

konstantinTarletski, line 156: I afraid that there can be problems with dates.

3 Авг 2016 г. 7:30 PM пользователь "Scott Draves" notifications@github.com
написал:

scottdraves:
we need a notebook that stress-tests quoting.

Issue #4482, opened 2016-07-14
add download as CSV item


scottdraves:
to the main table menu, in addition to save as CSV, there should be a download as CSV item.
this should produce the same CSV file (2 items one for the whole table and one for selected rows).
but instead of saving on server with the JS/html file dialog, it should just download a file to the client, producing a browser native file dialog.

For example, open the R tutorial and look at the 2nd cell and click on the hamburger menu in the top-left corner of the table.

PR #4593, opened 2016-08-03
PR for issue #4482 "add download as CSV item"

konstantinTarletski:


scottdraves:
kick jenkins

scottdraves:
CSV files I get from this have no rows, just the headers.

konstantinTarletski:
Strange. I used same method as export. Will check tomorrow.

3 Авг 2016 г. 19:28 пользователь "Scott Draves" notifications@github.com
написал:

scottdraves:
👍 function is good, but the test failure is legit.
since this is an emergency, just disable the test (test/tests/tables.js).
also a UI detail: remove the hrule between "Save selected as CSV" and "Download all as CSV".

Issue #4605, opened 2016-08-05
Null handling in JavaScript cell output


sztobar:
Right now JavaScript cells don't have any way to handle payload / output with `null` value.

fun fact: `(typeof null === 'object') === true`

Also when output from cell is promise that resolves to undefined it get mapped by `beakerObj.transform()` from `payload: undefined` to `payload: null` and this makes further errors as result / payload is checked only for being `undefined` and not for being `null`.

scottdraves:
<img width="614" alt="screen shot 2016-08-05 at 12 28 10 pm" src="https://cloud.githubusercontent.com/assets/963093/17443348/34e267ae-5b08-11e6-868e-4cb109e7a639.png">

PR #4643, opened 2016-08-15
#4605 Null handling in JavaScript cell output

sluhtoja:


scottdraves:
kick jenkins

scottdraves:
this looks like the right idea, but it prints "Undefined" instead of null.
And for values that really are undefined, it should not show any result.
screen shot 2016-08-17 at 10 44 17 am

scottdraves:
also, i still don't understand what "//noinspection JSUnresolvedVariable" does...

sluhtoja:
Intellij is giving an error on the variable define. Basically it is supressing an Intellij based error.
Will remove it.

scottdraves:
👍

PR #4652, opened 2016-08-17
dmtrius/4574

dmtrius:
timedelta column blocks interactive display


scottdraves:
should also apply to IPython, not just Python3.

scottdraves:
kick jenkins

dmtrius:
kick jenkins

dmtrius:
Have a fix. I would create new branch and push fix there.

PR #4654, opened 2016-08-17
dmtrius/4647 - create test notebook for SQL quoting

dmtrius:


scottdraves:
good but put the notebook in test/notebooks, not core/config/tutorials

Issue #4616, opened 2016-08-08
notebook path hard to select


scottdraves:
if you click and drag to select the path to the current notebook in the top bar,
it is almost impossible.

i should be able to click to the right of the filename, and drag left, and select either just the filename, or keep dragging and select the whole path.

sluhtoja:
Don't understand the task, visual reference needed.

scottdraves:
![capture](https://cloud.githubusercontent.com/assets/963093/17623056/24da4d7c-606c-11e6-936f-f32846b940a9.PNG)

if you click at the red X and drag to the left, i should be able to select "wats.bkr"

PR #4656, opened 2016-08-19
#4616 notebook path hard to select

sluhtoja:
needs refractoring, idea testing commit/pr


scottdraves:
i see no change in the behavior of clicking here, there must be a misunderstanding.

scottdraves:
unchanged.

there should be no JS here, just change the HTML so the browser's native click and drag selection works.

Issue #3203, opened 2015-12-29
render tex errors in red instead of throwing


scottdraves:
https://github.com/Khan/KaTeX/issues/237

disable the throwOnError option then KaTeX will render the unsupported source command in red. See https://github.com/Khan/KaTeX#rendering-options for details.

scottdraves:
looks like we need this: https://github.com/Khan/KaTeX/issues/425
to fix this here: https://github.com/twosigma/beaker-notebook/blob/f44d5547ab636b2bc18395f4164026b64e46b731/core/src/main/web/app/helpers/helper.js#L542

scottdraves:
until then, we should just copy this file and fix this function on our own.

scottdraves:
ended up reverting: https://github.com/twosigma/beaker-notebook/pull/4696

new plan: 
1) forget vendorizing and altering auto-render.
2) for TeX cells do it like the PR above did, show it as a red error, not with a popup.  throwOnError: false seems to work here, so keep using it.
3) for text cells, if there is a math error then show the bad math source in red and put the error in a tooltip (appears on hover).

PR #4663, opened 2016-08-22
#3203 render tex errors in red instead of throwing.

sluhtoja:
still throwing with \ though. In my opinion it's understandable.


scottdraves:
needs to apply this to TeX in text cells, not just TeX cells.  See the text and formatting tutorial, you can put TeX in a text cell inside of $ characters.

PR #4665, opened 2016-08-23
4661 add test for tables with timedelta column

EfimovVladimir:


test/tests/tables.js
```diff
@@ -94,7 +109,7 @@ describe('Beaker Tables', function () {
       });
     });
   });
-
+   /*
```
scottdraves: why is this commented out?

EfimovVladimir: It is my fault. I will correct.

EfimovVladimir:
kick jenkins

Issue #1466, opened 2015-04-15
add search and replace feature


scottdraves:

choct155:
Just bumping the request.  It would be a nice addition to a solid product.

scottdraves:
thank you!  agreed.

scottdraves:
design:
1) add item to notebook menu "Search and replace...", it should have a keybinding
2) when selected, a persistent panel appears (placement TBD).  panel has "x" close button.
3) panel has a dialog with search-for text field, replace-with text field, search button, back button, replace button, and replace-all button.
4) search field updates as you type, highlighting all matches and scrolling to the 1st one.
search button scrolls to next match, back button to previous, replace button acts on current match and moves to next match, replace all acts on all matches.

klstack3:
Hey Scott, some quick initial thoughts: http://wzeecq.axshare.com

I’ve placed the find and replace link under the “notebook” menu as it seems to be the closest thing we currently have to “edit” where this sort of functionality would typically be found. I think the bottom bar placement could work but might take up a fair bit of real estate, what if it was a drag and drop menu that users could adjust to their liking? I’ve mocked up a couple limited functionality prototypes at the link.

scottdraves:
thx.  base menu solution looks good, 
the text boxes don't have to be so big (def 1 line), remove the title & explanatory text too.
case sensitive and selected cell vs whole notebook are good ideas (the later should be grayed out if the cursor was not in a cell).  also does need a "previous" button.

if you press ESC then it closes, if you press enter in the find box then it goes next, if you press enter in the replace box then it replaces current and moves to next.

klstack3:
For sure, the explanatory text probably isn't necessary and I see why you might want to remove the title though I think the title should remain to help maintain hierarchy and the distinction that this is a separate module... 

I also wonder if users will want to find and replace large sections of code or text and if the find and replace cells should expand with input for easier viewing? Not sure how often that would be a concern or if they’re more likely to try to replace smaller elements... but since it's code it may take up a couple of lines.

I also left room for more filters if there is something else you think they may be likely to want to ignore or customize in the search.

A revised version: http://wzeecq.axshare.com

scottdraves:
cool thx that's close enough.  large sections of code generally not used here, no need to worry about expansion.  someday we'll have more switches (like regexps) so good to know there is room for them.

scottdraves:
note we have a similar UI element already you can reuse.
from the view menu, select "Show stdout/err" to see it.

PR #4670, opened 2016-08-24
PR for issue #1466 "add search and replace feature"

konstantinTarletski:


scottdraves:
default should be whole notebook, not one cell.

when you close the dialog, the highlighting should go away, and the text cells should return to rendered.

when the dialog opens, the cursor should be in the search-for field.

when you type into the "find" field, it should highlight all matches as you type (in yellow).

there should be a "find previous" button also.

remove the title "find and replace"

the menu item should be "Find and replace" not "Find > Replace"

major bug:
1) open the main tutorial.bkr
2) click the Find > Replace menu item.
3) click in the Find field, type "Beaker"
4) click "Find Next" button.
result: nothing happens.
expected: the first word of the first text cell would highlight.

scottdraves:
Can you make a "find previous button", not a "reverse search" checkbox?

Also, highlight for matches should be yellow.

Please change the menu item "Find > Replace" to "Find and replace"

Control-F to open the dialog is not working.

Hitting return while the cursor is in the Find box should work like clicking Find,
hitting return while the cursor is in the replace box should work like clicking replace.

still need it to highlight matches as you type in the find box.  it should wait until you stop typing for about 150ms, then run the search.

scottdraves:
this is getting close :)

1) highlight current match with orange like chrome does, not dark blue.
2) the 4 widgets (entire notebook, selected cell, case sensitive, wrap search) should be active if i click on the text, not just on the button/box.
3) the "Replace" button should operate on the current match, and then find the next match.
4) add 2 more options to the choice of entire notebook and selected cell "all code cells" and "all text and cells and section headings".  convert the radio buttons into a pop-up menu.
5) in ambiance theme, fix the checkboxes, and instead of yellow use blue (#2060f0).

scottdraves:
also:
0) control-R and control-F do not seem to be working on this branch.

scottdraves:
1) if the current match is not visible, then the notebook should scroll to bring the match to the top of the screen (10% of the way down the visible page).

scottdraves:
2) on Mac, the binding is to command-F, but the menu is showing control-F.  
please use control-F.
3) ESC should close the dialog.

core/src/main/web/app/styles/theme/base-theme.scss
```diff
@@ -35,6 +35,17 @@
       cursor: $text-cursor;
     }
   }
+    
+  $cm-search-mark-all-color:	#2060f0;
+  $cm-search-mark-color: 		#F26D00;
+    
+  .search-find-all-selected-background{
+  	background-color: $cm-search-mark-all-color;
```
scottdraves, line 43: 2 space indentation please.

core/src/main/web/app/styles/theme/base-theme.scss
```diff
@@ -35,6 +35,17 @@
       cursor: $text-cursor;
     }
   }
+    
+  $cm-search-mark-all-color:	#2060f0;
+  $cm-search-mark-color: 		#F26D00;
```
scottdraves, line 40: no tabs.

scottdraves:
there seem to be 3 problems:
1) the calculation of the coordinate to scroll to is wrong it does not account for the distance of the cm from the top of the document
2) after finding a few matches it stops working completely.  try a big notebook like plot-features.bkr.  search for "Plot".  after clicking "Find Next" 6 times, it stops scrolling completely.
3) it should only scroll if the match is off the page and not visible.

PR #4671, opened 2016-08-25
Luhtoja/4616 notebook path hard to select

sluhtoja:
doesn't select the beaker logo text anymore, but still select the whole path.

### DIFF
--- a/core/src/main/web/app/helpers/core.js
+++ b/core/src/main/web/app/helpers/core.js
@@ -1743,6 +1743,64 @@
       }
     };
 
+    var mouseDown = true;
+    $scope.filePathSelect = function() {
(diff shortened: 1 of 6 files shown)


scottdraves:
great this is better!  but still not 100% fixed.  if i click to the left of the file path and drag right, then it selects correctly until i get to the end of the text, and then the selection vanishes.  it should not vanish.

if i click to the right of the file path and then drag left, the text selects from the start of the text to the cursor, instead of from the end to the cursor.

scottdraves:
great, selection is correct from the right now, but the progress display is broken.  try "run all cells" and i get:
screen shot 2016-08-28 at 8 08 34 pm

Issue #4628, opened 2016-08-09
page up/down keys do not work in code cells


scottdraves:
if the cursor is in a code cell, then the page keys do not work.

PR #4672, opened 2016-08-26
#4628 - page up/down keys do not work in code cells

dmtrius:


core/src/main/web/app/mainapp/mainapp.js
```diff
@@ -1265,7 +1265,13 @@
         }
 
         var keydownHandler = function(e) {
-          if (bkHelper.isSaveNotebookShortcut(e)) { // Ctrl/Cmd + s
+          if (e.which === 33) { // page up
+            window.scrollBy(0, -100);
```
scottdraves: should use the page height, not an arbitrary constant.

dmtrius:
kick jenkins

scottdraves:
if the cursor is in a cell, then you can't page far.  to reproduce:
1) open the tutorial "catalog of plot APIs and features"
2) click in the first code cell.  the blinking cursor appears.
3) type "page down", the window scrolls correctly.
4) type it again, the window scrolls a little bit.  keep typing it, scrolling stops.

expected: scrolling continues all the way to the bottom of the document.

core/src/main/web/app/mainapp/mainapp.js
```diff
@@ -1267,9 +1267,11 @@
         var keydownHandler = function(e) {
           var windowHeight = $(window).height();
           if (e.which === 33) { // page up
+            $(':focus').blur();
```
scottdraves: blurring isn't right.  if you must, then move the cursor to a cell that is visible (could be a text cell or heading, not just code).

Issue #4673, opened 2016-08-28
IPython doesn't display cells with unicode as a table...


mig2:
Evaluate {'a': 'a'} and {'a': u'a'} in IPython. The first displays as a table, the second as text. This doesn't happen with Python 3.

scottdraves:
confirmed: https://pub.beakernotebook.com/publications/5bab39ac-6d72-11e6-a631-8f25c1b38f1a
Python2 isn't handling unicode correctly.

PR #4674, opened 2016-08-29
Issue #4673 - IPython doesn't display cells with unicode as a table

dmtrius:
fixed unicode text processing


plugin/ipythonPlugins/src/dist/ipython/beaker_runtime.py
```diff
@@ -206,8 +208,9 @@ def transform(obj):
     return transformNaN(obj)
 
 def transformNR(obj):
-    if type(obj) == unicode:
-        return str(obj)
+    # this breaks unicode characters with non-ASCII symbols, default encoding is ASCII not utf8
```
scottdraves: don't leave commented out code, just delete it.
put the explanation into the commit message.

PR #4678, opened 2016-08-30
#3203 render tex errors in red instead of throwing

sluhtoja:


sluhtoja:
kick jenkins

core/src/main/web/app/helpers/katexhelper.js
```diff
@@ -0,0 +1,194 @@
+/* eslint no-console:0 */
```
scottdraves, line 1: add this to top of the file:

```
copied from https://github.com/Khan/KaTeX/tree/master/contrib/auto-render
until https://github.com/Khan/KaTeX/issues/425 is resolved.
```

then in comment include their license:
https://github.com/Khan/KaTeX/blob/master/LICENSE.txt

core/src/main/web/app/template/index_template.html
```diff
@@ -106,7 +106,7 @@
   <script src="vendor/bower_components/moment-timezone/builds/moment-timezone-with-data.min.js"></script>
   <script src="vendor/zeroclipboard-2.2.0/dist/ZeroClipboard.js"></script>
   <script src="vendor/bower_components/katex/dist/katex.min.js"></script>
-  <script src="vendor/bower_components/katex/dist/contrib/auto-render.min.js"></script>
+  <!--<script src="vendor/bower_components/katex/dist/contrib/auto-render.min.js"></script>-->
```
scottdraves: delete instead of commenting out

scottdraves:
OK looks like the file has been added but the throwOnError wasn't added so I am still not getting red errors from bad tex in a text cell.  (make a text cell and put in $\asdfasdf$).

scottdraves:
I am still getting a popup for a syntax error like:
a TeX cell that has just a {

in this case, it should leave the input unchanged, format it red, and put a tooltip on it with the text of the error message from katex.

a text cell with ${$ should also show as red with tooltip of the error.

scottdraves:
error in text cell is not handled the same, it should format as much as possible like TeX cells do:
screen shot 2016-08-30 at 2 35 18 pm

scottdraves:
also the main tutorial is giving an TeX error popup about indexOf not being defined...

sluhtoja:
Kick jenkins

scottdraves:
https://pub.beakernotebook.com/publications/66fbe76c-6f7e-11e6-b0b9-8fe2204af9ff

I am still getting a popup error from a TeX cell with just a {

and in a text cell a syntax error like that should be formatted in red.

and still need this to pass jenkins.

and please add this to top of the file:

```
copied from https://github.com/Khan/KaTeX/tree/master/contrib/auto-render
until https://github.com/Khan/KaTeX/issues/425 is resolved.
```

sluhtoja:
kick jenkins

sluhtoja:
kick jenkins

scottdraves:
instead of the popup, should generate a proper error like this:
screen shot 2016-09-01 at 9 04 58 am

except "unknown unexpected token" should be replaced with the error text from katex which now appears in the popup.

sluhtoja:
kick jenkins

PR #4683, opened 2016-08-31
fix flaky error (vim-mode test)

EfimovVladimir:


scottdraves:
it just passed

scottdraves:
kick jenkins

PR #4684, opened 2016-09-01
4682 hide more than a constant number of columns?

sluhtoja:


sluhtoja:
kick jenkins

scottdraves:
if columns are hidden, then show a popup just over the table (not over the notebook) that says "hiding columns 50 to XXX. View complete list to change?" OK/cancel.  If cancel then just close it, if OK then open the dialog/menu for Showing columns selecting from the list with the search box.

sluhtoja:
kick jenkins

scottdraves:
for style, it should not have a blue bar at top, it should not have the "x" to close, should not have a title, and should be much smaller, no bigger than it needs to be.

for function, when you click OK it should open menu (not just show all the columns).
here's the menu:
screen shot 2016-09-08 at 7 27 46 am

scottdraves:
The dialog looks basically right.  We can fine-tune it later.

But the big problem is that after I dismiss it, the table is not interactive.  I can't click on any cells or column headers or column menus.

scottdraves:
please merge in master, there is a table menu bug fix needed.

PR #4685, opened 2016-09-02
3203 rendering tex errors

sluhtoja:


sluhtoja:
kick jenkins

scottdraves:
OK cool, still need to report errors in TeX cells like this:
instead of the popup, should generate a proper error like this:
screen shot 2016-09-01 at 9 04 58 am

except "unknown: Unexpected token (1:1)" should be replaced with the error text from katex which now appears in the popup.

here's how the JS plugin does it:
https://github.com/twosigma/beaker-notebook/blob/master/core/src/main/web/plugin/evaluator/javaScript.js#L178

sluhtoja:
kick jenkins

sluhtoja:
kick jenkins

scottdraves:
kick jenkins

scottdraves:
kick jenkins

scottdraves:
kick jenkins

Issue #4553, opened 2016-07-22
ending autocomplete early


scottdraves:
should delete the unfilled in arguments.

if you are doing an autocomplete, and there are optional arguments, and the user hits ESC or otherwise exits the autocomplete early, then the remaining optional arguments should be deleted.

scottdraves:
from https://twitter.com/andmib/status/756451420128342016

PR #4695, opened 2016-09-07
Issue #4553 - ending autocomplete early - fixed

dmtrius:


scottdraves:
doesn't seem to work.  try:
1) create Python3 cell, import pandas as pd, run it.
2) create Python3 cell, type "pd.Da" and hit TAB.  autocomplete UI comes up.  hit TAB again to select the first option.  it should fill in a call to pd.DataFrame with 5 arguments: data, index, columns, dtype, copy.
3) type "x", then hit TAB.  first argument has now been filled in, cursor on 2nd argument.
4) hit ESC

result: 4 remaining arguments are left in text.
desired: remaining arguments are removed.

dmtrius:
I didn't check that case. Fixing.

EfimovVladimir:
kick jenkins

PR #4696, opened 2016-09-07
Revert "3203 rendering tex errors"

scottdraves:
Reverts twosigma/beaker-notebook#4685

breaks master because replacing a span with a div causes problems.
just changing it back doesn't work either.

also, the math with color is not working.

i think the throwOnError false option is not well enough supported by katex and we need to rethink how to handle errors.


scottdraves:
screen shot 2016-09-07 at 7 45 04 am

scottdraves:
screen shot 2016-09-07 at 7 45 54 am

PR #4698, opened 2016-09-08
3203 render tex errors in red instead of throwing

sluhtoja:
somehow got it to work


scottdraves:
does not fix the blue lambda.  please follow the new plan i wrote yesterday on the issue.

PR #4701, opened 2016-09-09
jarek/4628: override default behaviour  for page up/down buttons 

jaroslawmalekcodete:
…irror library


scottdraves:
this works well but does not account that part of the notebook is hidden under the menus at top.
it should subtract the menu height from windowHeight before scrolling.

PR #4715, opened 2016-09-14
4702 submenus should not close when mouse leaves

sluhtoja:


scottdraves:
the submenus only open on click, this is fatal.
they should open on hover over their super-item.

scottdraves:
thanks that fixes that.

however this branch is 52 commits behind master, and it includes an old bug so I can't test the column menus.  i'll bet that is also why this branch has conflicts that prevent clean merge.

please rebase and always start new branches from the tip of master.

sluhtoja:
changed branch, created new PR

PR #4716, opened 2016-09-15
4702 submenus should not close when mouse leaves

sluhtoja:
Still needs javascript to be global. If there is no previous table open, dropdowns doesn't work as wanted. Refresh page where no tables on screen, open beaker menu and select dropdown and hover away from it- submenu closes.


sluhtoja:
kick jenkins

sluhtoja:
kick jenkins

PR #4723, opened 2016-09-16
jarek/4526: add z-index property to make dark blue rectangle top

jaroslawmalekcodete:


scottdraves:
function looks fine, but the test failure is legit, the test needs to be updated:

```
Adding custom actions
[firefox #1-24]         - Failed: No element found using locator: By.cssSelector("ul.context-menu-list[style*=\"z-index: 1\"]")
```

EfimovVladimir:
kick jenkins

Issue #4727, opened 2016-09-18
shrink search/replace dialog to corner instead of tray


scottdraves:
this is how it looks with the tray:
<img width="1113" alt="screen shot 2016-09-17 at 10 22 28 pm" src="https://cloud.githubusercontent.com/assets/963093/18612354/b97ee728-7d25-11e6-81ba-897c994efbde.png">
it should instead have this shape, layout, and spacing (and not be resizeable):
<img width="1111" alt="panel screen shot 2016-09-17 at 10 23 11 pm" src="https://cloud.githubusercontent.com/assets/963093/18612444/17fbff14-7d28-11e6-81cc-c04800371d55.png">

PR #4736, opened 2016-09-20
PR for issue #4727 "shrink search/replace dialog to corner instead of tray"

konstantinTarletski:


core/src/main/web/app/styles/stdout.scss
```diff
@@ -68,7 +62,27 @@
   width: 100%;
   z-index: 1000;
   bottom: 0;
-  height: 300px;
+  height: 250px;
+}
+
+.search-replace {
```
scottdraves: this should not be in stdout.sccs anymore

Issue #4726, opened 2016-09-18
mouse should be pointing finger


scottdraves:
on hover over the check-boxes and adjoining text in the search/replace dialog

PR #4752, opened 2016-09-22
PR for issue #4726 "mouse should be pointing finger"

konstantinTarletski:


scottdraves:
1) hand cursor should be active when the mouse is over the text too, since clicking on it also toggles.
2) the border of the search pane flashes blue when i click on the checkboxes.
image

konstantinTarletski:
1) hand cursor should be active when the mouse is over the text too, since clicking on it also toggles. --FIXED
2) the border of the search pane flashes blue when i click on the checkboxes. --FIXED

Issue #4295, opened 2016-06-17
bad error message from R


scottdraves:
`read.csv("XXX")` produces just one line of error in Beaker:

> Error in file(file, "rt") : cannot open the connection

it's missing the additional warning message "cannot open file 'XXX': No such file or directory"

dmtrius:
Evaluation is done by this way:
com/twosigma/beaker/r/utils/RServerEvaluator: EXP result = connection.eval(tryCode);
org/rosuda/REngine/Rserve/RConnection: 
`if(var2 != null && var2.isOk()) {
        return this.parseEvalResponse(var2);
      } else {
        throw new RserveException(this, "eval failed", var2);
      }`

and after that in output object we have 'Error in file(file, "rt") : cannot open the connection'
This error is from RConnection, this error message is produced by R evaluator. I do not see any way to fix it.

scottdraves:
How is it that R shows the correct error when run in the terminal?
It must be doing something that we are not to catch it and show it.
We should do the same.

dmtrius:
`REXP result = connection.eval(tryCode);`
`connection = new RConnection("127.0.0.1", port);`
`org.rosuda.REngine.Rserve.RConnection` is a class from Rserve-0.6-1.jar library.

![screenshot from 2016-07-26 18-57-18](https://cloud.githubusercontent.com/assets/132371/17147393/ec4ebd3c-5362-11e6-928c-ade7b9d91934.png)

scottdraves:
here's another example of a warning that is lost.   cmd line R if I try to install a package but I can't I get an error like this:

```
> install.packages('xxxxxxzzzz', repos='http://cran.us.r-project.org')
Warning message:
package ‘xxxxxxzzzz’ is not available (for R version 3.3.1)
```

but in Beaker the warning message is lost:

![screen shot 2016-08-29 at 11 52 27 am](https://cloud.githubusercontent.com/assets/963093/18057706/195125e0-6ddf-11e6-9e26-a95108f40866.png)

the output is empty, no error at all.

scottdraves:
some discussion: https://stat.ethz.ch/pipermail/r-devel/2005-August/034189.html

PR #4759, opened 2016-09-23
PR for issue #4295 "bad error message from R"

konstantinTarletski:


scottdraves:
this looks good but I want to research it more.

plugin/r/src/main/java/com/twosigma/beaker/r/utils/RServerEvaluator.java
```diff
@@ -508,9 +488,16 @@ public void run() {
             // direct graphical output
             String tryCode;
             connection.eval("do.call(svg,c(list('" + file + "'), beaker::saved_svg_options))");
-            tryCode = "beaker_eval_=withVisible(try({ " + j.codeToBeExecuted + "\n},silent=TRUE))\n"+
-                    "list(beaker_eval_, beaker:::convertToJSON(beaker_eval_$value, beaker:::collapse_unit_vectors))";
+            tryCode = "warn <- {};\n"
```
scottdraves: instead of using variables named "warn" and "err",
use "beaker_warn_" and "beaker_err_" to avoid any conflicts with user code.

PR #4768, opened 2016-09-27
3203 rendering tex errors

sluhtoja:
need to modify katex autor-render.js


scottdraves:
thanks it works, just needs to pass jenkins the test failure is real:

✗ Preview Mode
[firefox #1-29]         - Expected '' to be 'bold times'.

scottdraves:
🐬

PR #4782, opened 2016-09-30
4778 column formatting change should keep column visible

sluhtoja:
will align, will set aligned element as checked in dropdown, will not create new table, will make notebook dirty, after save and refresh has the same alignment as wanted.


scottdraves:
this looks good, but i see that the format submenu has the same problem, can you fix that too please.

Issue #4763, opened 2016-09-23
1/0 hangs R


scottdraves:

scottdraves:
The fundamental problem is the server is sending bad json to the client.
it's very unfortunate that json does not support Infinity and Nan and eventually we will have to work around that.
In the meantime though I don't think adding quotes is a good idea because even correct values are changed by this.  pretty much all json results are altered.

I think a better approach for this bug is to avoid crashing the client on bad json.
this is the cometd library parsing the response:

```
  this.convertToMessages = function (response)
    {
        if (org.cometd.Utils.isString(response))
        {
            try
            {
                return org.cometd.JSON.fromJSON(response);
            }
            catch(x)
            {
                this._debug('Could not convert to JSON the following string', '"' + response + '"');
                throw x;
            }
        }
```

we should catch this error and so if there is any json problem from the server we should report and show it, instead of crashing the client.

furthermore, it looks like cometd has kindly set up its parser to be a hook that we can replace it with one that accepts Infinite and nan.  That would be ideal.

scottdraves:
Python and Groovy seem to work fine:
https://pub.beakernotebook.com/publications/12bad6c4-9150-11e6-b118-5396e5a7b835

PR #4790, opened 2016-10-04
PR for #4763 "1/0 hangs R"

konstantinTarletski:


scottdraves:
The fundamental problem is the server is sending bad json to the client.
it's very unfortunate that json does not support Infinity and Nan and eventually we will have to work around that.
In the meantime though I don't think adding quotes is a good idea because even correct values are changed by this.  pretty much all json results are altered.

I think a better approach for this bug is to avoid crashing the client on bad json.
this is the cometd library parsing the response:

```
  this.convertToMessages = function (response)
    {
        if (org.cometd.Utils.isString(response))
        {
            try
            {
                return org.cometd.JSON.fromJSON(response);
            }
            catch(x)
            {
                this._debug('Could not convert to JSON the following string', '"' + response + '"');
                throw x;
            }
        }
```

we should catch this error and so if there is any json problem from the server we should report and show it, instead of crashing the client.

scottdraves:
furthermore, it looks like cometd has kindly set up its parser to be a hook that we can replace it with one that accepts Infinite and nan.  That would be ideal.

PR #4791, opened 2016-10-04
jarek/4707: configure plugins during nginx startup

jaroslawmalekcodete:


scottdraves:
this looks great, except on the first run I get:

```
:dev:build
:dev:runException in thread "main" java.lang.NullPointerException
        at java.io.Reader.<init>(Reader.java:78)
        at java.io.InputStreamReader.<init>(InputStreamReader.java:113)
        at org.apache.commons.io.IOUtils.copy(IOUtils.java:1906)
        at org.apache.commons.io.IOUtils.toString(IOUtils.java:778)
        at org.apache.commons.io.IOUtils.toString(IOUtils.java:759)
        at com.twosigma.beaker.core.rest.PluginServiceLocatorRest.createPluginConfigs(PluginServiceLocatorRest.java\
:324)
```

Maybe gradle needs to be informed that  pluginsconfiguration.config has been generated? if I "gradle run" again, then it works.

core/src/main/java/com/twosigma/beaker/core/rest/PluginServiceLocatorRest.java
```diff
@@ -518,11 +533,25 @@ public Response locatePluginService(
...
+  }
 
-    pConfig.setProcess(proc);
-    logger.info("Done starting " + pluginId);
-
-    return buildResponse(pConfig.getBaseUrl(), true);
+  private PluginConfig createPluginConfig(final String pluginId, final String command, final String nginxRules) throws IOException, InterruptedException, ExecutionException {
+    String password = RandomStringUtils.random(40, true, true);
+    PluginConfig pConfig;// find a port to use for proxypass between nginx and the plugin
+    final int port = getNextAvailablePort(this.portSearchStart);
```
scottdraves: allow the user to define the ports to use in advance.
if there is an envar beaker_ports it should contain a list of integers separated by spaces.
each time getNextAvailablePort is called, it should just use the ports specified in this envar.
If it runs out of ports, then it should log a warning, then use its current method of searching from the base.

scottdraves:
The path used for the plugin should not be changed.  Right now on master for example, starting C++ produces a rule like:

```
location ^~ /b77053d/cpp.572981720881/
```

but this PR changes that to

```
location ^~ /300ed62/c++.525482719850/
```

it should be using the plugin-id not the name.  the ID is the first argument to bkHelper.locatePluginService(), but grepping it out of the JS will be painful as each of the plugins is written.
So to make getting it easy, it should be defined at the top of the plugin.js, similarly to how PLUGIN_NAME and COMMAND are defined.  just make it called PLUGIN_ID.  If PLUGIN_ID is not defined then can use PLUGIN_NAME.

also I am not sure why the command and name being put into pluginsconfiguration.config.
the command is only needed to start it, and that will be provided at that time, so no need to precompute it.  the only thing that's needed is the ID and the "nginxRules".  nginxRules also needs to be defined up time like PLUGIN_NAME.  See for example julia.js uses python rules.

In fact it looks like we cannot choose ipython1 or ipython2 style in advance :(

we will have to settle on assuming ipython2, and if a request comes in for 1, then allocate it dynamically as master does now.

core/src/main/java/com/twosigma/beaker/core/rest/BeakerPorts.java
```diff
@@ -0,0 +1,101 @@
...
+
+  public int getNextAvailablePort() {
+    while (portFromEnv.hasNext()) {
+      Integer nextPort = portFromEnv.next();
+      if (isPortAvailable(nextPort)) {
+        return nextPort;
+      }
+    }
+
+    int port = getNextGeneratedAvailablePort(this.portSearchStart);
```
scottdraves, line 50: if the envar was set and we run out of ports we need to log a warning.

scottdraves (changes requested):

scottdraves (approved):

PR #4806, opened 2016-10-07
jarek/4783: pass error message to client side when bad toString method

jaroslawmalekcodete:


jaroslawmalekcodete:
kick jenkins

scottdraves:
cool that's a big step in the right direction.  further changes:
1) the stack trace in the cell output should be an error (the bottom the cell is red).
2) the stack trace should not show up in the output tray.

jaroslawmalekcodete:
kick jenkins

scottdraves:
👍 

that works great but i am still concerned -- toString is only called once but now the object and its string form are both stored, so instead of double time, we spend double space.

is it possible to just do it once, in the normal place, and deliver the error?

jaroslawmalekcodete:
kick jenkins

scottdraves:
awesome this looks right.  will test more & waiting for jenkins...

jaroslawmalekcodete:
kick jenkins

scottdraves:
OK this is good but I found there is more to the bug.  Try the following Groovy:

```
class Qux {
  public String toString() {
    return this.failmenow()
  }
}
q = new Qux()
[q, q]
```

This fails like the original, with a stack trace that includes

```
com.twosigma.beaker.jvm.serialization.BasicObjectSerializer$CollectionSerializer.writeObject(BasicObjectSerializer.java:526)
```

Similarly `["a":q]` fails with `MapSerializer`

EfimovVladimir:
kick jenkins

EfimovVladimir:
kick jenkins

Issue #4750, opened 2016-09-22
List of objects not displaying as table


dlaflamme:
1. Put this code in a Groovy cell: data = new groovy.json.JsonSlurper().parseText(new URL("http://data.consumerfinance.gov/api/views.json").text)
2. Run it
3. You will see the string representation. Click the output cell menu, then Display -> Table and the output cell will be blank.

scottdraves:
Here's a simpler example:

```
[["a": 1, "b":2], ["a": 3, "b":4]]
```

scottdraves:
That does display as a table so I think there must be something missing in the data from that website.

dlaflamme:
OK, that may be the case... I inspected the response visually and didn't see anything malformed in any way. I'll check again and post here if i see anything.

dlaflamme:
I put the URL in a validator (https://jsonformatter.curiousconcept.com/) and it said it is invalid because the encoding is not what is expected ("Invalid encoding, expecting UTF-8, UTF-16 or UTF-32.[Code 29, Structure 0]") but I am a bit skeptical that it is invalid... Groovy can parse it to an object just fine, it seems.

dmtrius:
The problem is not in encoding or malformed JSON. Some values are objects so they couldn't be parsed as simple values. This one will not work also:
`new groovy.json.JsonSlurper().parseText('[{"a":1,"b":{"c":3}}]')`

dmtrius:
Sample notebook is published https://pub.beakernotebook.com/publications/c27f6880-8bca-11e6-b031-079f45436a22

scottdraves:
yea this looks like a bug, a table should appear where the b cell has the text {c=3} which is the string version of the dict.  it would look a lot like this:
<img width="192" alt="screen shot 2016-10-11 at 11 31 53 pm" src="https://cloud.githubusercontent.com/assets/963093/19296602/cff3268e-900a-11e6-86b7-0d2ed2f87f53.png">

scottdraves:
see discussion on https://github.com/twosigma/beaker-notebook/pull/4842 and https://github.com/twosigma/beaker-notebook/pull/4826

dmtrius:
Failing test:
`// output container - 14
import com.twosigma.beaker.jvm.object.OutputContainer
def o = new OutputContainer(['pippero', [1,2,3,4,5,6], [ [a:1, b:3], [a:4, c:5] ]])
beaker.v14 = o
o`

Then after you run 
[...]

PR #4826, opened 2016-10-17
#4750 - List of objects not displaying as table - Javascript version

dmtrius:


scottdraves:
i'm not seeing a difference between master and this branch evaluating groovy

```
new groovy.json.JsonSlurper().parseText('[{"a":1,"b":{"c":3}}]')
```

i just get a string both ways.

dmtrius:
You should switch to table view.

dmtrius:
Put this code in a Groovy cell: data = new groovy.json.JsonSlurper().parseText(new URL("http://data.consumerfinance.gov/api/views.json").text)
Run it
You will see the string representation. Click the output cell menu, then Display -> Table

scottdraves:
You are right, It looks great when I switch it, but it should appear as table by default, like the simpler example does.  

I guess whatever code is making that decision needs to be fixed.

PR #4841, opened 2016-10-21
jarek/4803 : open hide cells when matching text for the duration of the search

jaroslawmalekcodete:


jaroslawmalekcodete:
kick jenkins

PR #4842, opened 2016-10-21
#4750 - List of objects not displaying as table - Java version

dmtrius:


scottdraves:
we can't change the data.  with this PR if I evaluate groovy:
`beaker.x = new groovy.json.JsonSlurper().parseText('[{"a":1,"b":{"c":3}}]')`

and then in JS evaluate
`typeof(beaker.x[0].b)` I just get a string instead of an object like on master.

PR #4843, opened 2016-10-21
PR for issue #4763 "1/0 hangs R"

konstantinTarletski:


scottdraves:
thanks kostja that's better but I need to do more research on this issue

PR #4852, opened 2016-10-24
Issue #4750 - List of objects not displaying as table - Javascript version with display as table for arrays

dmtrius:


plugin/jvm/src/main/java/com/twosigma/beaker/table/serializer/TableDisplaySerializer.java
```diff
@@ -58,6 +58,7 @@ public void serialize(TableDisplay value,
       }
       jgen.writeBooleanField("headersVertical", value.getHeadersVertical());
       jgen.writeObjectField("values", value.getValues());
+      jgen.writeObjectField("hasMapValues", value.isHasMapValues());
```
scottdraves, line 61: can it be done without this?

dmtrius: yes, this is a flag for javascript to process only arrays with values as objects, so the part of the work is done on server. This is some kind of optimization.

Issue #4856, opened 2016-10-24
easyform date picker should accept java date, not just string


scottdraves:
https://pub.beakernotebook.com/publications/5584d6fc-9a11-11e6-9a2f-2bd3713b5b03

PR #4872, opened 2016-10-27
Issue #4856 - easyform date picker should accept java date, not just string

dmtrius:


plugin/jvm/src/main/java/com/twosigma/beaker/easyform/EasyForm.java
```diff
@@ -29,18 +29,19 @@
 import com.twosigma.beaker.easyform.formitem.TextField;
 import org.apache.commons.lang3.StringUtils;
 
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.Map;
+import java.text.SimpleDateFormat;
+import java.util.*;
```
scottdraves: import \* is against our style guide.

plugin/jvm/src/main/java/com/twosigma/beaker/easyform/EasyForm.java
```diff
@@ -271,32 +272,48 @@ public String getCaption() {
...
+  @Override
   public String get(final Object key) {
     checkComponentExists((String) key);
     return getComponentMap().get(key).getValue();
   }
 
-  public String put(final String key, final String value) {
+  @Override
+  public String put(final String key, final Object value) {
+    String _value = processValue(value);
```
scottdraves: do not start regular variable name with _

plugin/jvm/src/main/java/com/twosigma/beaker/easyform/EasyForm.java
```diff
@@ -271,32 +272,48 @@ public String getCaption() {
...
     checkComponentExists(key);
     EasyFormComponent component = getComponentMap().get(key);
-    if (!component.checkValue(value)) {
+    if (!component.checkValue(_value)) {
```
scottdraves: should only allow strings and date objects, other objects should throw this exception.  eg this code should throw:

f = new EasyForm()
f.addDatePicker("d")
f["d"] = new ArrayList<Integer>()
f

plugin/jvm/src/main/java/com/twosigma/beaker/easyform/EasyForm.java
```diff
@@ -271,32 +272,48 @@ public String getCaption() {
...
     setChanged();
     notifyObservers();
     component.fireChanged();
     return previousValue;
   }
 
+  private String processValue(final Object value) {
+    // String
+    if (value instanceof String) {
+      return String.class.cast(value);
```
scottdraves: i don't believe this is needed

plugin/jvm/src/main/java/com/twosigma/beaker/easyform/EasyForm.java
```diff
@@ -271,32 +272,48 @@ public String getCaption() {
...
+    // Date
+    } else if (value instanceof Date) {
+      return formatDate(Date.class.cast(value));
+    // any other type
+    } else {
+      return value.toString();
```
scottdraves: this case should be an error, as explained above

plugin/jvm/src/main/java/com/twosigma/beaker/easyform/EasyForm.java
```diff
@@ -29,18 +29,19 @@
...
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.Map;
+import java.text.SimpleDateFormat;
+import java.util.*;
 
 @SuppressWarnings("unchecked")
-public class EasyForm extends ObservableMap<String, String> {
+public class EasyForm extends ObservableMap<String, Object> {
```
scottdraves, line 38: no need to change this

scottdraves:
also, only date components should take Date objects.
if you try to set a regular text field to a date, it should error out.

plugin/jvm/src/main/java/com/twosigma/beaker/easyform/formitem/DatePickerComponent.java
```diff
@@ -29,4 +35,14 @@ public void setShowTime(final Boolean showTime) {
   public Boolean getShowTime() {
     return showTime;
   }
+
+  @Override
+  protected boolean checkValue(final Object value) {
+    return value instanceof Date;
```
scottdraves: needs to take Dates AND Strings

plugin/jvm/src/main/java/com/twosigma/beaker/easyform/EasyForm.java
```diff
@@ -35,7 +35,7 @@
 import java.util.Map;
 
 @SuppressWarnings("unchecked")
-public class EasyForm extends ObservableMap<String, String> {
+public class EasyForm extends ObservableMap<String, Object> {
```
scottdraves, line 38: Why change this from String to Object?  seems unneeded.

dmtrius: Type of value is checked at run time, so that exception is thrown `java.lang.ClassCastException: java.util.Date cannot be cast to java.lang.String`.
To accept different types that parameter should be Object.

plugin/jvm/src/main/java/com/twosigma/beaker/easyform/formitem/DatePickerComponent.java
```diff
@@ -29,4 +35,18 @@ public void setShowTime(final Boolean showTime) {
...
+  protected boolean checkValue(final Object value) {
+    return value instanceof Date || value instanceof String;
+  }
+
+  @Override
+  public String formatValue(Object value) {
+    if (value instanceof Date) {
+      return dateFormat.format(value);
+    } else {
+      return String.class.cast(value);
```
dmtrius, line 49: Should I check for correct value format?

plugin/jvm/src/main/java/com/twosigma/beaker/easyform/EasyForm.java
scottdraves: thanks yea we'll need that

Issue #4871, opened 2016-10-26
time column shows up as [object Object]


scottdraves:
to reproduce:
- open file:config/tutorials/ipython-examples.bkr
- scroll down to the table, select "download as CSV" from the table menu
- look at the contents of the csv file

result: Index column every row is `[object Object]`
expected: 1st column are the dates & times as seen in the notebook

PR #4873, opened 2016-10-27
PR for issue #4871 "time column shows up as [object Object]"

konstantinTarletski:


scottdraves:
the escapeHTML function is important we can't delete it.  try the following JS code:

```
var quarks = new beaker.DataFrame();
quarks.addColumn("name", ["u<p", "charm", "top", "down", "strange"]);
quarks
```

the result "u<p" should be visible as it is but this change shows just "u" because the < is not quoted.
we have a different format, "html" for that.

konstantinTarletski:
Yes, sorry for my bad testing.
Now Fixed.
Test all 3 cases.

Issue #4836, opened 2016-10-19
Save to csv not working


Irubelsky:
Copy to Clipboard: Chrome - OK, Firefox - OK, Safari - doesn't work
Download All as CSV: Chrome - OK, Firefox - doesn't work, Safari - OK
Download Selected as CSV: Chrome - OK, Firefox - doesn't work, Safari - OK 
Referenced to [2899](https://github.com/twosigma/bunsen/issues/2899)

sluhtoja:
Referenced issue does not exist.

scottdraves:
It exists but you do not have permissions on it.  Don't worry about that though, just try these table features on firefox and safari.

sluhtoja:
Copy to Clipboard: Chrome - OK, Firefox - OK, Safari - doesn't work:

clipboard API support to browsers:
http://caniuse.com/#feat=clipboard
I upgraded my mac safari to 9.1.2 and still doesn't allow. Possible reason is that the OS doesn't allow it.
firefox take on browser compatibility
https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand

Accepted answer gives sufficient answer:
http://stackoverflow.com/questions/34045777/copy-to-clipboard-using-javascript-in-ios

this gives insight how you can manipulate cut,copy,delete(in handler), when user has called the event manually.
https://developer.apple.com/library/content/documentation/AppleApplications/Conceptual/SafariJSProgTopics/CopyAndPaste.html

right now zeroclipboard.js is used
possible exchange libraries:
https://clipboardjs.com/

but not sure if the clipboard.js will fix the issue though

sluhtoja:
Download All as CSV, Download Selected as CSV:
Firefox doesn't allow click() with jQuery, manual click on the element is required attributes already assigned.
One possibility is to add attributes to the dropdown element when creating table, notebooks tables, especially big tables could have performance issues. 
Second one is to create a popup to confirm download.

scottdraves:
Thanks.  Please use the popup (in ffox only).

scottdraves:
How about this way of doing downloads on firefox?
http://stackoverflow.com/questions/20303682/file-download-not-working-on-firefox

sluhtoja:
Firefox 4 and earlier does not have a click() method on anchors at all.
Firefox 5 and 6 do have such a method, and the method dispatches an untrusted click event, but they do not allow untrusted events to trigger a link's href.
[...]

PR #4874, opened 2016-10-27
#4836 Save to csv not working. 

sluhtoja:
Safari copy to clipboard not solved.
Download CSV for firefox solved with a popup


scottdraves:
please see my comment on the issue, i would like to understand better why a dialog is required for ffox.  can you provide some links?  thx

Issue #4870, opened 2016-10-26
add :q :wq vim commands


scottdraves:
:q should be a synonym for the "close" menu item.
:wq should save and then close (we already support :w)

PR #4880, opened 2016-10-28
PR for issue #4870 "add :q :wq vim commands"

konstantinTarletski:


core/src/main/web/app/mainapp/mainapp.js
```diff
@@ -729,6 +729,28 @@
...
               }
             },
+            saveNotebookAndClose: function() {
+              saveStart();
+              bkFileManipulation.saveNotebook(saveFailed).then(
+                  function(ret){
+                    //bkSessionManager.close();
+                    //bkCoreManager.gotoControlPanel();
+                    
+                    if (bkEvaluateJobManager.isAnyInProgress() ) {
```
scottdraves: why not call closeNotebook() here?

core/src/main/web/app/mainapp/mainapp.js
```diff
@@ -729,6 +729,28 @@
                 }
               }
             },
+            saveNotebookAndClose: function() {
+              saveStart();
+              bkFileManipulation.saveNotebook(saveFailed).then(
+                  function(ret){
+                    //bkSessionManager.close();
+                    //bkCoreManager.gotoControlPanel();
```
scottdraves: delete

core/src/main/web/app/mainapp/mainapp.js
konstantinTarletski: Because "closeNotebook" tries to save notebook. 
All after save actions take place in "saveDone" (marked us saved).
I do not call "saveDone" because if I call it I need to call "scope.apply", but I do not need this scope at all.
So if I call "closeNotebook" , application think that it is not saved and shows dialog where ofers to me save it.
The I decided to write my own "close".

scottdraves: OK, then please refactor so that the code is not duplicated.

scottdraves:
thanks.  but if i execute the :wq command, the notebook should vanish from the list of open noteobooks.
but i still see it there...

PR #4884, opened 2016-10-31
jarek/4865: override css styles when focus

jaroslawmalekcodete:


jaroslawmalekcodete:
kick jenkins

scottdraves:
i see two problems here:
1) the button should stay dark while the cell is executing, and only return to normal color on completion.  this code it returns to light gray right away.
2) after i run a cell, the hover-darken stops working.

scottdraves:
will get back to this later

PR #4895, opened 2016-11-02
ability to set css styles on plot via customStyles prop

ssadedin:
This change lets the user style plot elements directly using CSS.

I figured I would offer this change even though you might have some reservations that it exposes too much internals of Beaker to the end user. But I have found it gives me tremendous flexibility to style plots how I want them which you can't get through the plot API itself. For example, sometimes I want to put a plot into a presentation but the labels are too small and don't have enough contrast. So I need to beef them up. Then other times I find labels are cut off if I rotate them in certain plots and although it should probably be fixed in Beaker itself, the ability to have some custom CSS in the plot helps me workaround these small layout issues. 

There is a limitation / bug in this implementation in that you can only use one selector per line. Adding a second selector works but it escapes to affect all plots rather than just the one that you are styling. This could be avoided but would require some CSS parsing code to be added in.

Example code:

```groovy
r = new Random()
new Plot(title: "Random Gaussian Points",
         customStyles: [
           ".plot-label { font-size:32px; font-weight: bold; font-family: courier; fill: green;}",
           ".plot-gridline { stroke: purple; stroke-width: 3; }",
         ]
        ) \
  << new Points(
                x: (1..1000).collect { r.nextGaussian() * 10.0d },
                y: (1..1000).collect { r.nextGaussian() * 20.0d }
               )
```

Result:

image

I won't be offended if you don't accept this one, but it might be useful for someone else out there  :-)


scottdraves:
Hi Simon thanks I like it!  But maybe wrapping it up a bit would help, something like
`labelStyle: "font-size:32px; font-weight: bold; font-family: courier; fill: green;"` and
`ruleStyle: "stroke: purple; stroke-width: 3;"`?  that would allow the JS to apply them correctly to the plot at hand.

ssadedin:
Nice idea! Having well defined properties for the common cases will insulate people from messing directly with the classes in the plot while still giving full access to any CSS styles that people can dream up. I think this can possibly be implemented on top of the code I already did fairly trivially. Will do that and come back with an updated pull request.

thanks!

ssadedin:
Note: somehow I've messed up this branch by merging up inconsistent upstream changes. I am opening a new pull request: https://github.com/twosigma/beaker-notebook/pull/4902 which has updates in accordance with the discussion above.

Issue #4910, opened 2016-11-08
closed section with error


scottdraves:
If a section is closed, and there is an error in that section, then show the section title in red (same color as errors are elsewhere).

scottdraves:
the +- open/close toggle should also turn red.

PR #4919, opened 2016-11-10
#4910 change section title color when error occurs

jszalek:


scottdraves:
cool, looks good but doesn't work in ambiance theme.
also needs to pass jenkins (though in this case looks like just a flake failure).

jszalek:
kick jenkins

EfimovVladimir:
kick jenkins

Issue #4916, opened 2016-11-09
no time and sequence if output is missing


scottdraves:
If a cell has no output (e.g. assignment in Python), the dropdown menu with the elapsed time is not shown on the UI.

fix this by moving these items to the cell input menu (at the bottom).

PR #4920, opened 2016-11-10
#4916 move 'elapsed time' and 'sequence no' to input cell if there is no output

jszalek:
- move menu items (elapsed time, sequence no) from output to input section
- hide dropdown menu items without name/label


scottdraves:
1) please give your PR a descriptive title
2) the items should only be moved if the output is not visible.  eg if you evaluate the python code

```
import time
time.sleep(5)
```

then there is no output, and so the menus should be moved.  but python code like

```
23+34
```

the menus should not be moved.

EfimovVladimir:
kick jenkins

Issue #4779, opened 2016-09-29
showing columns should scroll to make them visible


scottdraves:
1. create a big table with python3:

```
import pandas as pd
pd.DataFrame(np.random.randn(10,100))
```
1. click "OK" to show a hidden column.
2. scroll down and click a hidden column to show it.

result: column is shown, but since the table is scrolled all the way to the left nothing is visible on screen.
expected result: table should scroll to make new column visible.

PR #4922, opened 2016-11-11
#4779 showing columns should scroll to make them visible. 

sluhtoja:
column which is made visible is in the middle of the table if possible


scottdraves:
the way it works seems fine, i will have to look into the code later.

Issue #4937, opened 2016-11-16
also in node: undefined results should not show


scottdraves:
run the following in a node cell:
```
var x = 10;
```
output shows as `"undefined"` but, just like if you evaluate `x=10` in a python cell, the output should not appear at all.

PR #4947, opened 2016-11-18
PR for  #4937 "also in node: undefined results should not show"

konstantinTarletski:


plugin/node/src/dist/node.js
```diff
@@ -62,15 +62,28 @@ define(function(require, exports, module) {
...
         },
-        evaluate: function (code, modelOutput) {
+        evaluate: function (code, modelOutput, refreshObj) {
           var deferred = Q.defer();
             var self = this;
             bkHelper.setupProgressOutput(modelOutput);
             bkHelper.httpPost(bkHelper.serverUrl(serviceBase + "/evaluate"), {shellID: self.settings.shellID, code: encodeURIComponent(code)})
             .success(function(ret) {
+
+              if("undefined" == ret || '"undefined"' == ret || "'undefined'" == ret){
```
scottdraves: this isn't the right place for this.  if you evaluate `"undefined"` in node, then the result does not appear.
the check for undefined values should be done in node on the server, not in JS on the client.

scottdraves:
also i am getting a popup error from starting the node plugin on this branch.

plugin/node/src/dist/app/app.js
```diff
@@ -75,23 +75,34 @@ app.post('/session', function (request, response) {
...
     try {
-      var transformed;
-      if(transformation.isCircularObject(result)) {
-        transformed = "ERROR: circular objects are not supported";
+      if (evaluationResult.processed) {
+        response.statusCode = 200;
+        if(transformation.isCircularObject(result)) {
+          transformed = "ERROR: circular objects are not supported";
+        } else {
+          if(undefined == result || "undefined" == result || '"undefined"' == result || "'undefined'" == result){
```
scottdraves: this is the same problem as before.

scottdraves:
just evaluate the cell "0" and get the following error:
```
express deprecated res.send(status): Use res.sendStatus(status) instead at app/app.js:105:16
```

scottdraves:
also try evaluating "null" this should not produce an empty result.

scottdraves:
closing all PRs and issues in big switch to BeakerX.

PR #4962, opened 2016-11-24
jarek/4904: initialize xs and ys by empty list

jaroslawmalekcodete:


scottdraves:
this seems to break some of the normal plot examples, try in the catalog of plot APIs, the one titled "Setting line properties".  on this branch i only get 2 lines but on master i see all 6.

scottdraves:
👍  please merge with master to pass jenkins

scottdraves:
kick jenkins

Issue #4966, opened 2016-11-29
while editing text area cursor can jump


scottdraves:
to reproduce:
* open the easyform tutorial
* scroll down to the multiline text area
* enter several lines of text, use copy and paste especially, keep editing and typing, with pauses
* eventually, it takes a minute sometimes, the cursor mysteriously jumps to the end of the text box.

PR #4987, opened 2016-12-08
#4966 fix infinite request loop in easyform; improve listeners for ea…

jszalek:
…syform input values


scottdraves:
kick jenkins

Issue #4950, opened 2016-11-18
in JS if cell eval result is a fn, nothing appears


scottdraves:
try evaluating in JS:
```
document.createElement
```
result is empty output.  instead it should say something like what you get on the JS console, which is that it's a function.

uluaiv:
See javaScript.js around line ~335:

The problem arises because output in this case is an object and not a string and the resolver struggles with it.

PR #4996, opened 2016-12-20
#4950  Javascript evaluation: when returned a function nothing is shown as output

uluaiv:
If the javascript to evaluate is a function, the result of the evaluation is an object (containing basically the body of such function). The output field resolver expects a string and is not able to handle properly an object. The solution proposed tranforms the function in a string containing its body.

### DIFF
--- a/core/src/main/web/plugin/evaluator/javaScript.js
+++ b/core/src/main/web/plugin/evaluator/javaScript.js
@@ -329,9 +329,9 @@ define(function(require, exports, module) {
               beakerObj.clearOutput();
             }
           } else {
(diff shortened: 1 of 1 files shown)


core/src/main/web/plugin/evaluator/javaScript.js
```diff
@@ -329,6 +329,9 @@ define(function(require, exports, module) {
               beakerObj.clearOutput();
             }
           } else {
+              if(typeof output === 'function') {
```
scottdraves: indent should be 2 spaces, please use google code style.

PR #5083, opened 2017-03-30
bkr to ipynb converter

fishera123:


beaker/bkr2ipynb.py
```diff
@@ -0,0 +1,43 @@
+"""Create a notebook containing code from a beaker notebook.
```
scottdraves, line 15: banner

beaker/bkr2ipynb.py
```diff
@@ -0,0 +1,43 @@
...
+
+nb = new_notebook()
+
+with open(sys.argv[1]) as data_file:
+  data = json.load(data_file)
+
+evaluators = list((cell['evaluator']) for cell in data['cells'] if 'evaluator' in cell)
+kernel_name = max(evaluators, key=evaluators.count)
+
+if kernel_name == 'IPython':
```
scottdraves: IPython for Beaker actually means Python 2

scottdraves:
needs to check its arguments:
```
$ python beaker/bkr2ipynb.py
  File "beaker/bkr2ipynb.py", line 26, in <module>
    with open(sys.argv[1]) as data_file:
IndexError: list index out of range
```

scottdraves:
should work on each of its parameters, so you can say
python bkr2ipynb.py *.bkr

scottdraves:
use https://docs.python.org/2/howto/argparse.html

Issue #5079, opened 2017-03-30
interact of a plot


scottdraves:
the last example of the interact demo notebook:
```
public Plot f(x){
   return new Plot(title: x)

}

com.twosigma.beaker.groovy.widgets.Interactive.interact(this.&f, "s");
```
every time the widget changes, it creates a new plot.
instead the plot should be updated.

konstantinTarletski:
There is example of "interact" with same "Plot".
It prints new Plot every time when we type something in text field, and that is not good.

![default](https://cloud.githubusercontent.com/assets/10614633/24700807/f3283056-1a00-11e7-8a77-b25438c6b939.png)

The main problem of this code that everything is correct:
Interact is working that way, when we update interactive widget then this code is called.
```
          KernelManager.get().publish(mc.buildClearOutput(message, true));
          Object result = function.call(getWidgetValues());
          MIMEContainer resultString = SerializeToString.doit(result);
          KernelManager.get().publish(mc.buildDisplayData(message, resultString));
```
1) clear output, like `from IPython.display import clear_output`
2) Interact function call
3) Gets text or widget to display (there is a main problem, will discuss later)
4) Prints text message to if interract function return text 

So: 
1) Clears only console output not widget area. To clear widget we need to call code like this : 
`button = widgets.ButtonWidget(description="Next")`
`button.close()`
Us I understand, there is no other "legal" ways to clear widgets.
But our `InternalWidget` do not have `close()` function.
2) There we should keep in mind, that what will returned by this function will be real output of the cell.
3) If `SerializeToString.doit(result)` "detects" a widget, then it call "display" method and retudns empty text.
Dispaly work like it should be (like in Python, see picture) it display widget so many times, how much it was called.

![default](https://cloud.githubusercontent.com/assets/10614633/24702013/9598ac68-1a05-11e7-804d-7be9df0a115c.png)


The only what is wrong is that our widgets do not support "update" : new Plot prints with correct data. But Plot printed before stay with old data. 

So to fix this problem we need as minimum to support "close" operation.
And provided example need to be redone, something like this :
```
import static com.twosigma.beaker.widgets.DisplayWidget.display;
import static com.twosigma.beaker.groovy.widgets.Interactive.interact;
import static com.twosigma.beaker.SerializeToString.showInternalWidget;

class PlotHolder{
    static String text = ""
    static Plot plot = null;
}

public static Object f(x){
    PlotHolder.text = x;
    if(PlotHolder.plot != null){
[...]

PR #5094, opened 2017-03-31
PR for issue #5079 "interact of a plot"

konstantinTarletski:
This PR contains only correct example with interract and Plot,
But "interact" working not correct :
1) It draws new "Plot" every time so, it You type 2 letters, tere will be 2 "Plots" in output area.
2) Also if interract function conatis "Ststem.out.println", then it will not displayed in output area.

This is all problems that I know about "interract"


scottdraves:
please add the example when it's working

Issue #5098, opened 2017-03-31
interrupt should not query kernel


scottdraves:
https://github.com/twosigma/beakerx/blob/master/beakerx/js/src/extension.js#L103

this code has 2 problems:
```
function interrupt() {
    getControllCommandList(function(resp) {
      if (undefined != resp.content.data) {
```

the first thing this does is query the kernel for its capabilities.  this is wrong.
this was already done when the kernel started up.  the result should be stored locally in this module,
and then used here when its needed.

also "getControllCommandList" has a typo, it should be "getControlCommandList"

konstantinTarletski:
Fixed. 
But one problem stay :
if we have our "beakerx" kernel then no problem.
But on python like this : I ask for "control commands" and do not get any answer.
And I can not define does I already asked for  "control commands" or not.
So on python kernels I asked for them everytime when I push interrupt.

To fix this we need to add callback handler with empty command list in python kernel.

PR #5100, opened 2017-04-03
PR for issue #5098 "interrupt should not query kernel"

konstantinTarletski:
Fixed. 
But one problem stay :
if we have our "beakerx" kernel then no problem.
But on python like this : I ask for "control commands" and do not get any answer.
And I can not define does I already asked for  "control commands" or not.
So on python kernels I asked for them everytime when I push interrupt.

To fix this we need to add callback handler with empty command list in python kernel.

### DIFF
--- a/beakerx/js/src/extension.js
+++ b/beakerx/js/src/extension.js
@@ -50,8 +50,8 @@ define([
   var base_url = utils.get_body_data('baseUrl');
   var config = new configmod.ConfigSection('notebook', {base_url: base_url});
   var comm;
-  var controlCommandList = undefined;
-
+  var beakerx = undefined;
+  
   config.loaded.then(function() {
     console.log('beaker extension loaded');
   });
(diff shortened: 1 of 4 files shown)


scottdraves:
that is a good point.  we need a better solution for recognizing our own kernels.
it has to be KernelInfo request, that is the obvious place.

scottdraves:
Please check for "BeakerX" in the KernelInfo banner field.

scottdraves:
actually we can add our own field. pls do it like this:
```
$ git diff
a/kernel/groovy/src/main/java/com/twosigma/beaker/groovy/handler/GroovyKernelInfoHandler.java b/kernel/groovy/src/main/java/com/twosigma/beaker/groovy/handler/GroovyKernelInfoHandler.java
index 00a0014..3794e55 100644
--- a/kernel/groovy/src/main/java/com/twosigma/beaker/groovy/handler/GroovyKernelInfoHandler.java
+++ b/kernel/groovy/src/main/java/com/twosigma/beaker/groovy/handler/GroovyKernelInfoHandler.java
@@ -56,6 +56,7 @@ public class GroovyKernelInfoHandler extends KernelHandler<Message> {
     map1.put("codemirror_mode", "groovy");
     map1.put("nbconverter_exporter", "");
     map.put("language_info", map1);
+    map.put("beakerx", true);
     map.put("banner", "BeakerX kernel for Apache Groovy");
     map.put("help_links", new ArrayList<String>());
```

Issue #5119, opened 2017-04-05
improve widget display architecture


scottdraves:
first step is to create a function can display any kind of widget, this is basically the SerializeToString.doit().  just need to make it more accessible.

then we should unify our widget classes: ipywidgets, tables, and other internal widgets.

konstantinTarletski:
useful links:
https://github.com/jupyter-widgets/ipywidgets
http://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html

fishera123:
when we are executing widgets, they are displayed with empty results in the end. I think would be better to check if seo.payload is widget here https://github.com/twosigma/beakerx/blob/master/kernel/base/src/main/java/com/twosigma/beaker/jupyter/msg/MessageCreator.java#L164 and display it right away. this way we will skip result message creating and the empty message won't be displayed.

konstantinTarletski:
Our internal widgets and python widgets are based on Comm.
Basically they work same way.
But they have different base classes.

The plan is do internal widgets and python widgets same in java and in JS side if possible.

What I want to do can be devided in to 3 steps:
1) Make `com.twosigma.beaker.widgets.Widget` class base for all widgets
2) Make widgets to support update.
2.1) support update fields
3) Remove "Serializer" classes.

#
**1) Make `Widget` class base for all widgets**
#

After interface `InternalCommWidget` was remowed.

The next step is remowe this class:
`com.twosigma.beaker.widgets.internal.InternalWidgetUtils`

Now this class creates comm for internal widgets.

For pyton widgets same performs in:
`com.twosigma.beaker.widgets.Widget`

And after remowing `InternalWidgetUtils` this `Widget` class will be base for all widgets.
This means that every widget will have Comm in base class like it should be, like it done in python.

This is the most hard step.


#
**2) Make widgets to support update.**
#

In this step I need to find every updatable field in widgets and put comm update in it(like it done in python).
Also I need assist from Cuba because I tried today send update message to JS side and internal widget was not updated.

#
**2.1) support update fields**
#

Also (like Cuba tolled ne today) our widgets do not support update fields, only whole model can be upated. But to redid it is a separate big task. And need cooperarion from JS programmer and Java programmer. Better way is redo this in scope of step 3.



#
**3) Remove "Serializer" classes.**
#

AS You can see in `com.twosigma.beaker.SerializeToString` there is serializers for internal widgets. This serializers are used to create JSON for Comm message.
[...]

PR #5122, opened 2017-04-06
PR for issue #5119 "improve widget display architecture" --- DisplayAnyWidget#display

konstantinTarletski:
added: com.twosigma.beaker.widgets.DisplayAnyWidget#display


scottdraves:
please add  com.twosigma.beaker.widgets.DisplayAnyWidget to the default imports and remove it from each cell where it is used.

Issue #5102, opened 2017-04-03
empty results?


scottdraves:
https://github.com/twosigma/beakerx/blob/5e879780a0ba2397a494bee27a005f825f7de379/kernel/base/src/main/java/com/twosigma/beaker/SerializeToString.java#L172

shouldn't these all be OutputCell.HIDDEN?

PR #5129, opened 2017-04-07
#5102 empty results

fishera123:


scottdraves:
i don't like how optional was used.
i am looking for a more local change.

PR #5143, opened 2017-04-09
Publish components to Maven local with 'gradle publishToMavenLocal'

hadim:
That's the only way I have found to push Beaker's kernels to my Maven local repo.

If you know another way, I am open to discussion.

Also that would be nice to provide Maven artifacts on MavenCentral. Is that something you will be willing to do ?

### DIFF
--- a/kernel/base/build.gradle
+++ b/kernel/base/build.gradle
@@ -15,6 +15,7 @@
  */
 apply plugin: 'jacoco'
 apply plugin: 'java'
+apply plugin: 'maven-publish'
 
 repositories {
   mavenCentral()
@@ -58,3 +59,14 @@ jacocoTestReport {
 
 test.finalizedBy jacocoTestReport
 
+publishing {
+    publications {
+        maven(MavenPublication) {
+            groupId 'com.twosigma'
+            artifactId 'beaker-kernel-base'
+            version '2.0-SNAPSHOT'
+
+            from components.java
+        }
+    }
+}
--- a/kernel/clojure/build.gradle
+++ b/kernel/clojure/build.gradle
@@ -18,6 +18,7 @@ group 'clojure'
 
 apply plugin: 'application'
 apply plugin: 'java'
+apply plugin: 'maven-publish'
 
 sourceCompatibility = JavaVersion.VERSION_1_8
 targetCompatibility = JavaVersion.VERSION_1_8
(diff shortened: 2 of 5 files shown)


scottdraves:
hi Hadrien, thanks for the PR.
i was planning more on conda than on maven, since that is what jupyter uses.
but i don't mind adding something extra if it's useful for people and doesn't interfere.
the build process is probably going to be reorganized over the next couple of weeks,
and i won't be able to maintain this.  how about waiting until it settles down?
also, i think some of the values in the artifact record could be improved:
COM.twosigma,
beakerX
and this is not v1.7.1 that is for beaker notebook and is no longer current.
i guess what we have now is more like 2.0 alpha.

hadim:
In my point of view, conda is definitively the way to distribute the "standalone" BeakerX kernels (and as I said before I can help with that if needed).

But for the integration with ImageJ, I am building a Scijava kernel that uses under the hood the base BeakerX kernel (see https://github.com/hadim/scijava-jupyter-kernel/pull/15). Because the whole ImageJ/Scijava ecosystem uses Maven and `pom.xml` I need those artifacts to be somewhere.

So for each BeakerX releases that would be really great to have : 

- A conda-force package so end-user can get the BeakerX kernels out-of-the-box in one command line and also each component.

- Maven artifacts published for each BeakerX components so developers can re-use BeakerX kernels in their own project (for `scijava-jupyter-kernel` for example).

That being said, I would like to have only Scijava specific code into `scijava-jupyter-kernel` and I am willing to backport eventual features that could benefit BeakerX when I can.

I am fine if you don't merge this PR but let it open so others devs and I can still use it waiting for you to reorganize the build process. I will also make the changes about `com`, `BeakerX` and the `2.0` version.

hadim:
Just for completeness, the Scijava kernel we are building won't be distributed with conda but directly inside ImageJ through our updating system (https://imagej.net/Updater) and some glue code will be added to run the `jupyter kernelspec install` command.

hadim:
Something similar to that : https://github.com/hadim/scijava-jupyter-kernel/blob/beakerIntegration/src/main/java/org/scijava/jupyter/DefaultKernelConfigurationFile.java#L57

scottdraves:
Cool, I just need a signed CLA to accept.
Please check the README.  thx.

hadim:
Good. 

I have sent the signed CLA.

Also, make sure to push artifacts to MavenCentral (or similar) when doing releases. For now, I use my local Maven repo and build BeakerX myself but in the future when beakerX will be stable enough that would allow people to use BeakerX artifacts without having to download and build themselves the source code.

scottdraves:
🐬

PR #5191, opened 2017-04-14
Switch some log message to debug level

hadim:
To avoid mixing output useful for debugging and developing and output shown to end-user.


scottdraves:
thanks i was thinking the same thing.

PR #5193, opened 2017-04-15
Update setup.py

dsblank:
beaker.plot is a separate import, so you need to list it here.


gnestor (commented):
If we use [`setuptools.find_packages`](http://setuptools.readthedocs.io/en/latest/setuptools.html#using-find-packages), then we won't need to manage this in the future.

```py
from setuptools import setup, find_packages

setup_args = dict(
    ...
    packages                = find_packages(),
    ...
)
```

scottdraves:
grant will include a fix for this in his next PR which should fix the plot library in python.

scottdraves:
thanks for bringing it to our attention!

PR #5205, opened 2017-04-17
Fix empty cell when result is null. Fix #5204

### LINKED ISSUE #5204: duplicate fn call
https://github.com/twosigma/beakerx/blob/master/kernel/base/src/main/java/com/twosigma/beaker/jupyter/msg/MessageCreator.java#L166

found by @hadim

hadim:
Ping @scottdraves 

Now it works for me.

### DIFF
--- a/kernel/base/src/main/java/com/twosigma/beaker/jupyter/msg/MessageCreator.java
+++ b/kernel/base/src/main/java/com/twosigma/beaker/jupyter/msg/MessageCreator.java
@@ -163,7 +163,7 @@ public class MessageCreator {
     if (EvaluationStatus.FINISHED == seo.getStatus() && showResult(seo)) {
       MessageHolder mh = createFinishResult(seo, message);
       if(mh != null){
-        ret.add(createFinishResult(seo, message));
+        ret.add(mh);
       }
     } else if (EvaluationStatus.ERROR == seo.getStatus()) {
       ret.add(createErrorResult(seo, message));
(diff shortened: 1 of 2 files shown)


hadim:
Except that now `gradle test` fails : 

```
com.twosigma.beaker.jupyter.msg.MessageCreatorTest > createMessageWithNullResult_shouldReturnNullStringForNull FAILED
    java.lang.NullPointerException at MessageCreatorTest.java:53

713 tests completed, 1 failed
:kernel:base:test FAILED
:kernel:base:jacocoTestReport

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':kernel:base:test'.
> There were failing tests. See the report at: file:///home/hadim/Insync/Documents/Code/postdoc/beakerx/kernel/base/build/reports/tests/test/index.html

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED
```

I don't know how you want to handle that.

hadim:
I fixed the test.

scottdraves:
this makes null a non-printing result which isn't something we want to do, since we have OutputCell.HIDDEN for that.

hadim:
Ok I see. I'll let you fix it then.

hadim:
Hi @scottdraves , are you still planning to fix this ?

scottdraves:
hi not sure what you mean -- OutputCell.HIDDEN works and widgets return it.
and that duplicate fn call was also fixed.  am i missing something?

hadim:
When the returned object is `null`, jupyter console or notebook display `null`. I am not using any widget on the Jupyter client side. 

Is this OutputCell.HIDDEN beakerx specific or Jupyter specific ?

scottdraves:
> object is null, jupyter console or notebook display null

This is by design.  I know it does not really match Python behavior, and I can see doing it either way, but my understanding is the users prefer to see it.

`OutputCell.HIDDEN` is BeakerX specific.  I would not mind having that value available by a better/shorter name as well as this legacy name.

hadim:
Oh, I see now.

Well, I don't really like to see that value returned but if you think the opposite maybe it's possible to make this behavior configurable?

What do you think ?

scottdraves:
I think if we do https://github.com/twosigma/beakerx/issues/5120 right then the user should be able to configure a repr for null be it "null", empty output, or whatever :)

even if it does not fall out of that i think it's a good idea i filed an issue for it.

hadim:
Ok perfect. Thanks.

PR #5207, opened 2017-04-18
beaker.getCodeCells

fishera123:


kernel/base/src/main/java/com/twosigma/beaker/NamespaceClient.java
```diff
@@ -125,7 +125,17 @@ protected Comm getAutotranslationComm() {
...
+
+  public void getCodeCells(String tagFilter) throws IOException {
+    Comm c = getAutotranslationComm();
+    HashMap<String, Serializable> data = new HashMap<>();
+    data.put("name", "CodeCells");
+    data.put("value", getJson(tagFilter));
+    data.put("parentMessageId",InternalVariable.getSimpleEvaluationObject().getJupyterMessage().getHeader().getId());
+    c.setData(data);
+    c.send();
+  }
```
scottdraves, line 137: this function has to block until it has the value to return.
Something like this should work: https://github.com/twosigma/beakerx/blob/1.7.1/core/src/main/java/com/twosigma/beaker/core/rest/NamespaceService.java#L130

Issue #5216, opened 2017-04-19
infinite loop from returning function in groovy


scottdraves:
evaluate this cell of groovy:
```
def x = { y -> y}
```
result is a crash from infinite loop:
```
java.lang.RuntimeException: com.fasterxml.jackson.databind.JsonMappingException: Infinite recursion (StackOverflowError) (through reference chain: java.util.LinkedHashMap["beaker"]->com.twosigma.beaker.NamespaceClient["outputObj"]->com.twosigma.beaker.jvm.object.SimpleEvaluationObject["payload"]->_run_closure1["delegate"]->script14926065475721000071615["binding"]->groovy.lang.Binding["variables"]->java.util.LinkedHashMap["beaker"]->com.twosigma.beaker.NamespaceClient["outputObj"]->com.twosigma.beaker.jvm.object.SimpleEvaluationObject["payload"]->_run_closure1["delegate"]->script14926065475721000071615["binding"]->groovy.lang.Binding["variables"]->java.util.LinkedHashMap["beaker"]->com.twosigma.beaker.NamespaceClient["outputObj"]->com.twosigma.beaker.jvm.object.SimpleEvaluationObject["payload"]->_run_closure1["delegate"]->script14926065475721000071615["binding"]->groovy.lang.Binding["variables"]->java.util.LinkedHashMap["beaker"]->com.twosigma.beaker.NamespaceClient["outputObj"]->com.twosigma.beaker.jvm.object.SimpleEvaluationObject["payload"]->_run_closure1["delegate"]->script14926065475721000071615["binding"]->groovy.lang.Binding["variables"]->java.util.LinkedHashMap["beaker"]->com.twosigma.beaker.NamespaceClient["outputObj"]->com.twosigma.beaker.jvm.object.SimpleEvaluationObject["payload"]->_run_closure1["delegate"]->script14926065475721000071615["binding"]->groovy.lang.Binding["variables"]->java.util.LinkedHashMap["beaker"]->com.twosigma.beaker.NamespaceClient["outp
```

konstantinTarletski:
test commented out:
com.twosigma.beaker.jupyter.msg.MessageCreatorTest#createMessageForCollection
line:68
com.twosigma.beaker.clojure.ClojureKernelTest#evaluate
line:58
Enable them !!!

Issue #5147, opened 2017-04-10
[clojure kernel] fix problem with result serialization (lazy sequence)


jaroslawmalekcodete:

scottdraves:
printing lazy sequence should work.
also, re-enable those tests.

scottdraves:
```
; lazy infinite sequence with recursive definition
(def fib-seq-lazy 
  ((fn rfib [a b] 
     (lazy-seq (cons a (rfib b (+ a b)))))
   0 1))
(take 20 fib-seq-lazy)
```
results in
```
clojure.lang.LazySeq@7c3704da
```
but it should should the actual sequence.

PR #5226, opened 2017-04-20
issue #5216 "infinite loop from returning function in groovy"

konstantinTarletski:
this is revert for issue #5147 "lazySeq"


scottdraves:
just disabling tests that fail is not correct.

konstantinTarletski:
This testd was added with change, which I deleted. I can delete this rests
too

20 апр. 2017 г. 18:09 пользователь "Scott Draves" <notifications@github.com>
написал:

Issue #5120, opened 2017-04-05
use jupyter/jvm-repr to control display types


scottdraves:
https://github.com/jupyter/jvm-repr

scottdraves:
the MIME classes and our widgets should be registered.

hadim:
Is that already possible to do something similar to that : 

```
  Object result = interpreter.eval(code);
  Map<String, String> resultByMIME = Displayers.display(result);
  Kernel.this.display(resultByMIME);
``` 
?

Can `this.seo.finished()` accept a MIMEType (`Map<String, String>`) ?

konstantinTarletski:
***Displayers.display***

We have similar functionality, but only with different class names and other small differences.

`Map<String, String> resultByMIME = Displayers.display(result);`
similar to 
`MIMEContainer  mimeResult = SerializeToString.doit(result)`

MIMEContainer  is holder for `mime` and `code`

so :
`Map<String, String> resultByMIME`
similar to 
`MIMEContainer  mimeResult`

Also this cannot be used to display widgets, because widgets is actually is Comm object.
And it cannot be represented with static HTML.

If our class `SerializeToString.doit` "detects" widget, it display it at once,
this part cannot be replaced by `MIMEContainer` :
```
    HashMap<String, Serializable> content = new HashMap<>();
    content.put(METHOD, DISPLAY);
    getComm().setData(content);
    getComm().send();
```

But in other cases it works same like `Displayers.display`
It  (`SerializeToString.doit`) can be renamed and extended, because now it works like so:
if it not widget it return result as text :
 MIME = "text/plain"

Exact code:
  ```
public static MIMEContainer doit(Object input) {
    MIMEContainer ret = null;
    if (input != null) {
      if (isWidget(input)) {
        DisplayAnyWidget.display(input);
        ret = HIDDEN();
      } else if (input instanceof MIMEContainer) {
        ret = (MIMEContainer) input;
      } else {
        ret = asString(input); // returns MIMEContainer with type: "text/plain"
      }
    } else {
      ret = Text("null");
    }
    return ret;
  }
```

***Returning multiple result***

Only diference with this code is that `Displayers.display` returns several variants,
but `SerializeToString.doit` just one.

Returning multiple result is good idea, also good idea to allow to users "register" new displayers.
But for now we do not have any functionality that allow to user chose way to display the result.
In this example no any suggestions was prowided in this direction:
```
Object result = interpreter.eval(code);
[...]

PR #5236, opened 2017-04-21
PR for issue #5120 "use jupyter/jvm-repr to control display types"

konstantinTarletski:
com.twosigma.beaker.groovy.evaluator.GroovyOutputContainerTest -- BROKEN

widgets do not display automatically !!!!

I can not control all thinks what I needed
also there is no "unregister" displayer

The main problem is that it always return default displayer.
and if I delete (unregister) it it crashes.
This makes me impossible to detect is there custom displayer or not.
And this prevents to display widgets automatically


konstantinTarletski:
If we decided to use this "jupyter/jvm-repr" as a external lib then we get this "problems":

 #
 **Default displayer.**
#

1)  There is default displayer, it can not be null (if set it to null, then displayer throws exception).

This means that code:
`Map<String, String> resultByMIME = Displayers.display(input);`
always will return something(resultByMIME can not be null or empty ).

 In current version default displayer in not wery good, it should be rewrited.
It makes problems like this :

Displayer :
```
jupyter.Displayers.register(String.class, new jupyter.Displayer() {

  public Map display(Object arg0) {
    java.util.Map<String, String> ret = new HashMap<>();
    ret.put("text/html", "HTML|" + arg0 + "|HTML");
    return ret;
  }
    
});
```

Code:
```
String[] abc = new String[2];
abc[0] = "FF"
abc[1] = "tt"
abc
```

Output:
`[null, null]`

To fix this, we need to register one more displayer (String[]) :
```
jupyter.Displayers.register(String[].class, new jupyter.Displayer() {

  public Map display(Object arg0) {
    java.util.Map<String, String> ret = new HashMap<>();
    ret.put("text/html", "HTML|" + arg0 + "|HTML");
    return ret;
  }
    
});
```

Code:
```
String[] abc = new String[2];
abc[0] = "FF"
abc[1] = "tt"
abc
```

Output:
`HTML|[FF, tt]|HTML`

2) Also, having "default displayer" do not "allow" to us determinate if there was assigned some displayer to class or not.
For example : 
Code :
```
import com.twosigma.beaker.widgets.floats.FloatSlider
def widget = new FloatSlider()
```

No any displayer was assigned.

Output:
`com.twosigma.beaker.widgets.floats.FloatSlider@428d1d33`

This is because "Displayers.display" returns output and I understand it like "custom displayer."

This is not a big problem, it can be fixed with our default displayer. But it should be done for all "not standart outputs"
 
To fix this problem, we need to write custom displayer for each widget , or "fix" this in default displayer

Displayer:
```
import com.twosigma.beaker.widgets.floats.FloatSlider

jupyter.Displayers.register(FloatSlider.class, new jupyter.Displayer() {

      @Override
      public Map display(Object arg0) {
        Map<String, String> ret = new HashMap<>();
        ret.put("application/vnd.jupyter.widget-view+json", "ABC");
        //ret.put("text/plain", "TEXT|" +arg0.value + "|TEXT");
        return ret;
      }
        
    });
```

Code:
```
import com.twosigma.beaker.widgets.floats.FloatSlider
def widget = new FloatSlider()
```

Output:
Will be widget.

#
**Multiple results**
#
First problem can be fixed, with custom default displayer.
But I do not know what to do with second one :

Displayer allow to user define multiple types :

For example (TEXT and HTML):
```
jupyter.Displayers.register(String.class, new jupyter.Displayer() {

  @Override
  public Map display(Object arg0) {
    java.util.Map<String, String> ret = new HashMap<>();
    ret.put("text/html", "HTML|" + arg0 + "|HTML");
    ret.put("text/plain", "TEXT|" + arg0 + "|TEXT");
    return ret;
  }
    
});
```

And in our application I do not know what to do with this:
our application to not allow to chose display type, so all I can do is just to take first one. Like this:

```
Map<String, String> resultByMIME = Displayers.display(input);
toDisplay = resultByMIME [0];
```

hadim:
Don't hesitate to submit PRs to jupyter/jvm-repr. I don't think a lot of people is using it at the moment. So it's us to decide what this lib should be.

konstantinTarletski:
I have a task to use this lib, and I write my "feeling" after I tried to use it.
I want to ask Scott to write same code without using this lib.
For my opinion, it not very "suit" for our application.
So I write this text mostly for Scott :)

PR #5243, opened 2017-04-22
Add .nb-gradle-properties to gitignore

hadim:
ping @scottdraves


scottdraves:
thanks for the suggestion -- not sure what this for though, can you explain?

hadim:
Just a Netbeans config file to configure custom profile (`publishToMavenLocal` for example).

PR #5244, opened 2017-04-22
Enable jitpack building

hadim:
According to https://jitpack.io/docs/BUILDING/, this PR should be sufficient to allow people to consume BeakerX dependency with Maven and/or Gradle.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,6 +15,7 @@
  */
 import org.apache.tools.ant.taskdefs.condition.Os
 
+
 task wrapper(type: Wrapper) {
   gradleVersion = '3.5'
 }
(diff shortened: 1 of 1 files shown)


hadim:
Well, jitpack can't build beakerx because of the pip dependency : https://jitpack.io/com/github/hadim/beakerx/jitpack-ildipo~beaker-in-bunsen2-g55ce5a3-2068/build.log

Once you have separated the `beakerx/` folder from the `kernel/` folder that should be possible to use jitpack.

hadim:
See here https://github.com/jitpack/jitpack.io/issues/1326

scottdraves:
thank you -- is there any reason not to merge this now?
when https://github.com/twosigma/beakerx/issues/5060 is done it will magically start working, right?

hadim:
If Beakerx Python module is removed from Gradle then is will magically works :-)

So yes feel free to merge.

PR #5248, opened 2017-04-24
fix column filter, replace debounce method

jszalek:


beakerx/js/src/tableDisplay/tableScope.js
```diff
@@ -1173,21 +1170,22 @@ define([
     }
     self.columnSearchActive = isSearch;
 
+
+    var filterDebounced = _.debounce(function(e, element) {
```
scottdraves: isn't this a dup? see below

beakerx/js/src/tableDisplay/tableScope.js
```diff
@@ -2547,14 +2545,20 @@ define([
     self.removeFilterListeners();
     var filterInputSelector = '.filterRow .filter-input';
     var clearFilterSelector = '.filterRow .clear-filter';
+
+
+    var debouncedFilter = _.debounce(function(e, element) {
```
scottdraves: dup of this.

PR #5254, opened 2017-04-26
tests for cpp module

EfimovVladimir:


EfimovVladimir:
For jacoco plugin to exclude auto generated classes:
on jenkins add on exclusion field 

`**/CPP14*.class`

Issue #5169, opened 2017-04-12
right-click and drag in plots broken


scottdraves:
it should allow user to drag out a rectangle (showing it shaded gray),
and on release, it should zoom the plot to that rectangle.
works in classic beaker, though there is bug there were it also opens the "save as" menu.

jszalek:
@scottdraves is there anything more to do here?

PR #5255, opened 2017-04-26
PR for issue #5169 "right-click and drag in plots broken"

konstantinTarletski:


scottdraves:
only works if i press both buttons at the same time.

Issue #5075, opened 2017-03-30
failed to install on Windows


altavir:
Using `gradle install` in `gradlex` gives a error.
I believe that the main problem is: `npm unavailable.  If you're running this command using sudo, make sure npm is available to sudo`
Npm is installed and accessible via `npm` command from the same console.

scottdraves:
sorry not sure what you mean.  we don't have an install target, and what is gradlex?

altavir:
Sorry, was writing in a hurry. I mean `beakerx` and it has `install` task:
```
task install() {
  doLast {
    exec {
      commandLine(*(winOrUnix('pip', "install", "-e", ".")))
    }
    exec {
      commandLine(*(winOrUnix("jupyter", "nbextension", "install", "--py", "--symlink", "--sys-prefix", "beakerx")))
    }
    exec {
      commandLine(*(winOrUnix("jupyter", "nbextension", "enable", "--py", "--sys-prefix", "beakerx")))
    }
    project('js').tasks.install.execute()
  }
}
```

altavir:
The same result if I use `linkBeakerNbextension` from root project

scottdraves:
can you provide a complete log?
it should work.

altavir:
Here is the log:
```
:beakerx:install
Obtaining file:///D:/beakerx/beakerx
    Complete output from command python setup.py egg_info:
    setup.py entered
[...]

PR #5256, opened 2017-04-26
PR for issue #5075 "failed to install on Windows"

konstantinTarletski:


scottdraves:
https://github.com/twosigma/beakerx/issues/5075

altavir:
There is another line with `check_call` in this file. I believe it also should be modified.

PR #5257, opened 2017-04-26
JS API for Plot

jszalek:
demoFiles/plot-js-api


scottdraves:
https://github.com/twosigma/beakerx/issues/5073

PR #5289, opened 2017-05-10
jarek/5119: Make Widget class base for all widgets

jaroslawmalekcodete:


scottdraves:
MapLikeTable_GroovyExample.ipynb has the DisplayAnyWidget problem.

PR #5290, opened 2017-05-10
#5169 fix zoom by right-click

jszalek:


scottdraves:
```
Uncaught TypeError: _.isMatch is not a function
    at PlotScope.zoomEnd (:8888/nbextensions/beakerx/index.js?v=20170510094230:3548)
    at SVGSVGElement.<anonymous> (:8888/nbextensions/beakerx/index.js?v=20170510094230:3667)
    at d.apply (:8888/nbextensions/beakerx/index.js?v=20170510094230:1753)
    at S (:8888/nbextensions/beakerx/index.js?v=20170510094230:1751)
    at c.emit (:8888/nbextensions/beakerx/index.js?v=20170510094230:1757)
    at c.end (:8888/nbextensions/beakerx/index.js?v=20170510094230:1757)
    at e (:8888/nbextensions/beakerx/index.js?v=20170510094230:1757)
    at :8888/nbextensions/beakerx/index.js?v=20170510094230:1751
```
on release of right mouse

scottdraves:
```
_.VERSION
"1.5.2"
```
should be 1.8.3

scottdraves:
I have to assume we are getting _ from jupyter and we have different versions of jupyter.   my current env has 4.3 and I'll bet you have 5.  I will just upgrade and this should go away, I don't think we have to support old versions.

But if we are just picking it up from jupyter then it should not be in our package.json so there is a bug to fix here.

jszalek:
@scottdraves I have 1.8.3 version.

PR #5292, opened 2017-05-10
treemap

fishera123:
implements #5160


beakerx/js/src/plot/plotScope.js
```diff
@@ -2417,7 +2417,7 @@ define([
       self.model.model = data;
     }
 
-    // self.element = $('div#'+self.wrapperId+' .dtcontainer');
+    self.element = $('div#'+self.wrapperId+' .dtcontainer');
```
scottdraves: @jszalek is this OK?

fishera123: we can put it inside if statement to be sure that it only assigns when there is treemap

jszalek: @scottdraves it is not ok, this line should be removed.
@fishera123 you can change order of method execution in Plot.js (64,65 lines) file
set the element 
`currentScope.setElement(tmplElement.children('.dtcontainer'));`
before setting a model
`currentScope.setModelData(data);`

Issue #5287, opened 2017-05-09
Java imports need to be further adjusted for Scala


jpallas:
Since the Scala kernel is using the default JVM imports, two adjustments are needed beyond the already implemented changing of the wildcard.  They are:

1. Remove `static`
2. Escape the package name `object`, which is a Scala keyword.  (This is used in one of the default import paths, `com.twosigma.beaker.jvm.object.OutputCell`.)

On the second point, it's conceivable that other Scala keywords that aren't Java keywords might show up in package names, but no others seem very likely, so a one-off replacement is probably fine.

scottdraves:
Thank you Joe.  Agreed.

jpallas:
I've got some code for this in my workspace.  Working on a unit test.

PR #5295, opened 2017-05-10
#5287 Adjust scala imports

jpallas:
The unit test is a little indirect, because failed imports log a message but don't otherwise change the visible state of the evaluator.


jpallas:
Not sure why this didn't auto link to issue #5287.  Github documentation on auto-linking is hard to locate.

scottdraves:
cool thx will review shortly.

scottdraves:
🐬

Issue #5305, opened 2017-05-12
problems with handling of magic commands


jpallas:
There seem to be a couple of problems with magic commands.  Currently, output from the first magic command appears in the notebook, but the cell is never marked as completed and subsequent executions are blocked.

In `ExecuteRequestHandler.handleMsg`, the non-magic case passes a callback lambda to `executeCode` that sends an idle message and releases the sync object.  Since the magic command path doesn't use queue-with-callback model, it seems like these two operations are needed inline.

Also, in `MessageCreator.createMagicMessage`, the reply-without-status message is being "published" (to the I/O socket) instead of being "sent" (to the shell socket).

I have these changes ready for a PR.

In the broader picture, there something inconsistent about `createMagicMessage` and the control flow.  Its peer `createMessage` returns a list of `MessageHolder`s, while `createMagicMessage` actually dispatches the messages instead of returning them.  Maybe that was because of the callback/no-callback difference, but it's not ideal for code comprehension.  On the other hand, the processing of a list of `MessageHolder`s is inside `ExecutionResultSender` and not exposed, so there would need to be either some duplication of functionality or non-trivial refactoring to clean that up.

scottdraves:
Thanks for the report.  The message handling was recently changed for Juptyer Lab and I think the magics were not updated to match.

autotranslation_groovy_demo.ipynb is a good way to see the problem, even the %%html and %%javascript cells hang.

PR #5306, opened 2017-05-12
#5305 Fix messaging issues with magic commands

jpallas:
Changes described in #5305 to address magic command messaging.


scottdraves:
awesome!

Issue #5304, opened 2017-05-12
table menu "copy to clipboard" broken


scottdraves:
when i paste into excel i get
<img width="460" alt="screen shot 2017-05-12 at 12 32 11 pm" src="https://cloud.githubusercontent.com/assets/963093/26007571/19f7cd10-370f-11e7-9415-4f973f28c57e.png">
instead of the selected content.  works in beaker classic.

PR #5313, opened 2017-05-16
#5304: fix table menu copy to clipboard

MariuszJurowicz:


scottdraves:
fixes https://github.com/twosigma/beakerx/issues/5304

Issue #5307, opened 2017-05-13
update groovy autotranslation demo to use displayHTML


scottdraves:
like autotranslation_python

scottdraves:
proposal results in no output at all for the javascript cell with the d3 calls.
no error msgs.  what's wrong?

PR #5323, opened 2017-05-17
#5307: Fixed executing %%javascript from groovy kernel.

michalgce:
https://github.com/twosigma/beakerx/issues/5307


kernel/base/src/main/java/com/twosigma/beaker/mimetype/MIMEContainer.java
```diff
@@ -79,6 +79,10 @@ public static MIMEContainer HIDDEN() {
     return addMimeType(MIME.HIDDEN);
   }
 
+public static MIMEContainer JavaScript(Object code) {
```
scottdraves: indentation is off

scottdraves:
need to update the ipynb to use displayHTML in the JS now that it should work.

michalgce:
Added update to ipynb

Issue #5311, opened 2017-05-15
hover over column header should show type


scottdraves:
in tables, the type of the column should appear as a tooltip on hover.

PR #5346, opened 2017-05-19
#5311 hover over column header should show tooltip

MariuszJurowicz:


scottdraves:
cool, this works but we need the following improvements:
1) show the tooltip above the mouse, not below.
2) wait about 0.3 seconds before showing it and only show it if the mouse remains over the column.

scottdraves:
plus a bug, after playing with the table menus for a while i got some leftovers:
screen shot 2017-05-19 at 10 41 33 pm
screen shot 2017-05-19 at 10 41 11 pm

scottdraves:
```
tooltip.js:223 Uncaught TypeError: Cannot read property 'nodeType' of undefined
    at $.(anonymous function).(anonymous function)._updateContent (http://localhost:8888/nbextensions/beakerx/index.js?v=20170522111922:90438:59)
    at $.(anonymous function).(anonymous function)._updateContent (http://localhost:8888/nbextensions/beakerx/index.js?v=20170522111922:48923:26)
```
i got this after i hit the "1" keyboard command.

scottdraves:
thx the changes for these items look good:

show the tooltip above the mouse, not below.
wait about 0.3 seconds before showing it and only show it if the mouse remains over the column.

MariuszJurowicz:
I've change the way the tooltip is initialized to delegate down from `document.body` rather then initializing on header cell.

Issue #5317, opened 2017-05-16
remove antlr4 from kernel/base


scottdraves:
related: https://github.com/twosigma/beakerx/issues/5182

PR #5355, opened 2017-05-21
#5317 Removed ANTLR4 from kernel/base.

michalgce:


scottdraves:
every file needs banner

kernel/groovy/src/main/java/com/twosigma/beaker/groovy/autocomplete/AutocompleteGroovyResult.java
```diff
@@ -0,0 +1,16 @@
+package com.twosigma.beaker.groovy.autocomplete;
```
scottdraves, line 17: every file needs banner

michalgce: 👍

Issue #5041, opened 2017-03-30
make java side of ipywidgets implementation complete


scottdraves:
we got IntSlider (yea!) now fill in the rest, and handle every attribution.

konstantinTarletski:
useful links:
https://github.com/jupyter-widgets/ipywidgets
http://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html

unimplemented 20 items from 42:

Accordion			--- READY
BoundedFloatText	--- READY
BoundedIntText	--- READY
IntText			--- READY
ButtonStyle		---SKIPPED
**Represents a game controller**
Controller			---SKIPPED
**Represents a gamepad or joystick axis.**
ControllerAxis		---SKIPPED
**Represents a gamepad or joystick button.**
ControllerButton	---SKIPPED
FloatRangeSlider	--- READY
FloatText			--- READY
HTMLMath		--- READY
IntRangeSlider		--- READY
Play				--- READY
ProgressStyle		---SKIPPED
SelectionSlider		--- READY
SliderStyle			---SKIPPED
ToggleButtons		--- READY
Valid				--- READY
DirectionalLink		--- READY
Link				--- READY


**Improvements:**

In "ValueWidget" added methods like "getInteger(Object input)" 
It allow to use widgets more flexible, not only with certain types.
Example : 
```
import com.twosigma.beaker.widgets.integers.IntRangeSlider
w = new IntRangeSlider()
w.value = ["10",49.6]
w.orientation = "horizontal"
//w.orientation = "vertical"
display(w)
```
As You can see it is "Int" slider.
but I put to value String and Float and it is automaticlly converted to "Int"

or even:
`w.value = ["10",[49.6]]`
also to value can be passed not array but `java.util.Collection`:

```
import com.twosigma.beaker.widgets.integers.IntRangeSlider
import java.util.Collection

Collection<Object> list = new ArrayList<>()
list.add(["19"])
list.add("53")

w = new IntRangeSlider()
w.value = list
w.orientation = "horizontal"
//w.orientation = "vertical"
display(w)
```
This conversion works not only with "Int" but also with "String", "Double" (Float) and Boolean
Also it works not only with `value` but with other properties like `min` , `max` , `step`

soerenklingner:
How do i observe a IntSlider widget? When the slider value changes i want a function to be called.

Where can i find a good documentation of the beakerx java?

PR #5363, opened 2017-05-22
issue #5041 "make java side of ipywidgets implementation complete" 

konstantinTarletski:
Widgets:
DirectionalLink
Link

### DIFF
--- a/kernel/base/src/main/java/com/twosigma/beaker/widgets/link/DirectionalLink.java
+++ b/kernel/base/src/main/java/com/twosigma/beaker/widgets/link/DirectionalLink.java
@@ -1,3 +1,18 @@
+/*
+ *  Copyright 2017 TWO SIGMA OPEN SOURCE, LLC
+ *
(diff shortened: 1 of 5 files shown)


kernel/base/src/main/java/com/twosigma/beaker/widgets/link/DirectionalLink.java
```diff
@@ -0,0 +1,32 @@
+package com.twosigma.beaker.widgets.link;
```
scottdraves, line 16: all files need banner

kernel/base/src/main/java/com/twosigma/beaker/widgets/link/Link.java
```diff
@@ -0,0 +1,72 @@
+package com.twosigma.beaker.widgets.link;
```
scottdraves, line 16: all files need banner

konstantinTarletski:
Fixed

Issue #5356, opened 2017-05-22
Autocomplete in scala and clojure use wrong insertion point


jpallas:
Scala autocomplete doesn't keep track of multiple lines.  Both scala and clojure autocomplete always set the replacement index to the end of the input, rather than the beginning of the input being completed.  That makes, for example, the completion of `toL` as `toList` come out as `toLtoList`.

PR available shortly.

PR #5373, opened 2017-05-22
5356 Fix scala and clojure autocompletion

jpallas:
Proposed fixes for #5356.

I think this fixes the most egregious problems.  The clojure completer is very broad.  I'm not a clojure user so I don't know if this is desirable.  The scala completer has a feature where it returns type information when you ask it to complete an already complete method name again (automatically raises the verbosity).  That's useful information but not actually a valid replacement, so maybe it should be suppressed or displayed by some other means.


scottdraves:
🐬

Issue #5388, opened 2017-05-24
enter key in easyform textfield should press button


scottdraves:
This is a regression from classic Beaker.

If there is a button in the form, and the cursor is in a text field (not a text area), editable combo box, or a date field, then when you press the "enter" key, it should act the same as if the button was clicked.

PR #5391, opened 2017-05-26
#5388 enter key in easyform textfield should press button

MariuszJurowicz:
There is one more issue with datepicker (flatpickr):
Flatpickr consumes the Enter button `keypress` event and it seems that https://github.com/chmln/flatpickr/issues/355 solves the issue only for `keydown`.
@scottdraves  is it OK to use the `keydown` instead of the `keypress` here?
Or maybe we should use different datepicker library eg. jquery-ui?

Second thing is if we should move the `Text` implementations as it uses the `beakerx` module instead the `jupyter-js-widget`. That is to disable the text widget preventing the Enter `keypress` event delegation.


beakerx/js/src/EasyForm.js
```diff
@@ -14,13 +14,22 @@
...
 var widgets = require('jupyter-js-widgets');
 var _ = require('underscore');
 
 var buttonWidget = require('./easyForm/buttonWidget');
 var selectMultipleWidget = require('./easyForm/selectMultipleWidget');
 var selectMultipleSingleWidget = require('./easyForm/selectMultipleSingleWidget');
 var datePickerWidget = require('./easyForm/datePickerWidget');
+var textWidget = require('./easyForm/textWidget');
+
+widgets.TextView.prototype.handleEnterKeyPress = function() {
```
scottdraves: we don't want to do this if the text widget is not part of easyform.

scottdraves:
i'm getting this:
screen shot 2017-05-30 at 9 29 22 am

scottdraves:
oh no it's in master

Issue #5316, opened 2017-05-16
add table column menu item: reset formatting


scottdraves:
add to the bottom of the table column menu "Reset Formatting" to restore the column to its original state.  This is similar to the "Reset all interactions" menu item for the whole table.

![screen shot 2017-05-16 at 2 00 17 pm](https://cloud.githubusercontent.com/assets/963093/26120863/3327c464-3a40-11e7-86fb-69b07ce00231.png)

scottdraves:
more specifically, the following state should be reset:
Filter, Search, Format, Sort, Align, heatmap, bars, Fix left/right, and the width of the column (you can drag to change it).

The following would not be changed: show/hide and order (drag to reorder the column).

PR #5396, opened 2017-05-29
#5316 add column reset formatting menu item

MariuszJurowicz:
This also fixes wrong column heatmap bug after col reorder;


beakerx/js/src/tableDisplay/tableScope.js
```diff
@@ -1611,7 +1640,7 @@ define([
       'processing': true,
       'autoWidth': true,
       'ordering': true,
-      'order': self.tableOrder ? _.cloneDeep(self.tableOrder) : [],
+      'order': self.tableOrder ? _.clone(self.tableOrder) : [],
```
MariuszJurowicz, line 1643: Underscore does not have method `cloneDeep`. It's the `lodash` method perhaps.

Issue #5399, opened 2017-05-29
 js build is broken


scottdraves:
a fresh clone fails to build:
```
$ gradle :beakerx:install
:beakerx:install
yarn install v0.24.5
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
warning "sass-loader@6.0.5" has incorrect peer dependency "webpack@^2.0.0".
warning "watch-ignore-webpack-plugin@1.0.0" has incorrect peer dependency "webpack@1.5.x".
[4/4] Building fresh packages...
error /Users/spot/tmp/beakerx/beakerx/js/node_modules/jupyter-js-widgets: Command failed.
Exit code: 127
Command: sh
Arguments: -c npm dedupe
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Directory: /Users/spot/tmp/beakerx/beakerx/js/node_modules/jupyter-js-widgets
Output:
sh: npm: command not found
:beakerx:install FAILED
```

PR #5400, opened 2017-05-29
#5399 add npm back to app dependencies

MariuszJurowicz:


scottdraves:
thx but i changed my mind on this.
if the user installs yarn they should get node and npm.
i think i only had a problem because something about my brew install was broken, i cleaned that out and reinstalled node and everything seems good now.

Issue #5386, opened 2017-05-24
too many columns


scottdraves:
https://github.com/twosigma/beakerx/blob/master/beakerx/js/src/tableDisplay/table.html#L3

you can see there is a commented out warning message for too many columns.
bring back this feature from classic beaker.

PR #5411, opened 2017-05-31
#5386 add too many columns modal

MariuszJurowicz:


scottdraves:
This looks good.  can you please add a demo notebook that creates a table that triggers this msg?
Also include a notebook with too many rows (>10k) in this notebook.

scottdraves:
does not seem to work, i don't get any warning from the demo notebook, though all but 50 are correctly hidden.

also the demo code should use LinkedHashMap not HashMap, to keep them in order.

MariuszJurowicz:
@scottdraves my bad, i've used wrong column limit variable. Seems to be fixed now.

Issue #5071, opened 2017-03-30
make a scala demo notebook 


scottdraves:
including widgets
![screen shot 2017-03-17 at 11 07 16 am](https://cloud.githubusercontent.com/assets/963093/24049333/1155e946-0b02-11e7-99b7-a32aee488aad.png)

scottdraves:
basic notebook added, needs widgets: easyform, plots, tables, etc.

see the java widgets demo.

scottdraves:
please continue with this:
tables, output containers, java widgets.

also, move all scala tutorials into their own subdirectory of demoFiles.

jpallas:
The recent commit for this issue moved Scala demos into `demoFiles/scala`.  A couple of examples broke because they look for data files (`.csv`, `.png`, etc.) in the current directory, but those are now in the parent directory.  That makes Scala an exception: none of the other kernels have a separate subdirectory for demos.  Either approach seems fine, but consistency would make it easier to compare or translate demos between languages.  Is there a plan to reconcile this?

scottdraves:
thx for the report, agreed, see linked issue.

PR #5415, opened 2017-05-31
#5071 - Added Easyform scala API and demo.

michalgce:


kernel/scala/src/main/java/com/twosigma/beaker/scala/easyform/EasyForm.scala
```diff
@@ -0,0 +1,111 @@
+package com.twosigma.beaker.scala.easyform
```
scottdraves, line 17: all files need banner

scottdraves:
thanks, good work, looks like the right idea, but I didn't see how to access the values in the form.
can you add a cell to be run that works?  i tried this:
screen shot 2017-05-31 at 11 27 36 am

kernel/scala/src/main/java/com/twosigma/beaker/scala/easyform/EasyForm.scala
michalgce: 👍

kernel/base/src/main/java/com/twosigma/beaker/chart/xychart/plotitem/XYGraphics.java
```diff
@@ -92,7 +92,7 @@ public void setX(List<Object> xs) {
   }
 
   public List<Number> getX() {
-    if (xs == null) {
+    if (xs == null || xs.isEmpty()) {
```
scottdraves, line 95: why was this added?  doesn't seem right.  an empty xs array seems like an error.

scottdraves:
🐬

Issue #5410, opened 2017-05-30
make selection of rows in table be a sub-rectangle


scottdraves:
right now you can have any subset of the rows selected.

instead, you should be able to select any contiguous subrectangle.
if you click and drag, then it should select a rectangle.
if you shift-click, it should extend the rectangle from from the current cell.

remove the "reverse selection" and "select all" menu items.
rename "deselect all" to "clear selection".

PR #5417, opened 2017-06-01
#5410 sub-rectangle selection

MariuszJurowicz:
@scottdraves is it ok to use ctrl key instead of shift to add more rows to selection?
The ctrl key is used by the `jquery-ui` `selectable` lib and it's not configurable.
As we are using the `jquery-ui` lib everywhere i opt for leaving the `selectable` lib behavior as is and use the ctrl key here.


scottdraves:
sorry, i didn't explain this clearly enough.
i am seeing the drag-rectangle, but on release it is just selecting rows.

the change we need is that the selection can be a *rectangle*, including not all the columns.
and skip the dotted-line rectangle.  indicate while dragging what the selection will be just by showing it.

here's a faked screenshot of what it might look like:
screen shot 2017-06-01 at 2 04 05 pm

scottdraves:
also, this should not remove the brighter blue current cell outline, which you can move with the arrow keys.

scottdraves:
click and drag looks good, but just a click still selects a row
screen shot 2017-06-08 at 10 48 44 am

MariuszJurowicz:
@scottdraves 
I removed the click selection but the `dbclick` row selection is still there. Should i remove `dbclick` row selection as well?

scottdraves:
thanks. double click should also be removed.

also, bigger problem, they keyboard commands seem to be broken.  if i click on a cell and type "1" it should change the precision of the columns, but instead the event is going to jupyter and it is changing the cell into markdown and making it a heading.

scottdraves:
also we have a small problem, now that dragging is selecting a rectangle, it is now impossible to select the text to get it into the clipboard as regular text (not csv).  let me think about that.

scottdraves:
OK can you please add a menu item to the table menu that says "Copy to clipboard as text"
that works like copy to CSV except it omits the column headers.

PR #5440, opened 2017-06-07
add test module to use e2e tests

EfimovVladimir:


scottdraves:
can you set this up to use yarn instead of npm?

EfimovVladimir:
Changed to use yarn instead npm.
Changed to use Webdriver.io instead Protractor.

test/runner.py
```diff
@@ -0,0 +1,62 @@
...
+import os
+import sys
+import subprocess
+import signal
+
+here = os.path.abspath(os.path.dirname(__file__))
+beakerx_dir = os.path.abspath(os.path.join(here, ".."))
+test_dir = here
+
+# update environment
```
LeeTZ: We should get rid of Windows line end char '^M' here

test/tests/beakerx.po.js
```diff
@@ -0,0 +1,35 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+var BeakerXPageObject = function() {
+
+    this.baseURL = 'http://127.0.0.1:8888/tree/demoFiles';
```
LeeTZ: *.js files should have two-space indentation.

test/wdio.conf.js
```diff
@@ -0,0 +1,247 @@
+exports.config = {
```
LeeTZ, line 17: should have a copyright banner here.

LeeTZ:
For the integration test It works well!

Issue #5107, opened 2017-04-04
help links in kernel info


scottdraves:
BeakerX Home -> https://github.com/twosigma/beakerx
File an Issue -> https://github.com/twosigma/beakerx/issues/new
Two Sigma Open Source -> http://opensource.twosigma.com/

https://github.com/twosigma/beakerx/blob/395c8ae34fac60492c659faf65adf2ff109afad7/kernel/base/src/main/java/com/twosigma/beaker/KernelInfoHandler.java#L67

PR #5446, opened 2017-06-08
#5107 Added Help Links for kernels.

michalgce:


kernel/base/src/main/java/com/twosigma/beaker/jvm/object/HelpLink.java
```diff
@@ -0,0 +1,36 @@
+/*
```
scottdraves: just use a helper function in KernelInfoHandler.java, this class it too lightweight.

michalgce: In my opinion this solution is more readable creating hashmap for each text and url, furthermore i think this solution allows us to reuse HelpLink in any other Kernels if we would.

What you think?

Issue #5437, opened 2017-06-07
Unable to save as png or svg specific plot


jszalek:
plot code:
```
def points = 100;
def logBase = 10;
def expys = [];
def xs = [];
for(int i = 0; i < points; i++){
  xs[i] = i / 15.0;
  expys[i] = Math.exp(xs[i]); 
}

def test =new Plot(title: "Linear x, Log y", yLabel: "Log", logY: true, yLogBase: logBase);
test << new Line(x: xs, y: expys, displayName: "f(x) = exp(x)");
test << new Line(x: xs, y: xs, displayName: "g(x) = x");
```

scottdraves:
i cannot reproduce

scottdraves:
still can't reproduce

PR #5454, opened 2017-06-09
#5437 replace &nbps with space in 'save as' method

jszalek:


scottdraves:
closing until i understand what can cause this

Issue #5441, opened 2017-06-07
error in cell run by plot action goes to wrong location


scottdraves:
open plotActions.ipynb and run just the 3rd cell, then click on the line.  the plot vanishes and the following error appears on the console:

```
Exception in thread "Thread-0" groovy.lang.MissingPropertyException: No such property: abc for class: script1496843416196163566422
        at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.unwrap(ScriptBytecodeAdapter.java:51)
        at org.codehaus.groovy.runtime.callsite.PogoGetPropertySite.getProperty(PogoGetPropertySite.java:49)
        at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callGroovyObjectGetProperty(AbstractCallSite.java:304)
        at script1496843416196163566422$_run_closure1.doCall(script1496843416196163566422.groovy:6)
...
```
and the kernel is stuck and cannot evaluate any more cells (any cell will hang).

1) the error should appear on the cell output, not on the console.
2) the kernel should not get stuck.

konstantinTarletski:
The same problem was in "inractive" , I fixed it.
Inractive works fine now, but not table/plot API.
Now kernel not get stuck at all casees.
But there is the problem with displaying something in console (table/plot API). 
The problem is "parrent message": we need to pass correct parent message, to get ability print anything to the jupyter console.
I afraid it is not simple task.
Need to check what message passed in "interactive" and pass same in plot/table API.

scottdraves:
👍 cool does not get stuck.
what's the difference between errors for tables and plots though?
i am seeing the ones for tables correctly but not plots.

konstantinTarletski:
TableDisplay.setDoubleClickAction() works so : 
if Yoou put tag, it deletes handler
if You put handler, it deletes tag
this behavior was taken from "old beaker"

```
  public void setDoubleClickAction(Object listener) {
    this.doubleClickTag = null;
    this.doubleClickListener = listener;
  }

  public void setDoubleClickAction(String tagName) {
    this.doubleClickListener = null;
    this.doubleClickTag = tagName;
  }
```

About same happens in 
`TableDisplay.addContextMenuItem()`

In Plots it is possible to assign muptiple actions to `Graphics`
Graphics --- Base object, Plot line for example

Now "update" works for Table/Plot API

konstantinTarletski:
Now Table/Plot API supports any code in handlers :

[...]

PR #5462, opened 2017-06-12
PR for issue #5441 "error in cell run by plot action goes to wrong location"

konstantinTarletski:
The same problem was in "inractive" , I fixed it.
Inractive works fine now, but not table/plot API.
Now kernel not get stuck at all casees.
But there is the problem with displaying something in console (table/plot API).
The problem is "parrent message": we need to pass correct parent message, to get ability print anything to the jupyter console.
I afraid it is not simple task.
Need to check what message passed in "interactive" and pass same in plot/table API.


scottdraves:
👍  cool does not get stuck.
what's the difference between errors for tables and plots though?
i am seeing the ones for tables correctly but not plots.

PR #5467, opened 2017-06-13
PR for task "Update for Table/Plot API"

konstantinTarletski:
Now update automatically works for this API actions:

Graphics  --- Base object, Plot line for example

Graphics.onClick()
Graphics.onKey()
TableDisplay.setDoubleClickAction()
TableDisplay.addContextMenuItem()

TableDisplay.setDoubleClickAction() works so :

if Yoou put tag, it deletes handler
if You put handler, it deletes tag
this behavior was taken from "old beaker"
```
  public void setDoubleClickAction(Object listener) {
    this.doubleClickTag = null;
    this.doubleClickListener = listener;
  }

  public void setDoubleClickAction(String tagName) {
    this.doubleClickListener = null;
    this.doubleClickTag = tagName;
  }
```
About same happens in
TableDisplay.addContextMenuItem()

In Line it is possible to assign muptiple actions to Graphics


kernel/base/src/main/java/com/twosigma/beaker/chart/Graphics.java
```diff
@@ -16,28 +16,23 @@
...
 import java.io.Serializable;
-import java.io.StringWriter;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 
 public abstract class Graphics implements Serializable, Cloneable{
+
+  private static final long serialVersionUID = -1878979081955090695L;
```
scottdraves, line 34: what is this?

konstantinTarletski: This is Java standart : every "Serializable" should have unicue number

altavir: Serial UID is not necessary but it prevents deserializing object serialiazed with different version of the code.

Issue #5421, opened 2017-06-01
Regenerate our Dockerfile


scottdraves:
We used to have this: https://github.com/twosigma/beaker-notebook-archive/blob/1.7.1/Dockerfile
get it working again.

Use a base image like we did before:
https://github.com/twosigma/beaker-notebook-archive/blob/1.7.1/base_docker_image/Dockerfile

scottdraves:
base it on ubuntu 16.04 LTS

prskr:
Hi,

I created a Dockerfile of beakerx [https://github.com/baez90/beakerx-docker](https://github.com/baez90/beakerx-docker).
I took the scripts from [https://github.com/jupyter/docker-stacks](https://github.com/jupyter/docker-stacks) and adopted them to beakerx.

I'm not an docker expert so it might be necessary to do some optimization but it might help you and it's working for me.

PR #5494, opened 2017-06-16
#5421 Regenerate DockerFiles.

michalgce:


scottdraves:
cool.  put all these into a subdirectory "docker", except the base image, put that in "docker/base".

scottdraves:
👍 please update to master and support openjdk from conda.

docker/base/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64
+
+RUN apt-get update && apt-get dist-upgrade -y
+RUN apt-get install -y sudo curl unzip software-properties-common apt-transport-https git clang bzip2 wget
+RUN apt-get update
+
+##########
+#  Java  #
+##########
+RUN apt-get install -y openjdk-8-jre
```
scottdraves: install this with conda not apt-get.

docker/setup.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+cd /home/beakerx/beakerx
+
+conda create -y -n beakerx python=3.5 jupyter pandas
+source activate beakerx
+./gradlew --no-daemon build
```
scottdraves: --no-daemon not needed.
gradlew not needed since gradle was installed directly.
either remove install of gradle or use it.

docker/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM beakerx-base:latest
+
+MAINTAINER Beaker Feedback <beaker-feedback@twosigma.com>
+
+ENV PATH /opt/conda/bin:/opt/gradle/gradle-3.5/bin:$PATH
```
scottdraves: /opt/gradle/gradle-3.5/bin
delete

docker/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+ENV SHELL /bin/bash
+ENV NB_USER beakerx
+ENV NB_UID 1000
+ENV HOME /home/$NB_USER
+ENV LC_ALL en_US.UTF-8
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US.UTF-8
+
+COPY setup.sh /home/beakerx
+RUN chmod +x /home/beakerx/setup.sh
```
scottdraves: just chmod+x on the original in github (better yet eliminate it, see below comment)

docker/base/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+
+MAINTAINER Beaker Feedback <beaker-feedback@twosigma.com>
+
+RUN useradd beakerx --create-home
+
+ENV JAVA_HOME /opt/conda/bin/java
+ENV PATH /opt/conda/bin:$PATH
+
+RUN apt-get update && apt-get dist-upgrade -y
+RUN apt-get install -y sudo curl unzip software-properties-common apt-transport-https git clang bzip2 wget
```
scottdraves: we don't need clang

docker/setup.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
scottdraves, line 1: why are these commands in their own script instead of just RUN from the Dockerfile?
delete this script if you can.

docker/base/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM ubuntu:16.04
+
+MAINTAINER Beaker Feedback <beaker-feedback@twosigma.com>
+
+RUN useradd beakerx --create-home
+
+ENV JAVA_HOME /opt/conda/bin/java
```
scottdraves: is this really needed?  i think the "source activate beakerx" is enough to use java from conda.

docker/base/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
+echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
+apt-get update && apt-get install yarn -y && \
+apt-get clean
+
+##################
+#    MiniConda   #
+##################
+RUN echo 'export PATH=/opt/conda/bin:$PATH' > /etc/profile.d/conda.sh && \
+    wget --quiet https://repo.continuum.io/miniconda/Miniconda2-4.3.21-Linux-x86_64.sh -O ~/miniconda.sh && \
```
scottdraves: use Miniconda3

scottdraves:
add a README.md with the commands used to build the containers.

docker/setup.sh
michalgce: it need to be separate because we need to use ```source activate beakerx``` which is only avaiable on /bin/bash shell

scottdraves: OK makes sense

docker/Dockerfile
scottdraves: still need this

scottdraves:
still need README

Issue #5456, opened 2017-06-09
add table menu item


scottdraves:
after https://github.com/twosigma/beakerx/pull/5417 goes in, it will be impossible to select the text to get it into the clipboard as regular text (not csv).

please add a menu item to the table menu that says "Copy to clipboard as text"
that works like copy as CSV except it omits the column headers.

rename the current item to "Copy to clipboard as CSV"

PR #5504, opened 2017-06-19
#5456 add copy as text menu item

MariuszJurowicz:


scottdraves:
cool but please title it "Copy to Clipboard as Text" (cap T)

and do not include a newline at the end.

scottdraves:
the goal is to select a single cell with a click, then "Copy to Clipboard as Text" and then paste in another window and get just a number like `42.3` without any spaces or newlines.

scottdraves:
actually, just merge this menu item with "Copy to Clipboard as CSV".
there should be just one menu item "Copy to Clipboard", and if
exactly 1 cell is selected, then it does "as Text", and if more then one (a drag selection of a rectangle) then as CSV.

Issue #5482, opened 2017-06-14
value returned by handlers should not be displayed


scottdraves:
followup to https://github.com/twosigma/beakerx/issues/5441

1) stdout should always go to the cell output, not to the console
2) a value returned by a handler should not be displayed, only errors and stdout.

scottdraves:
example of 1:
```
def random = new Random()
def p = new Plot(showLegend: true, useToolTip: false);
p << new Line(x: [1, 2, 3], y: [2, 3, 4], width: 10, displayName: "line 1")

//Plot update not ready, so this is executed but not displayed 
.onClick({info -> info.graphics.displayName = "new name"; println("xyz")})


p << new Line(x: [1, 2, 3], y: [5, 6, 7,], width: 10, displayName: "line 2")

//Plot update not ready, so this is executed but not displayed 
.onClick({info -> info.graphics.y[1] = random.nextInt(10)})
```
the xyz from this println should appear in the cell output, not on the console.

scottdraves:
example of 2:
```
import com.twosigma.beaker.table.*
def mapList4 = [[a:1, b:2, c:3],[a:4, b:5, c:6],[a:7, b:8, c:5]]
def display7 = new TableDisplay(mapList4)
display7.setDoubleClickAction { row, col, tableDisplay ->
   tableDisplay.values[row][col] = tableDisplay.values[row].sum()
}
display7
```
then double click in the table.  an extra value incorrectly appears below the table:
![screen shot 2017-06-14 at 2 12 35 pm](https://user-images.githubusercontent.com/963093/27147658-8d0f3908-510b-11e7-9b4d-79654a30cece.png)

scottdraves:
point 2 is fixed.

still need point 1.

PR #5506, opened 2017-06-19
PR fr issue #5482 "value returned by handlers should not be displayed"

konstantinTarletski:


scottdraves:
Point 1 does not seem to be addressed.

Ran the code for point 2 and got this error from the server:
```
[E 14:05:57.988 NotebookApp] Notebook JSON is invalid: None is not valid under any of the given schemas

    Failed validating 'oneOf' in display_data['properties']['data']['additionalProperties']:

    On instance['cells'][0]['outputs'][0]['data']['x-beakerx/empty']:
    None
```
to get this error you have to wait for autosave, or hit save explicitly.

scottdraves:
in the code for point 2, errors in the closure are handled correctly, but println is going to the console.  it should go to cell output area.

Issue #5508, opened 2017-06-19
Scala evaluator gives no feedback for some malformed input


jpallas:
The Scala interpreter has three return cases: success, error, and "incomplete".  If the input has unclosed braces, parentheses, or quotes, it is incomplete.  Right now that case gives no feedback to the user, because it is handled the same as an error but there is no error message.

Coming PR will distinguish all three cases and generate an "incomplete input" message when appropriate.

scottdraves:
👍  many thanks

PR #5509, opened 2017-06-19
#5508 Generate useful error output when the input is incomplete

jpallas:
This addresses #5508 by producing an error message for the interpreter's `Incomplete` return state.


scottdraves:
cool thx will try this out after lunch

Issue #5060, opened 2017-03-30
redo build system


scottdraves:
perhaps it should be based on python?

at least we need a way to
1) install all kernels with target and
2) a regular build should do both the JS and the Java.

scottdraves:
~~top level command should be `pip install -e .`~~

as described in https://github.com/twosigma/beakerx/issues/5476, we should do it like jupyter lab,
which means we run yarn and gradle in the recipe (and as documented in the readme and docker container for non-conda users).  `pip install -e .` just does the python stuff.

the gradle part is https://github.com/twosigma/beakerx/issues/5141

scottdraves:
part of https://github.com/twosigma/beakerx/issues/5134

Issue #5476, opened 2017-06-14
recipe in repo for conda forge


scottdraves:
https://github.com/twosigma/staged-recipes/blob/master/recipes/beakerx/meta.yaml

part of https://github.com/twosigma/beakerx/issues/5134

looks like we need to do a complete build and upload all artifacts (webpacks and jars) to pypi,
and then the recipe gets a tarball from there to make the conda package.

here's how Jupyter Lab does it (we'll have jars too):
https://github.com/jupyterlab/jupyterlab/blob/master/RELEASE.md

scottdraves:
as per https://github.com/twosigma/beakerx/issues/5476
we only need runtime dependencies in conda.

scottdraves:
https://github.com/twosigma/beakerx/issues/5576

PR #5513, opened 2017-06-20
#5060 redo build system

gnestor:


scottdraves:
```
$ jupyter nbextension install --py --symlink --sys-prefix beakerx
  File "/Users/spot/anaconda/envs/testx/bin/jupyter-nbextension", line 6, in <module>
    sys.exit(notebook.nbextensions.main())
  File "/Users/spot/anaconda/envs/testx/lib/python3.5/site-packages/jupyter_core/application.py", line 267, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "/Users/spot/anaconda/envs/testx/lib/python3.5/site-packages/traitlets/config/application.py", line 658, in launch_instance
    app.start()
  File "/Users/spot/anaconda/envs/testx/lib/python3.5/site-packages/notebook/nbextensions.py", line 900, in start
    super(NBExtensionApp, self).start()
  File "/Users/spot/anaconda/envs/testx/lib/python3.5/site-packages/jupyter_core/application.py", line 256, in start
    self.subapp.start()
  File "/Users/spot/anaconda/envs/testx/lib/python3.5/site-packages/notebook/nbextensions.py", line 678, in start
    self.install_extensions()
  File "/Users/spot/anaconda/envs/testx/lib/python3.5/site-packages/notebook/nbextensions.py", line 657, in install_extensions
    **kwargs
  File "/Users/spot/anaconda/envs/testx/lib/python3.5/site-packages/notebook/nbextensions.py", line 211, in install_nbextension_python
    m, nbexts = _get_nbextension_metadata(module)
  File "/Users/spot/anaconda/envs/testx/lib/python3.5/site-packages/notebook/nbextensions.py", line 1037, in _get_nbextension_metadata
    'it is missing the `_jupyter_nbextension_paths()` method.'.format(module))
KeyError: 'The Python module beakerx is not a valid nbextension, it is missing the `_jupyter_nbextension_paths()` method.'
```

also the pip cmd runs for a very long time without showing any output.  that's a usability problem, is there a way to get it to show the output of its subcmds?

gnestor:
> also the pip cmd runs for a very long time without showing any output

The best thing that I have found is using the `-v` flag on `pip install -e . -v`. This can also be set using a config object: https://pip.pypa.io/en/stable/user_guide/#configuration and I assume that it can be set in some options somewhere, I just can't find any good docs about it.

> 'The Python module beakerx is not a valid nbextension, it is missing the `_jupyter_nbextension_paths()` method.'

I'll investigate now...

gnestor:
@scottdraves This resolves the 2 above issues for me :+1:

scottdraves:
cool thanks, yup that works :)

but i think we need some adjustments.

1) this has pip running npm and gradle. my research on how to make this work with conda: https://github.com/twosigma/beakerx/issues/5476
what jupyter lab does is not build the JS from pip.  we should do the same, and for java too.
also, we are using yarn not npm.  So can you pls remove those parts?

2) also this line was lost i think it's needed `conda install -y -c conda-forge openjdk`

3) the "gradle kernelInstall" should be done by pip with python though.

scottdraves:
On 2nd thought I think there is another change needed:

4. the installation of the kernelspec has to be done in the recipe, so it should be left as a separate cmd, not built into pip install.

gnestor:
> this has pip running npm and gradle. my research on how to make this work with conda: #5476
what jupyter lab does is not build the JS from pip. we should do the same, and for java too.
also, we are using yarn not npm. So can you pls remove those parts?

I pushed a commit that switches from npm to yarn. 

Until recently (when the JupyterLab extension system was refactored), lab was installed with just a `pip install -e .` that would install JS dependencies. It used the `setupbase.py` from https://github.com/jupyter/jupyter-packaging. Now, the contributor installation instructions include an `npm install` and `npm run build:main`. Ideally, we want a single installation command like `pip install -e .` for non-conda users. Let's see how we can accomplish this with a conda-forge recipe and circle back.

> also this line was lost i think it's needed conda install -y -c conda-forge openjdk

I added openjdk to the `conda create` step: `conda create -y -n beakerx python=3.5 jupyter pandas openjdk`. I tested and it works 👍  

It looks like the conda-recipe takes care of this: https://github.com/twosigma/staged-recipes/blob/master/recipes/beakerx/meta.yaml#L28

> the "gradle kernelInstall" should be done by pip with python though.

See https://github.com/jupyterlab/mimerender-cookiecutter/pull/45#issuecomment-280940999: "So far it has been our "policy" and recommendation to not install or enable lab/nb extensions from setup.py scripts. The reason is that a user may have multiple extensions (a theme for example) that does the same thing. Because of this, we are currently recommending that packagers force users to do the separate lab/nb extension installation. "

> the installation of the kernelspec has to be done in the recipe, so it should be left as a separate cmd, not built into pip install.

This is a better solution and follows best practices. For example, if you look at the [ipywidgets install instructions](https://github.com/jupyter-widgets/ipywidgets#install):

> With pip:
> ```
> pip install ipywidgets
> jupyter nbextension enable --py --sys-prefix widgetsnbextension
> ```
> With conda:
> ```
> conda install -c conda-forge ipywidgets
> ```

gnestor:
Any suggestion regarding how to handle this addition:

```
## Build including C++ kernel
./gradlew --no-daemon build -DincludeCpp=true
./gradlew --no-daemon kernelInstall -DincludeCpp=true
```

scottdraves:
thx grant, will study this later, as for C++ we are leaving it out of the 1st release.

beakerx/js/package.json
```diff
@@ -16,8 +16,8 @@
     "ipywidgets"
   ],
   "scripts": {
-    "prepublish": "webpack",
-    "postinstall": "webpack",
+    "build": "webpack",
+    "prepublish": "npm run build",
```
scottdraves: should be `yarn install` right?

beakerx/beakerx/install.py
```diff
@@ -1,105 +0,0 @@
...
-    if enable == disable:
-        log.error("Please provide (one of) --enable or --disable")
-        raise ValueError(enable, disable)
-
-    log.info("{}abling BeakerX server config...".format("En" if enable else "Dis"))
-
-    path = jupyter_config_dir()
-
-    if prefix is not None:
-        path = join(prefix, "etc", "jupyter")
```
scottdraves: has to be pjoin right?  this just crashes when i run it.

beakerx/setupbase.py
```diff
@@ -0,0 +1,499 @@
...
+            if enable == disable:
+                log.error("Please provide (one of) --enable or --disable")
+                raise ValueError(enable, disable)
+
+            log.info("{}abling BeakerX server config...".format("En" if enable else "Dis"))
+
+            path = jupyter_config_dir()
+
+            if prefix is not None:
+                path = join(prefix, "etc", "jupyter")
```
scottdraves: has to be pjoin right? this just crashes when i run it.

scottdraves:
>> the installation of the kernelspec has to be done in the recipe, so it should be left as a separate cmd, not built into pip install.

> This is a better solution and follows best practices. For example, if you look at the [ipywidgets install instructions](https://github.com/jupyter-widgets/ipywidgets#install):

i don't get it, ipywidgets don't have any kernels.  let's discuss by IM.

gnestor:
https://github.com/twosigma/beakerx/issues/5576 is fixed by the latest commit here. Next step is to start moving that and other parts to the conda recipe...

scottdraves:
that sounds cool but i get a problem from yarn:

```
$ conda create -y -n testx python=3.5 jupyter pandas openjdk yarn
Fetching package metadata .........

PackageNotFoundError: Package missing in current osx-64 channels:
  - yarn

Close matches found; did you mean one of these?

    yarn: yarl
```
then hit into this problem:
```
jupyter nbextension install --py --symlink --sys-prefix beakerx
```
produces
```
ImportError: No module named 'IPython.utils._tokenize_py3'
```
which I worked around like this: https://github.com/ipython/ipykernel/issues/151
but then got
```
RuntimeError: Python2 was found when trying to start Python3. _beaker_python_mismatch_
```

Issue #5523, opened 2017-06-21
Tables not displayed on OutputContainer TAB


konstantinTarletski:
To reproduce this issue, You need open groovy , paste code :

```
import com.twosigma.beakerx.jvm.object.TabbedOutputContainerLayoutManager;
import com.twosigma.beakerx.jvm.object.OutputContainer;

mapList = [
   [a:1, b:2, c:3],
   [a:4, b:5, c:6],
   [a:7, b:8, c:5]
]
mapList2 = [
   [a:1, b:2, c:3],
   [a:4, b:5, c:6],
   [a:7, b:8, c:5]
]
tableDisplay = new TableDisplay(mapList)
tableDisplay2 = new TableDisplay(mapList2)

def l = new TabbedOutputContainerLayoutManager()
l.setBorderDisplayed(false)
def o = new OutputContainer()
o.setLayoutManager(l)
o.addItem(tableDisplay, "T1")
o.addItem(tableDisplay2, "T2")
o
```

Then switch to the "T2" tab. Table not displayed!

**_For now it can be reproduced on branch `kostja/5465`_**

Issue #5465, opened 2017-06-13
outputcontainers broken


scottdraves:
2nd cell crashes with OutputContainersAndLayoutManagers.ipynb:
```
TableDisplay.js:53 Uncaught (in promise) TypeError: Cannot read property 'tooManyRows' of undefined
    at TableDisplay.js:53
    at <anonymous>
(anonymous) @ TableDisplay.js:53
```

scottdraves:
we have a workaround now https://github.com/twosigma/beakerx/commit/1c8b440c2debfcf76302bbbf1d4f7cc0f74f497e

PR #5524, opened 2017-06-21
PR for issue #5465 "outputcontainers broken"

konstantinTarletski:
While fixing this issue, discovered one bug:
Tables not displayed on Output Container TAB #5523
Cuba already fix it.


scottdraves:
this works but the contents of tables are now sent twice.  this is too big a performance cost.

scottdraves:
If I evaluate this cell:
```
mapList = [
   [a:1, b:2, c:3],
   [a:4, b:5, c:6],
   [a:7, b:8, c:5]
]
tableDisplay = new TableDisplay(mapList)
tableDisplay
```
Then on this branch I get:
screen shot 2017-06-21 at 2 45 41 pm
and on master:
screen shot 2017-06-21 at 3 10 09 pm

you can see the packet of size 1459 is sent twice vs once.

scottdraves:
at this point, a workaround is probably good enough, and easier.
how about we just have the user call sendModel() at the right time?

konstantinTarletski:
I think it is possible to fix this issue without sending content of the widget twice.
But we need to decide which way we will do our widgets. There was spent a lot of efforts to make our widgets similar to python. But still we have some diferences with their behavier. There is not much left to do, to become our widgets to be like python.
But threre we need to decide several problems.

**Python widgets work this way:**

They send all his model while creation, and after they send only "update" notification if something in the model is changed. This is ok because this "update" notification have not mush data. Example :

```
import com.twosigma.beakerx.widgets.integers.IntSlider

w = new IntSlider() // creates widget, open comm, sends all model
w.value = 60 // send "update" notification
w // call Widget.display() , just display widget
```

All python widgets work this way. Python keep his widgets updated. And I think, if we want to to combine our widgets and python widgets we need to do our widgets similar.

**Our widgets**

Jaroslaw and Cuba did big job our widgets can operate like python widgets, but still work not same way:

```
tableDisplay = new TableDisplay(mapList) //Create widget, open comm, but DO NOT send model.
tableDisplay // call BeakerxWidget.display() // SEND MODEL, display widget
```

You can see the diference: model send only with calling "display" method.

Also our widgets support "update" like python widgets (It was last big change did by Jaroslaw and Cuba), we can do like this:

```
tableDisplay = new TableDisplay(mapList)
tableDisplay.setColumnOrder( ... ) // send "update" like python widgets
tableDisplay.setDataFontSize( ... ) //  send "update" like python widgets
 ... and others
tableDisplay // call BeakerxWidget.display() // but in there we send all model with changes we already send.
```

And as I discovered, send model in `BeakerxWidget.display()` method is not good practice. It can be happened, that `BeakerxWidget.display()` can not be called at all:

If widget belongs to container (`com.twosigma.beakerx.jvm.object.OutputContainer`), then called layout manager `display` method and `BeakerxWidget.display()` do not call at all.
In that case, with python widgets all OK, because thet have "updated" model, but not with our widgets, they may have no model at all.

In this case I can "support" our widgets: find all our layouts, and rewrite their `display` methods with calling "sendModel", but may be we find some case in future where it will be hard to fix.

I sugges to rewrite our widgets to be like python, it is not hard, we need to do this :
1) Move `sendModel` from "display" to widget constructor
2) Check  that all fields support "update"
That is all need to be done. So, seams everything ready, if we want do our widgets python way. Stay fiew steps.

**Problem**

But there we have one problem, as I said before python widgets have not much data inside, and sending "update" is not a problem. But not with our widgets. `TableDisplay` can have big data inside, all others our widgets is ok.

**Solution**

To resolve this problem, there are several ways:
1) send not all model, but only changed data. Like this:`{col : 1; row : 4 ; value = 10}`
In this case we will generare lot of "update" events, but thet will be small.
2) Another way, is do table content immutable.  And allow to user modify only all table values, in this case we will generate only one "update" event, if user will modyfy lot off table content.
```
def mapList4 = [[a:1, b:2, c:3],[a:4, b:5, c:6],[a:7, b:8, c:5]]
def display7 = new TableDisplay(mapList4)
display7.setDoubleClickAction { row, col, tableDisplay ->
   tableDisplay.values[row][col] = -tableDisplay.values[row][col] // will be be error
}
```
Users should do like this:

```
def mapList4 = [[a:1, b:2, c:3],[a:4, b:5, c:6],[a:7, b:8, c:5]]
def display7 = new TableDisplay(mapList4)
display7.setDoubleClickAction { row, col, tableDisplay ->
   Int [][]  content = tableDisplay.values;
   content[row][col] = -content[row][col];
   tableDisplay.values = content;
}
```
unmodifiable example:
```
List<String> strings = new ArrayList<String>();
List<String> unmodifiable = Collections.unmodifiableList(strings);
unmodifiable.add("New string"); // will fail at runtime
strings.add("Aha!"); // will succeed
System.out.println(unmodifiable);
```
3) May be some one have better suggestions, also we can allow modyfy only "small" tables: Sending "update" event can ne conditional. For example, if table have less then 10.000 elements we can allow "update", if more then disable it. And also inform user about it.
4) Also our widgets can stay like they are now. And I fix this, sending model 2 times, bug. But I think it is not right way.

Issue #5293, opened 2017-05-10
after drag-zoom, cannot unzoom completely


scottdraves:
The drag-zoom fix https://github.com/twosigma/beakerx/pull/5290 does not seem to be complete.
After using it, i cannot use the scroll wheel to unzoom completely.

PR #5528, opened 2017-06-21
#5293 fix zoom wheel after rectangle zooming

MariuszJurowicz:


scottdraves:
didn't work for me.  i was able to zoom a little but not far in or out.

Issue #5536, opened 2017-06-22
change table interaction action names to be compatible with classic beaker


scottdraves:
in classic beaker, a cell run by the table API would look like:
```
def details = display.details
switch(details.actionType){
  case 'DOUBLE_CLICK':
  print ("You clicked on the cell [" + details.row + ", " + details.col + "]")
    break;
  case 'CONTEXT_MENU_CLICK':
   print ("You selected context menu '" + details.contextMenuItem + "' on the cell [" + details.row + ", " + details.col + "]")
  break;
}
```
in BeakerX this is now:
```
def details = display.details
switch(details.actionType){
  case 'ONDOUBLECLICK':
  print ("You clicked on the cell [" + details.row + ", " + details.col + "]")
    break;
  case 'ONCONTEXTMENU':
   print ("You selected context menu '" + details.contextMenuItem + "' on the cell [" + details.row + ", " + details.col + "]")
  break;
}
```
please change to use the original Beaker names "DOUBLE_CLICK" and "CONTEXT_MENU_CLICK"

PR #5548, opened 2017-06-26
#5536 "change table interaction action names to be compatible with classic beaker"

konstantinTarletski:


scottdraves:
need to update the demo files to match.

Issue #5247, opened 2017-04-23
make repr of null configurable


scottdraves:
some users prefer null to be displayed as an empty result instead of "null"
maybe https://github.com/twosigma/beakerx/pull/5120 can handle that?

hadim:
You mean #5120 I guess :-)

scottdraves:
What do you think a good way to configure it is?

hadim:
As an attribute I can set overwriting `Kernel` ?

hadim:
See https://github.com/hadim/scijava-jupyter-kernel/blob/master/src/main/java/org/scijava/jupyter/kernel/ScijavaKernel.java#L53

PR #5565, opened 2017-06-28
PR for issue #5247 "make repr of null configurable"

konstantinTarletski:
In the example offered by Hadim, in `Kernel` uses `parameter`from some 3rd party lib.
`org.scijava.plugin.Parameter` 
Our libs do not have it. So, I implement it like `static` variable :

`com.twosigma.beakerx.kernel.Kernel.showNullExecutionResult` default value is `true`


hadim:
Nice. Thanks for that.

Issue #5544, opened 2017-06-23
Current Scala version (2.12) is not compatible with current Spark


jpallas:
The current stable Spark release (2.1) is based on Scala 2.11 and incompatible with Scala 2.12, and the upcoming 2.2 release is also not supporting Scala 2.12 (see [mail thread](http://apache-spark-developers-list.1001551.n3.nabble.com/RFC-removing-Scala-2-10-tp21116p21124.html)).

Of course, Scala 2.12 has other benefits, including better integration and performance with Java 8.  And unless/until the Spark/Scala integration of classic Beaker can be ported to the BeakerX Scala kernel, maybe it's not that important.

scottdraves:
thx good point

konstantinTarletski:
I tried to switch to version 2.11.11. But get some Scala compilation errors.
```
:kernel:scala:compileScala
C:\WorkspaceBeakerx\beakerx\kernel\scala\src\main\java\com\twosigma\beakerx\scala\fileloader\CsvPlotReader.scala:25: not found: value asScalaBuffer
    val refactoredOutput : List[util.Map[_,_]] = asScalaBuffer(javaOutput).toList
                                                 ^
C:\WorkspaceBeakerx\beakerx\kernel\scala\src\main\java\com\twosigma\beakerx\scala\fileloader\CsvPlotReader.scala:27: not found: value mapAsScalaMap
    refactoredOutput.map(ro => mapAsScalaMap(ro).toMap)
                               ^
C:\WorkspaceBeakerx\beakerx\kernel\scala\src\main\java\com\twosigma\beakerx\scala\table\TableDisplay.scala:31: not found: value mapAsJavaMap
    new com.twosigma.beakerx.table.TableDisplay(mapAsJavaMap(v))
                                                ^
C:\WorkspaceBeakerx\beakerx\kernel\scala\src\main\java\com\twosigma\beakerx\scala\table\TableDisplay.scala:42: not found: value mapAsJavaMap
    val javaStandardized : Array[java.util.Map[_, _]]= v.map(v => mapAsJavaMap(v)).toArray
                                                                  ^
C:\WorkspaceBeakerx\beakerx\kernel\scala\src\main\java\com\twosigma\beakerx\scala\table\TableDisplay.scala:42: value toArray is not a member of Array[Nothing]
    val javaStandardized : Array[java.util.Map[_, _]]= v.map(v => mapAsJavaMap(v)).toArray
                                                                                   ^
[...]

PR #5582, opened 2017-07-03
#5544 Downgrade scala to 2.11

michalgce:
…allity.


kernel/scala/build.gradle
```diff
@@ -43,11 +43,17 @@ repositories {
   mavenCentral()
 }
 
+compileScala {
+  scalaCompileOptions.additionalParameters = [
+      "-Xexperimental"
```
scottdraves: please do not use experimental options, instead eliminate the use of SAM.

kernel/scala/src/main/scala/com/twosigma/beakerx/scala/evaluator/ScalaEvaluatorGlue.scala
```diff
@@ -14,7 +14,9 @@
  *  limitations under the License.
  */
 
-package com.twosigma.beakerx.scala.evaluator;
+package com.twosigma.beakerx.scala.evaluator
+
+;
```
scottdraves: surely this is a typo

PR #5584, opened 2017-07-03
PR for "Java evaluator fix"

konstantinTarletski:


scottdraves:
i believe you said this can no longer be reproduced...

Issue #5578, opened 2017-07-01
implement CyclingOutputContainerLayoutManager


scottdraves:
```
import com.twosigma.beakerx.jvm.object.CyclingOutputContainerLayoutManager;
import com.twosigma.beakerx.jvm.object.OutputContainer;
def l = new CyclingOutputContainerLayoutManager()
l.setPeriod(2000); // milliseconds
l.setBorderDisplayed(false);
def o = new OutputContainer()
o.setLayoutManager(l)
o.addItem(plot1, "Scatter with History")
o.addItem(plot2, "Short Term")
o.addItem(plot3, "Long Term")
o.addItem(table, "1990/01")
o
```
add this cell to the OutputContainersAndLayoutManagers demo and make it work.
right now it gives:
```
java.lang.RuntimeException: Not implemented yet.
	at com.twosigma.beakerx.jvm.object.CyclingOutputContainerLayoutManager.display(CyclingOutputContainerLayoutManager.java:31)
```

PR #5589, opened 2017-07-04
PR for isue #5578 "implement CyclingOutputContainerLayoutManager"

konstantinTarletski:


scottdraves:
basically worked, when i ran the cell the cycling output appeared.
but the cell didn't finish executing, the notebook continued to say `In [*]`, instead of the * turning to a number like it should when a cell completes.

scottdraves:
Please do the animation on the client, not in Java.

PR #5590, opened 2017-07-04
PR for Kotlin kernel

konstantinTarletski:


altavir:
Oh, great idea! I wanted to propose a kotlin kernel (since I am moving a lot of code to kotlin from java/groovy), but decided to postpone it until first stable release since you seemed to have your hands full.

konstantinTarletski:
Thanks , it was my suggestion to do Kotlin kernel.

scottdraves:
when i run any cell i get
```
ERROR: warning: classpath entry points to a non-existent location: /Users/spot/src/beakerx/kernel/kotlin/build/libs/kotlin.jar;
warning: classpath entry points to a non-existent location: <no_path>/lib/kotlin-runtime.jar
warning: classpath entry points to a non-existent location: <no_path>/lib/kotlin-script-runtime.jar
warning: classpath entry points to a non-existent location: <no_path>/lib/kotlin-reflect.jar
```

konstantinTarletski:
Now it should work for linux/mac.
But for now we will have warnings on exeption. 
default

konstantinTarletski:
There is not much info about Kotlin compilers in internet.
But, us I know, there is only one way to run kotlin code in JVM.
This is `org.jetbrains.kotlin.cli.jvm.K2JVMCompiler`

Same class uses by `kotlinc` - kotlin command line compiller.

Problem that stay in kotlin kernel is that if code have errors, then we get warnings that some classpath is missing.

```
warning: classpath entry points to a non-existent location: <no_path>\lib\kotlin-runtime.jar
warning: classpath entry points to a non-existent location: <no_path>\lib\kotlin-script-runtime.jar
warning: classpath entry points to a non-existent location: <no_path>\lib\kotlin-reflect.jar
```
Standard flags like `suppressWarnings` do not help.

This `K2JVMCompiler` wants that kotlin should be installed on local PC.
And one of it mandatory parameters is `kotlinHome`. (will ne set in compiller if null)
`kotlinHome` is a path to main kotlin libs, and it should point where is folder where is "lib" folder with 3 jar files in it:

Example:

kotlinHome = "C:\abc\"

This means that it should be like this:
```
C:\abc\lib\kotlin-runtime.jar
C:\abc\lib\kotlin-script-runtime.jar
C:\abc\lib\kotlin-reflect.jar
```

With certains names! Jar names can not be others.

This error becames from `K2JVMCompiler`:

```
class K2JVMCompiler : CLICompiler<K2JVMCompilerArguments>() {
    override fun doExecute(arguments: K2JVMCompilerArguments, configuration: CompilerConfiguration, rootDisposable: Disposable): ExitCode {
        val messageCollector = configuration.getNotNull(CLIConfigurationKeys.MESSAGE_COLLECTOR_KEY)

        val paths = if (arguments.kotlinHome != null)
            KotlinPathsFromHomeDir(File(arguments.kotlinHome))
        else
            PathUtil.getKotlinPathsForCompiler()
```

In that code You can see that variable `path`  get values `arguments.kotlinHome`, if `kotlinHome` is not set it takes default varibles. But any way it adds this values to `paths` variable. 
Like `C:\abc\lib\kotlin-runtime.jar` or `<no_path>\lib\kotlin-runtime.jar`, depends on does `kotlinHome` was set or not.

And after that this `paths` variable is mixed with `classpath`
`val classpath = getClasspath(paths, arguments)` -- line 110

That is why we get warning that classpath do not have this lbs.

So, the most right way is rewrite this `K2JVMCompiler#doExecute` method with right `KotlinPathsFromHomeDir` value. 

`KotlinPathsFromHomeDir` is path holder, to each "jar" of Kotlin core.

Its content :
```
public class KotlinPathsFromHomeDir implements KotlinPaths {
    // kotlinc directory
    private final File homePath;

    public KotlinPathsFromHomeDir(@NotNull File homePath) {
        this.homePath = homePath;
    }

    @Override
    @NotNull
    public File getHomePath() {
        return homePath;
    }

    @Override
    @NotNull
    public File getLibPath() {
        return new File(homePath, "lib");
    }

    @Override
    @NotNull
    public File getRuntimePath() {
        return getLibraryFile(PathUtil.KOTLIN_JAVA_RUNTIME_JAR);
    }

    @NotNull
    @Override
    public File getStdlibPath() {
        return getLibraryFile(PathUtil.KOTLIN_JAVA_STDLIB_JAR);
    }

    @NotNull
    @Override
    public File getReflectPath() {
        return getLibraryFile(PathUtil.KOTLIN_JAVA_REFLECT_JAR);
    }

    @Override
    @NotNull
    public File getScriptRuntimePath() {
        return getLibraryFile(PathUtil.KOTLIN_JAVA_SCRIPT_RUNTIME_JAR);
    }
...
```

**How to fix:**

1) Write custom `K2JVMCompiler` compiler with "right" paths values.
2) Make fake (empty) jar files. I think that it should help, because it mix classpath and `kotlinHome`. And there is no problem if we will have empty jar in classparh.
3) Filter error message.
4) Make folder with symbolic links to this jar files with right name. (Not sure that is easy to do). We even do not need real jars, because our kernel `kotlin.jar` have all this inside. If we chose this way, then we will need to redo build of kotlin kernel, to not duplicating kotlin core files.

I think that best ways is 1 and 3.

altavir:
OK, I wanted to propose to write to kotliners themselves, but I see, that you already done that. :)

konstantinTarletski:
There is some implementation of jupyter kernel, written on Kotlin:
https://github.com/ligee/kotlin-jupyter

altavir:
Also You can try to look for some ideas in [kscript](https://github.com/holgerbrandl/kscript). They seem to solve problems for local script evaluation which is not far from what needed for remote kernel.

udalov:
FTR I've described the intended solution in https://discuss.kotlinlang.org/t/k2jvmcompiler/3629/3

konstantinTarletski:
PR fixed, thanks to "ilya.chernikov"

PR #5598, opened 2017-07-06
Avoid warnings running Scala tests by supplying non-empty string

jpallas:
The change in PR #5581 uses an empty string, which makes the tests issue (harmless but annoying) warnings (at run-time) from the compiler.  Substituting a trivial, valid expression silences the warnings.


scottdraves:
hi thanks for the PR -- which warning is this about?

jpallas:
Oh.  Um.  The gradle build doesn't actually show these warnings, because it is not configured to send test output to the console.  But if you run the Scala tests in IntelliJ, you see warnings like this:
```
[main] WARN com.twosigma.beakerx.scala.evaluator.ScalaEvaluator - ERROR creating beaker beaker: input is incomplete
[scala worker] WARN com.twosigma.beakerx.scala.evaluator.ScalaEvaluator - ERROR creating beaker object: input is incomplete
```

scottdraves:
interesting well let's fix that first.

scottdraves:
👍

Issue #5610, opened 2017-07-07
merge new content into plot features demo


scottdraves:
update doc/contents/PlotFeatures.ipynb with new content from doc/contents/plotDemo.ipynb
and fix the rates table to be read from the csv like in doc/contents/chartingTutorial.ipynb

part of #5487

samwincott:
Can I take this on? I'd love to contribute to the project.

scottdraves:
sure, send us a completed CLA. i would be happy to describe the issue in more detail.

samwincott:
Great, I'll fill one out now. I think I understand fixing the rates table, but may need more clarification on the first part.

scottdraves:
that would be awesome thank you

scottdraves:
note I will not be able to see the form or test your PR until Monday, but I can answer any questions you have here at any time.

samwincott:
What new content was it that you wanted moved from plotDemo to plotFeatures?

scottdraves:
the raster demos, anything else new

BeakerNotebook.com
ScottDraves.com
twitter.com/Scott_Draves

samwincott:
I have finalised the pull request (#5615) for this issue, and hope it was what you were looking for. I'm happy to change anything if needs be.

scottdraves:
thanks will lool at it tomorrow

PR #5615, opened 2017-07-08
PR for issue #5610

samwincott:
I believe this is what you were looking for in the second part of #5610, if not I'm happy to change it.
I will start on the first part now.


scottdraves:
awesome ty, will check it out tomorrow.

Issue #5586, opened 2017-07-03
Allow classpath wildcards


jpallas:
The recent fix for #5483 removed the ability to do all-jars-in-directory wildcards via `%classpath add jar /path/to/dir/*`.  I'm not sure if it was intentional or accidental that this worked, but I think it is useful.

scottdraves:
That was accidental, thanks for pointing it out.

PR #5618, opened 2017-07-10
#5586 Added support for wildcard in classpath.

michalgce:


kernel/base/src/main/java/com/twosigma/beakerx/kernel/commands/MagicCommand.java
```diff
@@ -178,45 +183,80 @@ private MagicCommandFunctionality classpathAddJar() {
           throw new IllegalStateException("Wrong command format: " + CLASSPATH_ADD_JAR);
         }
 
-        this.kernel.addJarToClasspath(new PathToJar(split[3]));
+        String path = split[3];
+        if (doesPathContainsWildCards(path)) {
+          getPaths(path)
+              .forEach(currentPath -> kernel.addJarToClasspath(new PathToJar(currentPath)));
```
scottdraves: so we are expanding wildcard instead of letting java handle it.
i think this could cause a problem on windows since there is a fairly small limit to total size of the classpath.
otoh i like the ability to report exactly the matched files.

kernel/base/src/main/java/com/twosigma/beakerx/kernel/commands/MagicCommand.java
```diff
@@ -178,45 +183,80 @@ private MagicCommandFunctionality classpathAddJar() {
...
 
+  private List<String> getPaths(String pathWithWildcard) {
+    String pattern = "**" + pathWithWildcard;
+    PathMatcher pathMatcher = FileSystems.getDefault().getPathMatcher("glob:" + pattern);
+    Path startPath = Paths.get("");
+
+    try {
+      List<String> paths = Files.find(startPath, Integer.MAX_VALUE,
+          (path, basicFileAttributes) -> pathMatcher.matches(path))
+          .filter(path -> path.toString().contains(".jar"))
```
scottdraves: this doesn't look right

michalgce: Could you provide more details? 

My target here is to import only files that contains .jar

kernel/base/src/main/java/com/twosigma/beakerx/kernel/commands/MagicCommand.java
```diff
@@ -190,6 +204,36 @@ private MagicCommandFunctionality classpathAddJar() {
     };
   }
 
+  private void validateWildcardPath(String path) {
+    if (!path.matches("^([^*]*\\*){1}[^*]*$") || !path.endsWith("*")) {
```
scottdraves: not sure what this does.
i believe all we want to say is if it has a star, it has to be at the end only.
matches is too complicated for just that.

kernel/base/src/main/java/com/twosigma/beakerx/kernel/commands/MagicCommand.java
```diff
@@ -190,6 +204,36 @@ private MagicCommandFunctionality classpathAddJar() {
...
+    }
+  }
+
+  private List<String> getPaths(String pathWithWildcard) {
+    String pathWithoutWildcards = pathWithWildcard.replace("*", "");
+    try {
+
+      List<String> paths = Files.list(Paths.get(pathWithoutWildcards))
+                                .map(Path::toString)
+                                .filter(path -> path.endsWith(".jar"))
```
scottdraves: should not be case sensitive

kernel/base/src/main/java/com/twosigma/beakerx/kernel/commands/MagicCommand.java
altavir: Here is a simple mask to pattern converter I made for myself:
```java
    public static String applyMask(String pattern) {
        return pattern.replace(".", "\\.").replace("?", ".").replace("*", ".*?");
    }
```

kernel/base/src/main/java/com/twosigma/beakerx/kernel/commands/MagicCommand.java
```diff
@@ -190,6 +204,36 @@ private MagicCommandFunctionality classpathAddJar() {
     };
   }
 
+  private void validateWildcardPath(String path) {
+    if (!path.matches("^([^*]*\\*){1}[^*]*$") || !path.endsWith("*")) {
+      throw new IllegalStateException("Please provide correct path. Ex demoResources/*");
```
scottdraves: error message should read "Wildcard can only appear at end of classpath: "+path

kernel/base/src/main/java/com/twosigma/beakerx/kernel/commands/MagicCommand.java
scottdraves: we are not supporting general globs, just a * at the end, like java classpaths normally do.

altavir: In this case it will be simply:
```java
path.equals(mask) || (mask.endWith('*') && path.startsWith(mask.substring(0,mask.length-2)))
```

michalgce: @scottdraves regex check if theres only one * symbol in path, and after OR statement i'm checking if statement is ended with *

scottdraves: regex is a bad way to check that.
they are 1) slow 2) too complicated.

scottdraves:
i tried the classpath demo and put in a * for the jar and it just hung:

screen shot 2017-07-11 at 6 33 56 pm

michalgce:
Strange, because here working 
image

scottdraves:
i see it hang even without the *

Issue #5619, opened 2017-07-10
fonts in plot save-as


scottdraves:
if you save a plot as SVG or PNG then the typeface is not lato

PR #5647, opened 2017-07-12
#5619 set Lato font for contextMenu

jszalek:


scottdraves:
this change is good, all our UI elements should be Lato, but what I mean for this issue is much bigger.
I don't mean the "save as" menu itself I mean the file produced by it.  here's a closeup of the title of a saved SVG plot to make it clear:
screen shot 2017-07-12 at 10 06 10 am

should be like this, as in the application:
screen shot 2017-07-12 at 10 06 20 am

scottdraves:
png looks good but svg gives me "Failed - Network error"

PR #5648, opened 2017-07-12
jarek/5627: get rid of fat jars

jaroslawmalekcodete:


EfimovVladimir:
checked on windows - all is works

scottdraves:
i am seeing this the kernel.json:
```
"java", "-cp", "/Users/spot/src/beakerx/kernel/groovy/build/install/groovy/lib/groovy.jar:/Users/spot/src/beakerx/kernel/groovy/build/install/groovy/lib/*",
```
i what is groovy.jar?  still seems like we have a fat one.  in the classpath on the cmd line, why specify it and then specify the directory that contains it?  that seems redundant.

and there is also base.jar, another fat jar and it is copied into each kernel.

i would expect two directories on the classpath, one for for base (shared between all kernels) and one for each kernel.  each directory would have a bunch of jar files.

Issue #5477, opened 2017-06-14
magic/empty code return value


scottdraves:
if the jar has already been loaded and this is a noop, then it should be like output.hidden, as if you ran an empty cell for example.

if the jar has not been loaded, then it should print as if it were stdout "added *jar-name*"

![screen shot 2017-06-14 at 9 29 31 am](https://user-images.githubusercontent.com/963093/27134590-09d4fb62-50e4-11e7-803a-fd2c82f10094.png)

jaroslawmalekcodete:
The issue was fixed in PR https://github.com/twosigma/beakerx/pull/5553, please verify.

scottdraves:
> if the jar has not been loaded, then it should print as if it were stdout "added jar-name"

not seeing this part...

michalgce:
May we close this issue @scottdraves ?

PR #5651, opened 2017-07-12
#5477 Print message after succesfully added classpath by magic command.

michalgce:


kernel/base/src/main/java/com/twosigma/beakerx/kernel/commands/MagicCommand.java
```diff
@@ -35,14 +38,15 @@
...
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.PathMatcher;
+import java.nio.file.Paths;
+import java.util.*;
```
scottdraves: wildcard not allowed in imports

scottdraves:
1) i don't think this is printing the status like it is stdout.  that means if you have a %classpath add, followed by a value that returns something, you should see the status from the classpath, and also the result.  i am seeing the status be lost.

2) the status should not be printed if the jar is already loaded.  if you run the same %classpath add twice, it should only print that it's loading the 1st time.

Issue #5635, opened 2017-07-11
EasyForm List component broken


scottdraves:
1) open EasyFormDemos.ipynb
2) run the cell commented with "//All Kinds of Fields"
3) click on the List widget, select one of the items, like 'a'
4) click "Go" button

result: table has row that says the List is empty
expected result: table has row that says the List has the value 'a'

PR #5664, opened 2017-07-14
#5635: fix EasyForm List component

jaroslawmalekcodete:


scottdraves:
Thanks great I can see the results now.
But it should return an ArrayList of the selected values, not a string with commas.
The same goes for the Check Boxes component!

Issue #5662, opened 2017-07-13
add new gradle task "static"


scottdraves:
for python distribution we need the jars and kernelspecs (today just the kernel.json but eventually there will be icons in them too) to be in the static directory (beakerx/beakerx/static).

so after the jars are built, copy them there with this directory structure:
```
static/kernel/base/*.jar
static/kernel/groovy/lib/*.jar
static/kernel/groovy/kernel.json
```

this kernel.json is the template, it has `__PATH__` (but it should have the main class filled in).

scottdraves:
part of https://github.com/twosigma/beakerx/issues/5134

PR #5667, opened 2017-07-14
#5562: add staticContent task

jaroslawmalekcodete:


scottdraves:
actually those commits are for #5662

altavir:
It is a bad way to work with gradle. You have to use `dependsOn` instead of `doLast` wherever it is possible. Or else you will have problems with cleaning and possibly much longer build times.

scottdraves:
thx for pointing that out: https://github.com/twosigma/beakerx/issues/5670

Issue #5640, opened 2017-07-12
closure callbacks not working


scottdraves:
i am pretty sure these used to work, but now i am getting nothing.
```
import com.twosigma.beakerx.table.*
mapList = [
   [a:1, b:2, c:3],
   [a:4, b:5, c:6],
   [a:7, b:8, c:5]
]
def display1 = new TableDisplay(mapList)
display1.setDoubleClickAction { row, col, tableDisplay ->
   println("dbl click")
}
display1.addContextMenuItem("negate") { row, col, tableDisplay ->
   println("ctx menu")
}
display1
```

PR #5684, opened 2017-07-17
#5640 Fixed Closure callback.

michalgce:


kernel/base/src/main/java/com/twosigma/beakerx/widgets/Widget.java
```diff
@@ -42,11 +42,11 @@
   
   public enum CommActions {
 
-    DOUBLE_CLICK("DOUBLE_CLICK"),
+    DOUBLE_CLICK("ondoubleclick"),
```
scottdraves: we need to keep this i think to maintain this fix as well:
https://github.com/twosigma/beakerx/issues/5536
it seems that 5536 was fixed incorrectly is the source of this regression...

kernel/base/src/main/java/com/twosigma/beakerx/table/TableDisplay.java
```diff
@@ -627,7 +629,17 @@ public void setDoubleClickAction(Object listener) {
   }
 
   private void handleDoubleClick(Message message) {
-    handleCommEventSync(message, CommActions.DOUBLE_CLICK, this::onDoubleClickAction);
+    if (isCorrectEvent(message, CommActions.DOUBLE_CLICK)) {
+      handleCommEventSync(message, CommActions.DOUBLE_CLICK, this::onDoubleClickAction);
+    }
+  }
+
+  private boolean isCorrectEvent(Message message, CommActions commActions) {
```
scottdraves, line 639: unclear to me what this function is for?
when does it return false? wouldn't this be an error condition?

kernel/base/src/main/java/com/twosigma/beakerx/table/TableDisplay.java
```diff
@@ -648,11 +665,17 @@ public void fireDoubleClick(List<Object> params, Message message) {
   }
 
   private void handleSetDetails(Message message) {
-    handleCommEventSync(message, CommActions.ACTIONDETAILS, this::onActionDetails);
+    if (isCorrectEvent(message, CommActions.ACTIONDETAILS)) {
+      handleCommEventSync(message, CommActions.ACTIONDETAILS, this::onActionDetails);
+    }
+    LOGGER.warn("Warning: set details handler received non-set details message " + message.toString());
```
scottdraves: the warning should be inside the conditional.

Issue #5639, opened 2017-07-12
vertical header text is lost


scottdraves:
```
import com.twosigma.beakerx.table.*
import com.twosigma.beakerx.fileloader.CsvPlotReader

def mapList5 = [
 [firstCol:1, secondCol:2, thirdCol:3],
 [firstCol:4, secondCol:5, thirdCol:6],
 [firstCol:9, secondCol:8, thirdCol:9]
]
def td4 = new TableDisplay(mapList5)
//td4.headerFontSize = 30

td4.setHeadersVertical(true)

td4
```
even at regular size i get "..." even though it looks like it would fit.
![screen shot 2017-07-11 at 8 59 29 pm](https://user-images.githubusercontent.com/963093/28097138-db6c5470-667b-11e7-919c-bfbf7df030eb.png)
if i increase the size then it gets worse:
![screen shot 2017-07-11 at 9 00 38 pm](https://user-images.githubusercontent.com/963093/28097161-02debeb2-667c-11e7-85d1-de32b4bf320c.png)

PR #5693, opened 2017-07-19
#5639 fix header large font clipped

MariuszJurowicz:


scottdraves:
is actually for https://github.com/twosigma/beakerx/issues/5638

PR #5705, opened 2017-07-20
#5657 plot tooltip should fade in out

MariuszJurowicz:


scottdraves:
looks good except it's possible to have more than one tooltip visible when the mouse moves over geometry.  leaves a weird trail and blinking.  if the tooltip is already visible it should just move, without fading in/out.  only one tooltip should be visible from hover.

A good place to see this is in the Area element of the "spread" plot in chartingTutorial.ipynb.

MariuszJurowicz:
Should work now.

scottdraves:
thanks yes that fixes that problem, but there is something also changed from master which bothers me and that's the tooltip no longer moves with the mouse.  it sounds minor but it actually causes a major problem, which is now the mouse can go over the tooltip, and then the tooltip vanishes.  this is easy to reproduce with categoryPlot.ipynb.

can you keep the old behavior where the tooltip moves, and also keep the fading?

Issue #5698, opened 2017-07-19
column search should be case insensitive


scottdraves:
The table menu -> Show Column search field should match regardless of case (xy matches XY).

PR #5708, opened 2017-07-20
#5698 column search should be case insensitive

MariuszJurowicz:


beakerx/js/src/shared/bkUtils.js
```diff
@@ -58,6 +58,9 @@ define([
     },
     newDeferred: function() {
       return jQuery.Deferred();
+    },
+    sanitizeRegExpString: function(string) {
```
scottdraves: actually prefer to allow user to type a regexp.
just change the hint text in the box from "search..." to "search regexp ..."

PR #5709, opened 2017-07-20
#5699 pressing escape should close column search menu

MariuszJurowicz:


scottdraves:
cool.  please add: pressing ESC should close table column menu as well.

Issue #5656, opened 2017-07-12
isolate and eliminate signal warnings


scottdraves:
refactor our use of signal classes into a single file and then compile that file with the right flags to disable these warnings:
```
/Users/spot/src/beakerx/kernel/base/src/main/java/com/twosigma/beakerx/kernel/Kernel.java:40: warning: SignalHandler is internal proprietary API and may be removed in a future release
import sun.misc.SignalHandler;
               ^
/Users/spot/src/beakerx/kernel/base/src/main/java/com/twosigma/beakerx/kernel/Kernel.java:165: warning: SignalHandler is internal proprietary API and may be removed in a future release
    SignalHandler handler = new SignalHandler() {
    ^
/Users/spot/src/beakerx/kernel/base/src/main/java/com/twosigma/beakerx/kernel/Kernel.java:165: warning: SignalHandler is internal proprietary API and may be removed in a future release
    SignalHandler handler = new SignalHandler() {
                                ^
/Users/spot/src/beakerx/kernel/base/src/main/java/com/twosigma/beakerx/kernel/Kernel.java:166: warning: Signal is internal proprietary API and may be removed in a future release
      public void handle(Signal sig) {
                         ^
/Users/spot/src/beakerx/kernel/base/src/main/java/com/twosigma/beakerx/kernel/Kernel.java:172: warning: Signal is internal proprietary API and may be removed in a future release
      Signal.handle(new Signal("INT"), handler);
                        ^
/Users/spot/src/beakerx/kernel/base/src/main/java/com/twosigma/beakerx/kernel/Kernel.java:172: warning: Signal is internal proprietary API and may be removed in a future release
      Signal.handle(new Signal("INT"), handler);
```

PR #5735, opened 2017-07-25
#5656: eliminate signal warnings

jaroslawmalekcodete:


kernel/base/build.gradle
```diff
@@ -24,6 +24,7 @@ repositories {
 
 compileJava {
   options.compilerArgs << '-XDignore.symbol.file'
+  options.forkOptions.executable = 'javac'
```
scottdraves: doesn't this disable warnings on all files in the kernel subproject?

kernel/base/build.gradle
```diff
@@ -24,6 +24,7 @@ repositories {
 
 compileJava {
   options.compilerArgs << '-XDignore.symbol.file'
```
scottdraves: possible to make this apply only to KernelSignalHandler.java?

jaroslawmalekcodete: I found a way to eliminate signal warnings for only one file, I used @SuppressWarnings("sunapi").

Issue #5741, opened 2017-07-25
bad wildcard should not crash


scottdraves:
just report a nice error to the user, do not crash:

```
Exception in thread "Thread-0" java.lang.IllegalStateException: Wildcard can only appear at end of classpath: demoResources/*.jar
        at com.twosigma.beakerx.kernel.commands.MagicCommand.validateWildcardPath(MagicCommand.java:220)
        at com.twosigma.beakerx.kernel.commands.MagicCommand.addJars(MagicCommand.java:202)
```

michalgce:
May we close it? @scottdraves

PR #5745, opened 2017-07-26
#5741 Catching crash on bad wildcard.

michalgce:


scottdraves:
This works but you should use conditionals instead of exceptions.

scottdraves:
much better thx but please have a more descriptive error msg.
If the wildcard is bad, say "Bad classpath wildcard syntax, path can only end with *"
If the file is not found then "Bad classpath, file not found"
screen shot 2017-07-26 at 9 49 44 am

PR #5754, opened 2017-07-27
Update build for jitpack

hadim:
ping @scottdraves 

That should allow the build on jitpack.


hadim:
I confirm, it worked.

PR #5760, opened 2017-07-28
Add test for classpath with wildcards resetting environment

jpallas:
This tests that a wildcard classpath add will only reset the environment once  (#5701).  The change also decouples this test suite from the structure of the doc directory.  Dummy jar files are used, since the tests don't depend on the files being valid jars.


scottdraves:
awesome, thank you.

PR #5178, opened 2017-04-13
#5158 add fieldset and legend, improve EasyForm implementation

jszalek:

PR #5762, opened 2017-07-31
#5178 fix heatmap width too low

MariuszJurowicz:


scottdraves:
here's the bug: https://github.com/twosigma/beakerx/issues/5718

Issue #5097, opened 2017-03-31
we need logo icons for our languages


scottdraves:
logo-32x32.png and logo-64x64.png in the same directory as kernel.json

scottdraves:
here's how they appear in lab:
<img width="995" alt="screen shot 2017-07-28 at 9 30 21 am" src="https://user-images.githubusercontent.com/963093/28750014-fcda1d86-74a9-11e7-8bde-b94255d1a4eb.png">

scottdraves:
use the same colored squares as we use on http://beakernotebook.com/
with the same letter codes, except replace the font (pt sans) with Lato.

MariuszJurowicz:
@scottdraves shouldn't this issue be closed already?

PR #5769, opened 2017-08-01
#5097 logo icons for beakerx languages

MariuszJurowicz:


scottdraves:
cool.  however for Kotlin, make the abbreviation be "Kt" and keep the text white like the others.

Issue #5778, opened 2017-08-02
allow quotes in magics.


scottdraves:
if there is a space in a filename, there is no way to load the jar for it.
we should allow:
```
%classpath add jar "path with spaces.jar"
```
we should use a standard library for this parsing as it can be quite complicated (backslashes and single quotes should be allowed).

PR #5787, opened 2017-08-03
#5778 Allow quotes in magics.

michalgce:


scottdraves:
please use a std library to handle the quoting.  one that uses the same syntax as groovy does would be ideal:
```
%classpath add jar 'file with spaces and "quotes"'
%classpath add jar "file with spaces and \"quotes\""
```

Issue #5794, opened 2017-08-03
bash magic crashes


scottdraves:
```
%%bash ls
```
crashes with this stack trace
```
        at java.util.Optional.get(Optional.java:135)
        at com.twosigma.beakerx.kernel.commands.MagicCommand.lambda$bash$36(MagicCommand.java:316)
```

it should report a nice error to the user.

PR #5806, opened 2017-08-08
#5794 - Handling bash magic crashes.

michalgce:


kernel/base/src/main/java/com/twosigma/beakerx/kernel/commands/MagicCommandFinder.java
```diff
@@ -41,6 +40,10 @@ public static MagicCommandFinder find(Code code, Map<String, MagicCommandFunctio
     List<MagicCommandItemWithResult> errors = new ArrayList<>();
     LinkedHashMap<String, MagicCommandFunctionality> functionalityToRun = new LinkedHashMap<>();
     code.getCommands().forEach(command -> {
+      if (command.startsWith("%%") && isCellmagicHeadNonEmpty(command)) {
+        errors.add(processIllegalCommand( "Cell magic head contains data move it to body.", message, executionCount, messageCreator));
```
scottdraves: Add comma.  It should read:
```
Cell magic head contains data, move it to body.
```

kernel/base/src/main/java/com/twosigma/beakerx/kernel/commands/MagicCommandFinder.java
```diff
@@ -55,22 +58,27 @@ public static MagicCommandFinder find(Code code, Map<String, MagicCommandFunctio
...
+        errors.add(processIllegalCommand("Cell magic " + command + " not found", message, executionCount, messageCreator));
       }
     });
     return new MagicCommandFinder(functionalityToRun, errors);
   }
 
+  private static boolean isCellmagicHeadNonEmpty(String command) {
+    List<String> tokens = new StrTokenizer(command).getTokenList();
+
+    return !(command.replace(tokens.get(0), "").replace(" ", "").length() < 1);
```
scottdraves, line 70: this is too complicated and it is not needed.
when it comes time to process the body, check if it is empty and then raise the error.

scottdraves:
disregard these comments, i fixed one of these myself and added the other to another issue.

PR #5833, opened 2017-08-10
Fix typo in readme

benmccann:
I'm assuming this is referring to https://github.com/jtablesaw/tablesaw. If so, then it should be one word

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ after a java change. See
 python -m beakerx.bkr2ipynb *.bkr
 ```
 
-## Groovy with Interactive Plotting and Table Saw:
+## Groovy with Interactive Plotting and Tablesaw:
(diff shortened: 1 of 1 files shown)


scottdraves:
thanks for pointing that out, indeed it's wrong.  we actually don't include tablesaw anymore, though you can load it and use it the same.
i fixed the readme in a slightly different way:
https://github.com/twosigma/beakerx/commit/aa7bcf4239efd6085f4530633a1279436db7609d

benmccann:
Is there a reason tablesaw was removed or anything I can do to help? I'm curious because I just started contributing very heavily to tablesaw and I haven't tried it with beakerx yet, but hope to do so soon.

scottdraves:
https://github.com/twosigma/beakerx/issues/5702

we changed from linking it in by default to making it an option you can install with the classpath magic: https://github.com/twosigma/beakerx/blob/master/doc/contents/tablesawDemo.ipynb

scottdraves:
oh and:
https://github.com/twosigma/beakerx/blob/master/doc/contents/Tablesaw_display_as_TableDisplay.ipynb

Issue #5810, opened 2017-08-08
%lsmagic output is badly formatted


scottdraves:
![screen shot 2017-08-08 at 7 16 40 pm](https://user-images.githubusercontent.com/963093/29098625-60942120-7c6e-11e7-9b87-4a09ab435ae1.png)

1) show one magic per line
2) show the parameters
3) do not show inapplicable magics (the data source ones only work for SQL)

PR #5839, opened 2017-08-11
#5810 Fixed lsmagic output.

michalgce:


kernel/base/src/main/java/com/twosigma/beakerx/kernel/commands/item/MagicCommandType.java
```diff
@@ -0,0 +1,34 @@
+package com.twosigma.beakerx.kernel.commands.item;
```
scottdraves, line 17: missing banner

kernel/base/src/main/java/com/twosigma/beakerx/kernel/commands/item/MagicCommandType.java
```diff
@@ -0,0 +1,34 @@
...
+  private final String parameters;
+  private final MagicCommandFunctionality functionality;
+  private final String suitableKernelsNames;
+
+  public MagicCommandType(String command, String parameters,
+      MagicCommandFunctionality functionality, String suitableKernelsNames) {
+    this.command = command;
+    this.parameters = parameters;
+    this.functionality = functionality;
+    this.suitableKernelsNames = suitableKernelsNames;
```
scottdraves: this isn't the right way to do this.
instead, have the kernel add the commands.
base kernel adds magics that work in all kernels.

kernel/base/src/main/java/com/twosigma/beakerx/kernel/KernelFunctionality.java
```diff
@@ -75,6 +75,8 @@
 
   void removeImport(ImportPath anImport);
 
+  String getName();
```
scottdraves: this method is not needed, see below.

kernel/base/src/main/java/com/twosigma/beakerx/kernel/commands/item/MagicCommandType.java
michalgce: good idea!

PR #5840, opened 2017-08-11
 #5909 Fixed Magic import with no arg.

michalgce:


scottdraves:
https://github.com/twosigma/beakerx/issues/5809

Issue #5782, opened 2017-08-03
allow setting dpi for png export


scottdraves:
would it be possible to allow the user to set the dpi for plots exported as png?
if i want to export at high quality this would be useful.
i tried zooming up my browser with control-+ and then saving as png but it didn't work....

jszalek:
@scottdraves I made some tests and it is possible to do it in easy way. It just need some changes in the code.

scottdraves:
awesome.  so add a 3rd item to the plot context menu "save as PNG at high DPI..." that has a submenus "2x" to "5x".  and move the SVG item to be first so the order is SVG, PNG, and PNG at high DPI.

PR #5841, opened 2017-08-11
#5782 allow setting dpi for png export

jszalek:


scottdraves:
this works great for most plots, but not CombinedPlots...

jszalek:
Added support for CombinedPlots

Issue #5843, opened 2017-08-13
%import garbage should fail


scottdraves:
when i do an %import of a class that does not exist it should give a nice error right away, not only when i later evaluate a regular cell.

PR #5855, opened 2017-08-17
#5843 %import garbage fail right now.

michalgce:


kernel/base/src/main/java/com/twosigma/beakerx/kernel/commands/MagicCommand.java
```diff
@@ -140,11 +143,30 @@ public MagicCommandFunctionality addImport() {
...
+
+      return sendErrorMessage(message, "Provided import doesn't exists. \n"
+          + "Try firstly add jar which contains class to classpath.", executionCount);
     };
   }
 
+  private boolean isValidImport(String part, int executionCount) {
+    try {
+      CompletableFuture<Boolean> validImportFuture = new CompletableFuture<>();
+      kernel.executeCode("Class.forName(\"" + part + "\")", new Message(), executionCount, seo -> {
```
scottdraves: There should be a more more direct way of doing this that doesn't use executeCode like this.
For example, this will fail with the Clojure kernel.

kernel/base/src/main/java/com/twosigma/beakerx/kernel/commands/MagicCommand.java
```diff
@@ -140,11 +143,30 @@ public MagicCommandFunctionality addImport() {
...
-      this.kernel.addImport(new ImportPath(parts[1]));
-      return getMagicCommandItem(code, message, executionCount);
+
+      if (isValidImport(parts[1], executionCount)) {
+        this.kernel.addImport(new ImportPath(parts[1]));
+        return getMagicCommandItem(code, message, executionCount);
+      }
+
+      return sendErrorMessage(message, "Provided import doesn't exists. \n"
+          + "Try firstly add jar which contains class to classpath.", executionCount);
```
scottdraves: Error message should read:
```
Could not import "$class", class not found.
```

kernel/base/src/main/java/com/twosigma/beakerx/kernel/commands/MagicCommand.java
scottdraves: how about `Class.forName(part);` with the shell's classpath installed?
can the autocomplete API or something like it help?

worst case call executeCode("") which is a hack but at least would work, I think.

michalgce: I tested it for Clorjure kernel and it works.

michalgce: Another approach could be a collect all classpaths and for each check if contains class that we would like to import - but i don't know how to collect this classpaths

kernel/base/src/main/java/com/twosigma/beakerx/kernel/commands/MagicCommand.java
```diff
@@ -144,11 +148,30 @@ public MagicCommandFunctionality addImport() {
...
+      }
+
+      return sendErrorMessage(message, "Could not import " + parts[1] + ", class not found.",
+          executionCount);
     };
   }
 
+  private boolean isValidImport(String part, int executionCount) {
+    try {
+      part = part.replace(";","");
```
scottdraves: i don't understand what this is for, there shouldn't be a semicolon, if there is it should be a syntax error reported to the user.

kernel/base/src/main/java/com/twosigma/beakerx/kernel/commands/MagicCommand.java
```diff
@@ -24,6 +24,8 @@
 import static java.util.Collections.singletonList;
 
 import com.google.common.collect.Lists;
+import com.twosigma.beakerx.autocomplete.ClasspathScanner;
```
scottdraves: used?

kernel/base/src/main/java/com/twosigma/beakerx/kernel/commands/MagicCommand.java
```diff
@@ -144,11 +148,30 @@ public MagicCommandFunctionality addImport() {
...
+      return sendErrorMessage(message, "Could not import " + parts[1] + ", class not found.",
+          executionCount);
     };
   }
 
+  private boolean isValidImport(String part, int executionCount) {
+    try {
+      part = part.replace(";","");
+      CompletableFuture<Boolean> validImportFuture = new CompletableFuture<>();
+      kernel.executeCode("Class.forName(\"" + part + "\");", new Message(), executionCount,
```
scottdraves: please use the empty cell hack.

PR #5868, opened 2017-08-21
jarek/groovy autocomplete

jaroslawmalekcodete:


jaroslawmalekcodete:
jenkins

scottdraves:
cool, looks like some great improvements, especially handling a class defined in the cell.

i had to disable the shouldImportBoolean test, IDK how it would pass here.

but mostly i am concerned about the parameter types that are showing up and getting included.  
screen shot 2017-08-22 at 12 19 41 pm

seems like just "println" is better, nothing more.

Issue #5875, opened 2017-08-23
tables with big integers failing


scottdraves:
Run 1st cell of bigInts.ipynb.  get the following on the JS console:
```
Exception in Comm callback SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at extension.js?v=20170823104218:50873
```

and note that the table displayed appears to treat the columns as floating point.
they should be transmitted as strings to preserve all the digits.

![screen shot 2017-08-23 at 11 02 32 am](https://user-images.githubusercontent.com/963093/29622871-973c9f42-87f2-11e7-98c6-7d99bd5bb072.png)

vs classic beaker

<img width="246" alt="screen shot 2017-08-23 at 11 03 54 am" src="https://user-images.githubusercontent.com/963093/29622934-c5070bd8-87f2-11e7-9e93-886c5b1f6399.png">

i am pretty sure this is a regression in beakerx

PR #5877, opened 2017-08-24
#5875: fix tables with big integers

jaroslawmalekcodete:


scottdraves:
thanks this fixes the crash but i am still seeing bits lost in the 64-bit example:
screen shot 2017-08-24 at 11 20 07 am

bignum looks correct for comparison:
screen shot 2017-08-24 at 11 20 12 am

PR #5882, opened 2017-08-25
fix the error of build conda package on windows

EfimovVladimir:


EfimovVladimir:
#5720

EfimovVladimir:
verified on windows

PR #5884, opened 2017-08-25
5798 scala interfaces part deux

jpallas:
This is the second part of the changes proposed in #5798.  I had thought there might be more, but there were only two places this was an issue, `TableDisplay` and `Line`.  This cleans up `TODO`s in the Scala table doc/demo.


scottdraves:
awesome, thank you!

Issue #5749, opened 2017-07-26
check tables for memory leaks


scottdraves:

scottdraves:
see error on https://github.com/twosigma/beakerx/pull/5894

PR #5886, opened 2017-08-29
#5749 fix tableDisplay memory leaks

MariuszJurowicz:


scottdraves:
cool thx, works well but one small problem: I am not seeing the triangles that should appear when I click on a column to sort it.

also, please bring up to date with master.

PR #5889, opened 2017-08-29
5798 scala interfaces part 2a

jpallas:
In my previous PR, I said I had thought there were more of these.  I was right.  I somehow overlooked the use of `AnyVal` in `Text`, `ConstantLine`, and `Rasters`.  Verified that `BigInt` and `BigDecimal` now work with these (subject to limitations in the JSON transcoding).

PS I hope that github is smart enough to ignore the commits that were already merged.


scottdraves:
awesome thank you.

Issue #5895, opened 2017-08-30
scala should auto-display tables the same as groovy


scottdraves:
![screen shot 2017-08-30 at 11 36 42 am](https://user-images.githubusercontent.com/963093/29881086-8cf8a028-8d77-11e7-9b71-0fc5d153884a.png)

jpallas:
I've been playing with using the jvm-repr stuff to make Scala `Map`s display as tables the way Java's do.  I added an override of `configJvmRepr` and did this:
```
    Displayers.register(AbstractMap.class, new Displayer<AbstractMap>() {
      @Override
      public Map<String, String> display(AbstractMap abstractMap) {
        TableDisplay td = new TableDisplay(abstractMap);
        td.display();
        HashMap<String, String> displayMap = new HashMap<>();
        displayMap.put(MIMEContainer.MIME.HIDDEN, "");
        return displayMap;
      }
    });
```

I'm not sure if this is the best or right way to do it, but the same approach should work with the Scala `TableDisplay` class.

PR #5920, opened 2017-09-05
5895 Use jvm-repr for auto-display of Scala DisplayTables and Maps

jpallas:
Here's some code to do the Scala auto-display of DisplayTables and (immutable) Maps (#5895).

As noted in the comments, there is a limitation in the current jvm-repr that makes it work only with classes and not interfaces (or Scala traits, which map to JVM interfaces modulo concrete methods).  It's just lucky that there is an abstract class for Map.  I opened an issue in jvm-repr for this, but I doubt anything will change in the near future.

Minor accompanying changes to the doc notebooks, mostly removing redundant `.display()` calls and unnecessary imports.

### DIFF
--- a/doc/contents/scala/EasyFormScalaDemos.ipynb
+++ b/doc/contents/scala/EasyFormScalaDemos.ipynb
@@ -10,10 +10,12 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
+    "collapsed": true
+   },
    "outputs": [],
    "source": [
-    "val form : EasyForm = new com.twosigma.beakerx.scala.easyform.EasyForm(\"Form and Run\")\n",
+    "val form : EasyForm = new EasyForm(\"Form and Run\")\n",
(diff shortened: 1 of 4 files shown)


scottdraves:
Thanks Joe we have to think about this and see what reaction we get from jvm-repr.
As you know I want to switch our internal display dispatching to jvm-repr but if there's a problem with it working with scala I don't want to stretch it.

jpallas:
That's reasonable.  I didn't necessarily expect this to be merged immediately, but I wanted to put it out in public, since it seems to work with what we have today.

rdblue:
We've added support in Toree for jvm-repr. The code is here: https://github.com/apache/incubator-toree/blob/master/scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaDisplayers.scala

The only annoyance was that Scala maps don't implement java.util.Map, but you just have to add `.asJava`.

scottdraves:
done this way: https://github.com/twosigma/beakerx/pull/5947

Issue #5815, opened 2017-08-09
implement time magics


scottdraves:
%timeit %%timeit %time (only the basic options)

http://ipython.readthedocs.io/en/stable/interactive/magics.html

part of https://github.com/twosigma/beakerx/issues/5048

PR #5934, opened 2017-09-06
#5815 %timeit %%timeit %time %%time magic commands

michalgce:
- [x] should not produce outputs for each run. seems it should suppress results completely.
- [x] also the numbers don't seem right, i am sleeping for 1s but %%timeit is reporting 2s.
- [x] please provide demo notebook showing all features.

### DIFF
--- a/doc/contents/TimeExecutionMeasurementWithMagicCommands.ipynb
+++ b/doc/contents/TimeExecutionMeasurementWithMagicCommands.ipynb
@@ -1,0 +1,150 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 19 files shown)


scottdraves:
1. should not produce outputs for each run.  seems it should suppress results completely.  
2. also the numbers don't seem right, i am sleeping for 1s but %%timeit is reporting 2s.
3. please provide demo notebook showing all features.

this PR:
screen shot 2017-09-06 at 11 02 31 am

Python3:
screen shot 2017-09-06 at 11 04 53 am

PR #5957, opened 2017-09-08
Create folders to organize examples by kernel

benmccann:
This puts all `.csv` files in a directory named `data`, all images in a directory named `images`, and all notebooks in a folder corresponding to their respective kernel


scottdraves:
Thanks.  That looks good except there are a lot of links between notebooks that also need to be updated, especially StartHere.ipynb

benmccann:
Thanks for the review Scott. I've updated the links in the PR. I had a couple other questions for you before you merge.

The files in nbviewer are duplicates and none are linked from `StartHere`. Can they be deleted? E.g. the two files below have equivalent contents:
https://github.com/twosigma/beakerx/blob/master/doc/contents/categoryPlot.ipynb
https://github.com/twosigma/beakerx/blob/master/doc/contents/nbviewer/NbviewerCategoryPlot.ipynb

I've organized this by language as a start, but I'm thinking it would make more sense to organize according to what each notebook is trying to demonstrate. This would match the headings in `StartHere.ipynb`. We would still have a folder for each language still as many of the notebooks are attempting to demonstrate usage of those language. Perhaps the list of folders would be better as:

- automation
- autotranslation
- clojure
- cpp
- forms
- display
- groovy
- java
- kotlin
- magics
- plotting
- python
- resources
- scala
- sql
- tablesaw

The other thing I was wondering is if I should get rid of the top-level `contents` folder. It doesn't serve much purpose and just adds an extra layer. You can see the current state of the directory tree with this PR here: https://github.com/benmccann/beakerx/tree/organize/doc

Issue #5850, opened 2017-08-16
bad try block in setupbase.py


scottdraves:
https://github.com/twosigma/beakerx/blob/master/beakerx/setupbase.py#L320

why is this try/except block needed to report errors in the code that it contains?
seems wrong.

ideally we would fix how these commands are being defined so they don't fail silently.
if that's not possible, then all of our commands need a try/except block.

PR #5963, opened 2017-09-11
#5850 Bad try block in setupbase.py

rbidas:


beakerx/setupbase.py
```diff
@@ -317,7 +317,6 @@ class InstallKernels(BaseCommand):
         description = 'Install all kernels in a directory'
 
         def run(self):
-            try:
```
scottdraves, line 320: the code inside the block has to be reindented

PR #5967, opened 2017-09-11
fix error in bkr2ipynb.py

EfimovVladimir:


scottdraves:
Thanks, but his blocks Linux and Mac users from accessing files with stars.
I think it should be solved at the windows shell level, or maybe making this call only on windows.
Not sure yet.

Issue #5968, opened 2017-09-11
easyform problems


scottdraves:
Looks like some major problems have crept into our EasyForm implementation.
Run the "All Kinds of Fields" cell in contents/EasyFormDemos.ipynb.

1) the two combo box fields do not show up (a bunch of JS console errors do), and
2) if you select an item in any of the 3 List widgets, and then click Go, the resulting value printed never changes.

PR #5969, opened 2017-09-12
#5968 fix missing jquery-ui widgets

MariuszJurowicz:


scottdraves:
this fixed the combo boxes but not the list components...

Issue #5930, opened 2017-09-05
table menus cropped


scottdraves:
![screen shot 2017-09-05 at 6 24 00 pm](https://user-images.githubusercontent.com/963093/30086088-7006c66c-9267-11e7-9a11-7c18f0f0c892.png)

scottdraves:
column menus fixed, still needs table menu.

PR #5975, opened 2017-09-13
#5930 reimplement menu with @phosphor Menu class

MariuszJurowicz:


scottdraves:
Cool that works great for the column menus, but the main table menu is still cropped.
It has to be converted as well.  However I am going to merge this as it is to lock in these gains so you can switch to https://github.com/twosigma/beakerx/issues/5970

Issue #5859, opened 2017-08-17
use python string templates instead of ad hoc for classpath in kernelspecs


scottdraves:
https://github.com/twosigma/beakerx/blob/master/beakerx/beakerx/install.py#L43
this means using something other than `__PATH__` in the kernel.json files.

scottdraves:
`__MAIN_CLASS__` should be done similarly.

PR #5981, opened 2017-09-15
#5859 Use python string templates instead of ad hoc for classpath in kernelspecs

rbidas:


kernel/base/kernel.json
```diff
@@ -1,5 +1,5 @@
 {
-  "argv": [ "java", "-jar", "__PATH__", "{connection_file}" ],
```
scottdraves: i think this file should be deleted since the base kernel never runs.

kernel/base/kernel_debug.json
```diff
@@ -1,5 +1,5 @@
 {
```
scottdraves: ditto, delete.

kernel/build.gradle
```diff
@@ -57,7 +57,7 @@ subprojects {
     ext.executeStaticContent = true
     doLast {
       if (executeStaticContent) {
-        def kernelJson = file('kernel.json').text.replace('__MAIN_CLASS__', mainClassName)
+        def kernelJson = file('kernel.json').text.replace('$MAIN_CLASS', '"' + mainClassName + '"')
```
scottdraves: why do we have a variable for main class at all?
since we have a kernel.json file for each kernel and we know the class for each kernel, this substitution is not needed right?  just put the class name into the kernel.json.

beakerx/setupbase.py
```diff
@@ -323,14 +324,10 @@ def install_kernel(source_kernelspec='', kernelspec_name=None):
                     classpath = (os.path.abspath(os.path.join(target_dir, 'base', 'lib', '*')) + (';' if sys.platform == 'win32' else ':') + os.path.abspath(os.path.join(target_dir, name, 'lib', '*'))).replace('\\', '/')
```
scottdraves: this should use json.dumps and OS independent logic the same as install.py does.

beakerx/beakerx/install.py
```diff
@@ -67,15 +72,10 @@ def _install_kernels():
...
-        # workaround #5864
-        if sys.platform == 'win32':
-            classpath = classpath.replace('\\', '/')
-        # #5859: replace with string.Template, though this requires the
-        # developer install to change too, so not doing right now.
+        classpath = json.dumps(os.pathsep.join([base_classpath, kernel_classpath]))
         template = pkg_resources.resource_string(
             'beakerx', os.path.join('static', 'kernel', kernel, 'kernel.json'))
-        contents = template.decode().replace('__PATH__', classpath)
+        contents = Template(template.decode()).substitute(PATH="\"" + classpath + "\"")
```
scottdraves: this doesn't look right, json.dumps returns a value with quotes already.

kernel/base/kernel.json
rbidas: done.

kernel/base/kernel_debug.json
rbidas: done.

beakerx/beakerx/install.py
rbidas: done

beakerx/setupbase.py
rbidas: done.

kernel/build.gradle
rbidas: done.

beakerx/beakerx/install.py
```diff
@@ -32,10 +33,14 @@ def _all_kernels():
         'beakerx', os.path.join('static', 'kernel'))
     return [kernel for kernel in kernels if kernel != 'base']
 
+def convert_path(path):
```
scottdraves: this should not be needed, that is the main point of this PR.
the reason for converting backslash to forward slash is because the old code was not adding quotes as needed.
but now that we are using json.dumps, it should insert the quotes for us, so we should not need to have special case handling for windows.

kernel/groovy/kernel.json
```diff
@@ -1,6 +1,6 @@
 {
   "argv": [
-    "java", "-cp", "__PATH__", "__MAIN_CLASS__",  "{connection_file}"
+    "java", "-cp", ${PATH}, "com.twosigma.beakerx.groovy.kernel.Groovy",  "{connection_file}"
```
scottdraves, line 3: thanks these look good now

Issue #5906, opened 2017-08-31
onClick of CombinedPlot fails


scottdraves:
run this cell:
```
def cplot = new CombinedPlot()
cplot << new Plot(title: "Top")
def p = new Plot(title:"Bottom")
p << new Bars(x:(1..5), y:[3,5,2,3,7]).onClick({info -> println(info)})
cplot.add(p)
cplot
```
click on a bar and get this crash:
```
Uncaught ReferenceError: plotService is not defined
    at Object.onClick (combinedPlotScope.js:162)
    at SVGRectElement.<anonymous> (plotScope.js:568)
    at SVGRectElement.<anonymous> (d3.js:934)
```

PR #5984, opened 2017-09-18
#5906 implement onClick and onKey handlers for combined plots

MariuszJurowicz:


scottdraves:
thx, the crash is fixed, but it's not running the onClick handler (which prints output that should show up below the plot)

Issue #6002, opened 2017-09-19
bad table display from python


scottdraves:
make this file tab.csv with these contents:
```
"a"
"1"
```
make a python3 notebook and run this cell:
```
import pandas as pd
import beakerx
from beakerx import *
pd.read_csv('tab.csv')
```
it crashes with `ValueError: Circular reference detected`

instead it should show a table.  looking at the code, I would say the ObjectEncoder class in utils.py looks wrong, especially the case for dictionaries is weird code.  there has to be a normal way to do this.  also, isn't it missing an else: clause?

PR #6007, opened 2017-09-21
#6002 Bad table display from python

rbidas:


scottdraves:
Thanks, this is works better, it doesn't crash, and I get a table.
But it also converts the integer to a double, which is wrong.
Here are two screenshots showing the correct results, as run by classic Beaker.
First, a regular integer like "1":
screen shot 2017-09-21 at 9 14 49 am
And then a large integer that would not be exactly representable with a double:
screen shot 2017-09-21 at 9 13 42 am
see doc/contents/bigInts.ipynb for how it works in groovy.

PR #6013, opened 2017-09-22
#5930 reimplement table index column menu

MariuszJurowicz:


scottdraves:
Thanks, this fixes the truncation but it introduces a problem. As you know, the old table menu item "Show Column" had some special behavior.  We need to keep that. When the user clicks on the items for the columns, the old menu stayed open. The new menu closes on click. It should stay open until the mouse leaves the menu, or the user hits ESC. Possible?

PR #6019, opened 2017-09-25
Clojure demo with %classpath

michalgce:


scottdraves:
Thanks for the PR, it works, but it introduced a typo into the 3rd cell "do." which now fails,
and there's a typo in the heading "toload" -> "to load"

michalgce:
@scottdraves fixed

Issue #6020, opened 2017-09-25
lost precision


scottdraves:
in python and in groovy, when a csv with a large integer is read, precision is lost due to fp.
here's the file:
```
"a"
"123456789123456789"
```
and the result of reading in groovy:
![screen shot 2017-09-25 at 9 49 52 am](https://user-images.githubusercontent.com/963093/30811885-6c12e386-a1d7-11e7-9cf0-a8c115c37cf9.png)
and python:

![screen shot 2017-09-25 at 9 56 38 am](https://user-images.githubusercontent.com/963093/30812069-dd6031ce-a1d7-11e7-9bab-7d7d555554ff.png)

correct result without our widget:
![screen shot 2017-09-25 at 9 55 22 am](https://user-images.githubusercontent.com/963093/30811994-ba39a8e2-a1d7-11e7-94b1-e068626c8e2a.png)

scottdraves:
done for groovy

PR #6032, opened 2017-09-26
#6020 lost precision (python)

rbidas:


beakerx/beakerx/tabledisplay/tabledisplay.py
```diff
@@ -91,7 +93,10 @@ def convertType(self, object_type, value):
         if object_type == "float64":
             return "double"
         if object_type == "int64":
-            return "integer"
+            if value > (pow(2, 31) - 1) or value < -(pow(2, 31)):
```
scottdraves: use sys.maxint etc, do not compute these values.

scottdraves: unclear to me that it's the right value at all.  js supports 56 bits before we have to switch.

scottdraves: use a standard constant for this eg https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.iinfo.html
if pandas has one that's better.

rbidas: done. numpy used.

Issue #5819, opened 2017-08-09
index column should have format submenu


scottdraves:
for times that should allow days/hours/minutes/seconds etc,
for integers formatted or not.
same as a regular column.

PR #6069, opened 2017-10-03
#5819 add format menu item to index column menu

MariuszJurowicz:


beakerx/js/src/tableDisplay/tableHeaderMenu/createFormatMenuItems.ts
```diff
@@ -0,0 +1,123 @@
+import MenuItem from "./MenuItemInterface";
```
scottdraves, line 17: needs banner

Issue #6064, opened 2017-10-02
easyform text area component size


scottdraves:
should be measured in chars and lines not pixels
![screen shot 2017-10-02 at 2 01 26 pm](https://user-images.githubusercontent.com/963093/31091835-4473b492-a77a-11e7-91a0-5af0e46c303d.png)
b should be the same width as a, and also 20 lines high.

PR #6076, opened 2017-10-04
#6064 fix setting the Textarea width and height

MariuszJurowicz:


MariuszJurowicz:
@scottdraves maybe we need the python `TextareaWidget` to be updated as well?

scottdraves:
i am not seeing the right width for the TextArea and the TextField:
screen shot 2017-10-04 at 9 54 58 am

scottdraves:

scottdraves:
actually the text field width seems broken on master as well

scottdraves:
added https://github.com/twosigma/beakerx/issues/6080
and https://github.com/twosigma/beakerx/issues/6079

Issue #6057, opened 2017-09-29
EasyForm CheckBox label on wrong side


scottdraves:
![screen shot 2017-09-29 at 7 40 59 pm](https://user-images.githubusercontent.com/963093/31039741-30da854c-a54e-11e7-9080-64da2c69f59a.png)

MariuszJurowicz:
@scottdraves 
The html structure is implemented in https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/controls/src/widget_bool.ts `CheckboxView` and i think it's not something configurable to change the position of the label. So the solution would be the dirty css hack, the js hack or overwritten `CheckboxView` `render` method. I can't recommend any.
Maybe we can ask Json to add the possibility of changing the position of the label?
The current implementation places the `input` inside the `label` node and appends it to next to the regular `widget-label` node.

scottdraves:
Just implement our own.

PR #6077, opened 2017-10-04
#6057 fix rendering standalone checkboxes

MariuszJurowicz:


MariuszJurowicz:
@scottdraves maybe we need the python `CheckboxWidget` to be updated as well?

PR #6091, opened 2017-10-05
Move scala sources to src/main/scala from src/main/java

jpallas:
It's just a convention.  Gradle doesn't care.  IntelliJ doesn't care.  Maybe some tool somewhere cares.  My low-grade OCD cares.  :-)


scottdraves:
thanks!

Issue #6084, opened 2017-10-04
python easyform component widths


scottdraves:
should be measured in characters, not pixels.
and small values are just vanishing...
![screen shot 2017-10-04 at 4 08 54 pm](https://user-images.githubusercontent.com/963093/31197235-7d7c943e-a91e-11e7-9554-8d2bca8deb14.png)

scottdraves:
ditto for addTextArea and its height parameter.

Issue #6074, opened 2017-10-03
Python plots not publishable


scottdraves:
Run a Python cell like:
```
from beakerx import *
Plot(title="published")
```
and then publish the result. 
 you will get a page like:
https://nbviewer.jupyter.org/gist/anonymous/b425cb708788ec2d0746d11b36744f2e
which does not show an interactive plot and has the following errors on the console:
```
Failed to load resource: the server responded with a status of 404 ()
beakerx.js Failed to load resource: the server responded with a status of 404 ()
index.js Failed to load resource: the server responded with a status of 404 ()
require.js?v=6da8be361b9ee26c5e721e76c6d4afce:140 Uncaught Error: Script error for "https://unpkg.com/beakerx@1.0.0/dist/index.js"
http://requirejs.org/docs/errors.html#scripterror
    at makeError (require.js?v=6da8be361b9ee26c5e721e76c6d4afce:165)
    at HTMLScriptElement.onScriptError (require.js?v=6da8be361b9ee26c5e721e76c6d4afce:1732)
index.js Failed to load resource: the server responded with a status of 404 ()
require.js?v=6da8be361b9ee26c5e721e76c6d4afce:140 Uncaught Error: Script error for "https://unpkg.com/@jupyter-widgets/base@1.0.0/dist/index.js"
http://requirejs.org/docs/errors.html#scripterror
    at makeError (require.js?v=6da8be361b9ee26c5e721e76c6d4afce:165)
    at HTMLScriptElement.onScriptError (require.js?v=6da8be361b9ee26c5e721e76c6d4afce:1732)
```
the equivalent groovy code works fine:
https://nbviewer.jupyter.org/gist/anonymous/e7b3d317724dd949530d17db7568030b

Issue #6062, opened 2017-10-02
EasyForm (python) CheckBox doesn't have label


EfimovVladimir:
test code:
```
f7 = EasyForm("form7")
f7.addCheckBoxes("field name7", ["one", "two", "three"])
f7
```

![image](https://user-images.githubusercontent.com/17744188/31082433-a461b788-a797-11e7-81b4-df935e9ddaf0.png)

PR #6111, opened 2017-10-10
#6084 #6062 #6074 Easy form changes and Python plots not publishable

rbidas:
Cover:
#6084
#6062
#6074


scottdraves:
text field is not the right width:
screen shot 2017-10-10 at 11 12 50 am

actually this seems to be a JS problem, groovy does the same.

scottdraves:
check box label is on wrong side:
screen shot 2017-10-10 at 11 14 11 am

PR #6128, opened 2017-10-11
#6127 fix opening the table colum menu outside jupyter's cell

MariuszJurowicz:


scottdraves:
in jupyter app seems to cause menus to be cropped again
screen shot 2017-10-11 at 10 42 56 am

Issue #6136, opened 2017-10-12
2nd click should close menu


scottdraves:
if you click the hamburger, the table menu opens.
if you click again in the exact same place, the menu shouild close.
i see them just blink (close and open again).
ditto for column menus.

PR #6141, opened 2017-10-13
#6127, #6136 fix duplicated menus, fix positioning submenu, toggle menu

MariuszJurowicz:


scottdraves:
menus open and close well now but clicking on column header to sort now does nothing...

PR #6162, opened 2017-10-18
#5998 add webpack production build

MariuszJurowicz:


beakerx/js/webpack.prod.js
```diff
@@ -0,0 +1,24 @@
+const webpack = require('webpack');
```
scottdraves, line 17: needs banner

scottdraves:
this works fine and makes the JS a lot smaller (about half). but it does not seem to be called by setup.py, is that part remaining to do?

MariuszJurowicz:
I've tweaked the webpack config a bit and set the production build as the default one.
For development you can use the `yarn build:dev` instead

PR #6171, opened 2017-10-20
fix e2e tests

EfimovVladimir:


test/README.md
```diff
@@ -21,5 +21,5 @@ Verify password: beakerx
 ## Run the tests
 
 ```
-python ./run_notebook.py && python ./run_tests.py
+source activate beakerx && python ./run_tests.py
```
scottdraves, line 24: i don't think you should hardwire the environment name.
can't you just assume this the user sets it up in advance?

Issue #6156, opened 2017-10-18
docker version missing version strings


scottdraves:
<img width="320" alt="screen shot 2017-10-17 at 10 33 22 pm" src="https://user-images.githubusercontent.com/963093/31698393-645ad41c-b38b-11e7-8295-df207f6480e3.png">

PR #6174, opened 2017-10-20
#6156 Removed dockerignore. Addded cleaning up after install.

michalgce:


scottdraves:
```
rm: cannot remove '.DS_Store': No such file or directory
rm: cannot remove '.idea/': No such file or directory
```
i got these errors.  probably best to use -f option, and put all arguments into one cmd.

scottdraves:
i confirmed the version is displayed correctly in the container 👍

Issue #5678, opened 2017-07-16
make a UI for setting JVM options


scottdraves:
heap size and properties are set via environment variables set before jupyter is launched.
but there should be a nice UI, maybe a tab on the tree page for setting these.  we can keep envars as the underlying mechanism.

PR #6182, opened 2017-10-23
#5678 Make a UI for setting JVM options 

rbidas:


beakerx/beakerx/static/html/settings_tab.html
```diff
@@ -0,0 +1,19 @@
+<div id="beakerx" class="tab-pane">
```
scottdraves, line 16: needs banner

scottdraves:
1) Tab heading should be "BeakerX" not "Beakerx Sett."
2) At the top of the tab, "Beakerx settings" should say "[BeakerX](http://BeakerX.com) from [Two Sigma Open Source](http://opensource.twosigma.com/) version X.Y.Z"
3) at the top of the box "JVM" -> "JVM Options"
4) there should always be a box for heap size.  If the box is empty then the default is used (no parameter set).  label on box is "Heap Size in GB".  form should verify they enter a number here, otherwise provide popup error.
5) there should be 2 "+" buttons one that says "add property" and gives you two fields, the name of the java property, and its value.  the other should say "add other option".
6) at the bottom it should show preview of the whole command line that will be used.

beakerx/beakerx/static/html/settings_tab.html
rbidas: Done

scottdraves:
Cool, the heap size works great :)

However you should check that the value they enter is legal number without spaces etc.
If the value is not legal then the "Submit" button should be disabled.
Also I don't believe the ":" is needed so remove that.

But then I tried to define a property, so I clicked +, and entered "myprop" and "100".
This resulted in the following cmd line:
> Result: java -myprop:100 -Xmx:10g

Which is illegal and so the JVM just crashes when I try to start a notebook.

1) then the user clicks +, do not have a popup, just add 2 fields to the page.
the user can fill them out in place.

2) it should say -Dmyprop=100 as the argument so it does not crash.

3) The "Submit" button should be labeled "OK"

4) each property and other argument that you add should be displayed with a "x" next to it, if you click the "x" then that property is removed.

beakerx/beakerx/static/tree.js
```diff
@@ -0,0 +1,304 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+define(function (require) {
+    var $ = require('jquery');
```
scottdraves: JS should be 2 space indent

scottdraves:
In the basic text "BeakerX from Two Sigma Open Source version ..."
the links should open in a new tab.

scottdraves:
also, we need to remember this configuration from run to run.
please save to ~/.jupyter

rbidas:
Done

scottdraves:
Cool thanks, we are getting there :)  Still to do:
* allow a floating point number in the GB (eg "5.5").  If they do use decimal like this, then multiply it by 1024 and use the MB version of the java option.
* put the "Heap Size in GB" before "add property"
* for the properties, need to add the `-D` and the `=` to the option, so this works:

screen shot 2017-10-30 at 2 45 44 pm
that should result in an option `-Dmyprop=myval`
* when you add a new property, the left field should be labeled "name" and the right field "value".

beakerx/beakerx/environment.py
```diff
@@ -0,0 +1,87 @@
...
+from os import environ, path
+
+
+class EnvironmentSettings:
+    suffix_other = '_other_'
+    suffix_java = '_java_'
+    _base_var_name = 'beakerx_java_arg'
+    other_var_name = _base_var_name + suffix_other
+    java_var_name = _base_var_name + suffix_java
+    path = '~/.jupyter/beakerx.json'
```
scottdraves: there should be a better way to make this path, using an internal jupyter API.

scottdraves: https://github.com/jupyter/jupyter_core/blob/master/jupyter_core/paths.py#L40

rbidas:
Done

scottdraves:
cool!  but i just found a big problem:
it's adding these arguments to all the kernels, including python, which makes it just crash.
they should only be added to ours, or those that start with "java".

beakerx/beakerx/kernel_spec.py
```diff
@@ -19,5 +19,6 @@
 class BeakerXKernelSpec(KernelSpec):
     def __init__(self, **kw):
         super(BeakerXKernelSpec, self).__init__(**kw)
-        args = EnvironmentSettings.read_beakerx_env_settings()
-        self.argv[1:1] = args
+        if 'java' in self.argv:
```
scottdraves: should only check for argv[0] is java.  another argument could randomly obscurely be "java"

rbidas: done

Issue #5641, opened 2017-07-12
clojure hangs, maybe a race condition


scottdraves:
<img width="551" alt="screen shot 2017-07-11 at 11 02 33 pm" src="https://user-images.githubusercontent.com/963093/28099834-c0736986-668c-11e7-8927-bf068c063128.png">

PR #6184, opened 2017-10-23
#5641: kill clojure threads

scottdraves:


scottdraves:
👍  works great.  can it be done with one call to runCode?

PR #6221, opened 2017-10-31
Factor out isNanoPlot condition test and make it work for subclasses

jpallas:
The plot-type test for NanoPlot uses class equality, which won't work if we subclass NanoPlot to add language-specific methods and/or do anonymous subclass initialization (e.g. for Scala).


scottdraves:
awesome, thank you.

PR #6231, opened 2017-11-03
6207: replace ivy to embedded mvn

jaroslawmalekcodete:


scottdraves:
The first time I built this it got an error downloading the zip and then it would not build and I had to delete a file and build again and then it worked great.  So I am concerned about that, I prefer we get the dependency from the env, like in https://github.com/twosigma/beakerx/pull/6215

Let's try again to debug that one on CI.

Issue #6202, opened 2017-10-25
[Groovy] println encoding


altavir:
Yet another encoding problem.
Java (and consequently groovy) uses default system encoding for `print` operations. So when system default encoding is not UTF and I try to write `println("привет миру")` in groovy, I get `???...`. The obvious solution is change JVM default encoding, but it could not be done after start.

PR #6240, opened 2017-11-06
#6202: fix println encoding problem

jaroslawmalekcodete:


scottdraves:
@altavir can you confirm this fixes the problem for you?

scottdraves:
altavir confirmed via gitter.

Issue #6239, opened 2017-11-06
jvm options validation issues


scottdraves:
1) allow properties with empty (length 0) value.  right now if i don't enter a value then no option is displayed and set, and if i hit OK then it just vanishes.
2) allow heap sizes that do not start with a 0, eg ".5" should work.  use IsNan and + as described here: https://stackoverflow.com/questions/175739/is-there-a-built-in-way-in-javascript-to-check-if-a-string-is-a-valid-number
do not make your own regexp.

Issue #6234, opened 2017-11-03
JVM options improvements


scottdraves:
1) There is an error message that shows up:
> All options should start with -X or -D

lots of JVM options start with other stuff (eg -ea) , so please remove this.

2) the beakerx.json files looks like:
```
{"payload": {"other": ["-Xms10m"], "-Xmx": "5", "jvm": {"another": "10", "myprop": "myval"}}}
```
please rename the fields as follows:
-Xmx -> heap_GB
jvm -> properties
payload -> beakerx
and make a section for jvm_options that contains other, heap_GB, and properties.  ie it should look like:
```
{"beakerx": {"jvm_options":{"other": ["-Xms10m"], "heap_GB": "5", "properties": {"another": "10", "myprop": "myval"}}}}
```

3) please pretty-print the file so it's easier to read.

4) if you have a beakerx.json like the above set, and then you do a build, then during beakerx-install the following error occurs:
```
  File "/Users/spot/src/tmp/beakerx/beakerx/beakerx/environment.py", line 67, in set_beakerx_env_settings
    EnvironmentSettings.clear_beakerx_env_setting()
  File "/Users/spot/src/tmp/beakerx/beakerx/beakerx/environment.py", line 94, in clear_beakerx_env_setting
    for x in environ:
  File "/Users/spot/anaconda/envs/beakerx/lib/python3.6/os.py", line 688, in __iter__
    for key in self._data:
RuntimeError: dictionary changed size during iteration

5) why are envars being used at all?  seems like an unneeded complication.
```

PR #6242, opened 2017-11-06
#6234 #6239 JVM options improvements

rbidas:


beakerx/beakerx/environment.py
```diff
@@ -14,20 +14,31 @@
 
 from os import environ, path
 from jupyter_core import paths
+import json
+
+sadsd = "" \
```
scottdraves: what is this?

beakerx/beakerx/environment.py
```diff
@@ -14,20 +14,31 @@
...
+        ""
+default_config = "{" \
```
scottdraves: use triple quotes

beakerx/beakerx/environment.py
rbidas: removed

beakerx/beakerx/environment.py
rbidas: done

PR #6249, opened 2017-11-07
Support plotting arrays

benmccann:
This will make it easier to use Tablesaw with BeakerX. E.g. the code in https://github.com/twosigma/beakerx/pull/6248 could be written a little more concisely after this change


scottdraves:
thanks

Issue #6105, opened 2017-10-09
`sudo pip install beakerx` has some wrong permissions


dsblank:
With latest (0.5.1) I was able to fix with:

```
sudo chmod -R a+r /usr/share/jupyter
sudo chmod a+x /usr/share/jupyter/kernels/*
```

scottdraves:
Hi Doug thanks for the report, what OS did you have this problem on?

dsblank:
Linux (Ubuntu 16.04 LTS 64-bit). Thanks!

scottdraves:
Hi Doug we made 2 fixes:
1) the python package now has python_requires set to >=3, so that you would get an immediate error before it even would try to install into you the python2 env that comes with ubuntu.
2) the permissions of the kernels now respect the umask.  we were creating a temporary directory for staging the kernel before install.  temporary directories are created only readable by the user who created it, and the permissions were copied too.  we now create a regular directory inside the temporary directory, and install from that.

please LMK if this doesn't work for you.

PR #6251, opened 2017-11-07
#6105 "sudo pip install beakerx" has some wrong permissions

rbidas:


scottdraves:
1) it looks like "jupyter kernelspec install" is creating directories with mode 700 and disrespecting my umask which should make them 755.  we should figure out the root cause of this.  there may be a better fix than chmod after the fact.

2) use version constraint in setup.py to avoid running python2.

scottdraves:
Digging into the code, I guess the problem is here:
https://github.com/twosigma/beakerx/blob/master/beakerx/beakerx/install.py#L93
tempfile.TemporaryDirectory uses mode 700.  to get normal modes from the user,
use os.mkdir.  can do this *inside* the TemporaryDirectory.

rbidas:
Done

beakerx/beakerx/install.py
```diff
@@ -90,15 +97,19 @@ def _install_kernels():
...
-        with tempfile.TemporaryDirectory() as tmpdir:
-            with open(os.path.join(tmpdir, 'kernel.json'), 'w') as f:
-                f.write(contents)
-            install_cmd = [
-                'jupyter', 'kernelspec', 'install',
-                '--sys-prefix', '--replace',
-                '--name', kernel, tmpdir
-            ]
-            subprocess.check_call(install_cmd)
+        tmpdir = os.path.join(gettempdir(), str(uuid.uuid4()))
```
scottdraves: please keep the tempfile.TemporaryDirectory(), do not try to reimplement it

beakerx/beakerx/install.py
```diff
@@ -90,15 +97,19 @@ def _install_kernels():
...
+        os.mkdir(tmpdir, 0o755)
```
scottdraves: please do not force a particular mode.  respect the user's umask.

beakerx/beakerx/install.py
```diff
@@ -90,15 +97,19 @@ def _install_kernels():
...
+        with open(os.path.join(tmpdir, 'kernel.json'), 'w') as f:
+            f.write(contents)
+        install_cmd = [
+            'jupyter', 'kernelspec', 'install',
+            '--sys-prefix', '--replace',
+            '--name', kernel, tmpdir
+        ]
+        subprocess.check_call(install_cmd)
+
+        shutil.rmtree(tmpdir, ignore_errors=True)
```
scottdraves: should not be needed if you use tempfile.TemporaryDirectory()

beakerx/setup.py
```diff
@@ -91,7 +91,9 @@
         'notebook >=4.4.0',
         'ipywidgets >=7.0.0',
         'pandas'
+
     ],
+    python_requires='>=3',
```
scottdraves, line 96: good

rbidas:
Done

PR #6266, opened 2017-11-08
Fix bad link

benmccann:
DeepLearning4J uses TableSaw internally, but I'm not sure you meant to link to it here @michalgce ?


michalgce:
No there should be a link to DL4J thanks 👍

Issue #6238, opened 2017-11-06
fix interactive recomputation


scottdraves:
see interactive.ipynb, try each example:
1) the widgets are vanishing.
2) use %import
3) don't fix API yet, but do fix #5871 
4) instead of straight lines, make the example be parametric curves with 100 samples.

PR #6270, opened 2017-11-08
#6238 Fixed interactive demo.

michalgce:
I decided to disable continuous_update for slider to provide user better experience during changing value


scottdraves:
cool works better.
i think setContinuous_update false is a problem, but need to see the parametric graph first.

michalgce:
Take a look on last cell in notebook

michalgce:
I've added new plot.

scottdraves:
ok the main thing this needs is interact() should run once when it's first creates the widgets.
in the last cell when i run it i get the widgets but the result is just "null" until i move the sliders the first time.  the plot appear from the get-go.

scottdraves:
and in the last example the sliders should be doubles.

scottdraves:
i tried making the sliders update continuously again and it didn't work.
i think that's needed.

scottdraves:
and actually, delete all the examples except the last.

michalgce:
[] first creates the widgets

scottdraves:
great it starts out rendered, thanks.

still needs
1) continuous update
2) floating point
3) delete all examples but last

PR #6295, opened 2017-11-11
Expand support for Java 8 date types

benmccann:
Fixes https://github.com/twosigma/beakerx/issues/6286

### DIFF
--- a/kernel/base/src/main/java/com/twosigma/beakerx/util/DateUtil.java
+++ b/kernel/base/src/main/java/com/twosigma/beakerx/util/DateUtil.java
@@ -1,0 +1,34 @@
+package com.twosigma.beakerx.util;
+
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.util.Calendar;
+import java.util.Date;
+
+public final class DateUtil {
+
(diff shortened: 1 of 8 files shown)


kernel/base/src/main/java/com/twosigma/beakerx/util/DateUtil.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.util.Date;
+
+public final class DateUtil {
+
+  public static long dateToLong(Object datelike) {
```
scottdraves, line 12: could overloading be used here to avoiding going back to Object?

scottdraves:
Thanks, yup

kernel/base/src/main/java/com/twosigma/beakerx/util/DateUtil.java
benmccann: It would result in more code duplication because we'd have to do the `instanceof` checks in multiple places to cast the input to the appropriate type to pass into the overloaded functions

scottdraves:
awesome! thank you.

PR #6306, opened 2017-11-13
revert f208361169414134360e89de00e57173f6d765c0 since it broke some p…

scottdraves:
…lotting demos


benmccann:
This seems really strange to me. Do you have any idea why it would have broken the samples? Maybe it's a Groovy thing I don't understand

scottdraves:
no i was also surprised and baffled, but since the feature did not seem to be used and i really need to cut a release since it's holding up a branch with a big important change i decided to just take it out without researching it.

benmccann:
@scottdraves I think it's because the existing `setX` and `setY` take `List<Object>` when they should take `List<?>` - https://stackoverflow.com/a/7065056

Would you be able to try adding the code back with that change?

scottdraves:
sure i made an issue for it.  0.8.0 is done so we have time for bigger changes.
thanks for looking that up.

Issue #6292, opened 2017-11-10
show YAxis even with no items


scottdraves:
```
p = new Plot() << new YAxis(label:"asdf")
l = new Line(y:[2, 5])
// l.setYAxis("asdf")
p << l
```
the axis should appear even with this commented out.

PR #6312, opened 2017-11-14
#6292 show yAxisR when set

MariuszJurowicz:
This fix should be heavily tested before merge. It changes the Plot right axis behavior.


scottdraves:
seems to break "Second Y Axis" in PlotFeatures.ipynb

scottdraves:
screen shot 2017-11-15 at 9 29 58 am

Issue #6247, opened 2017-11-06
add environment.yaml binder file 


scottdraves:
so that our documentation can be opened up live online.
similar to: https://github.com/imagej/tutorials/commit/e4cd9ad55e7cc60ad6684df986c5e450526a15df

ctrueden:
The relevant documentation which I followed was [here](http://repo2docker.readthedocs.io/en/latest/samples.html#conda-environment), along with the binder-examples like [this one](https://github.com/binder-examples/jupyter-rise/blob/c1fe3bd8afce4b9309762e489dab182ca70a5101/environment.yml).

PR #6320, opened 2017-11-15
#6247 Added environment.yaml

michalgce:
Result can be find here: https://mybinder.org/v2/gh/twosigma/beakerx/michal/6247

PR contains temporary workaround for https://github.com/twosigma/beakerx/issues/6319, after that we could remove ```postData``` file.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,6 @@
+name: beakerx
+channels:
+  - conda-forge
+dependencies:
+  - ipywidgets
+  - beakerx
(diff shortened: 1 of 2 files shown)


environment.yml
```diff
@@ -0,0 +1,6 @@
+name: beakerx
+channels:
+  - conda-forge
+dependencies:
+  - ipywidgets
+  - beakerx
```
scottdraves, line 6: please end files with newline

postBuild
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
```
scottdraves, line 1: yea let's wait until this bug is fixed.

Issue #6319, opened 2017-11-15
kernels fail without .jupyter directory


scottdraves:
when freshly installing beakerx, If i installed beakerx with conda, our kernel’s are invisible until i created ~/.jupyter directory.

PR #6327, opened 2017-11-16
#6319 kernels fail without .jupyter directory

rbidas:
Check on dev and mybinder environment.


piorek:
@scottdraves I've tested it locally and it fixed the issue form me

Issue #6314, opened 2017-11-14
fix docker image


scottdraves:
seems maven was missing but even after i added it:
```
-RUN conda install -y -c conda-forge openjdk
+RUN conda install -y -c conda-forge openjdk maven
```
then i tried to start a notebook i got errors on the console:
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/beakerx/.jupyter/beakerx.json'
```
and the kernel crashed.

scottdraves:
Also please install node via conda, not nodesource.com, to be consistent with the rest of our documentation.

scottdraves:
had to revert PR because it failed with:
```
Step 15/15 : RUN conda install -y -c conda-forge nodejs openjdk gradle
 ---> Running in 88a8b49af17e
Fetching package metadata .............

PackageNotFoundError: Packages missing in current channels:

  - gradle

We have searched for the packages in the following channels:

  - https://conda.anaconda.org/conda-forge/linux-64
  - https://conda.anaconda.org/conda-forge/noarch
  - https://repo.continuum.io/pkgs/main/linux-64
  - https://repo.continuum.io/pkgs/main/noarch
  - https://repo.continuum.io/pkgs/free/linux-64
  - https://repo.continuum.io/pkgs/free/noarch
  - https://repo.continuum.io/pkgs/r/linux-64
  - https://repo.continuum.io/pkgs/r/noarch
  - https://repo.continuum.io/pkgs/pro/linux-64
  - https://repo.continuum.io/pkgs/pro/noarch
```

PR #6331, opened 2017-11-16
#6314 Replace nodejs installation with conda.

michalgce:


scottdraves:
it should install nodejs with conda, where is that?

michalgce:
Opss, added.

docker/base/Dockerfile
```diff
@@ -44,13 +46,10 @@ echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.lis
...
-#    MiniConda   #
-##################
-RUN echo 'export PATH=/opt/conda/bin:$PATH' > /etc/profile.d/conda.sh && \
-    wget --quiet https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ~/miniconda.sh && \
-    /bin/bash ~/miniconda.sh -b -p /opt/conda && \
-    rm ~/miniconda.sh
+##########
+#  Node  #
+##########
+RUN conda install -y -c conda-forge nodejs
```
scottdraves: consolidate with the other conda install, it's best to minimize the number of Docker commands.
combine the apt-get and conda installs as much as possible.

scottdraves: https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices 
"Minimize the number of layers"

michalgce: GTK Thanks!

Issue #6344, opened 2017-11-18
Serializer for widget types can't handle anonymous subclasses


jpallas:
Many of the serializers set the `"type"` field by calling `Class.getSimpleName()`.  This fails if the object being serialized is an instance of an anonymous subclass.  Anonymous subclasses may be used in Java and Scala idioms to support terse initialization, such as

```java
new Points() {{
    setX(xValues);
    setY(yValues);
    setColor(Color.RED);
    setOutlineColor(Color.BLUE);
    setSize(20);
}}
```

Walking up the class chain to find the first class that is neither anonymous nor inner addresses the problem.

PR #6345, opened 2017-11-18
Serializer type name for anonymous classes

jpallas:
This addresses #6344.  A helper static method is introduced that walks up the class chain to find a usable class name.  Existing Java tests are modified and a Scala test is added.


scottdraves:
cool thanks i will test this shortly.

Issue #6245, opened 2017-11-06
Need better documentation for base in category items


jpallas:
(This was going to be a comment on #6237, to which it is still relevant, but it's a separate issue.)

I can't figure out how `setBase` is supposed to work for the Category items.  The examples for `CategoryBars` in the `categoryPlot` notebook show base being set to a single number, a list of numbers, a list of lists of numbers, and a mixed list of numbers and lists of numbers.  The single number case sets `baseBase` and the other cases set `bases`, which is declared `List<Number>`.  So, this is a case where the usage is inconsistent with the type declarations in the Java code, and trying to work backwards from the types to the intended interface doesn't work.

scottdraves:
thanks joe, uh idk let me get back to you.
@michalgce

scottdraves:
I agree the declared type does not match the arguments.
IDK why Groovy allows this.  The implementation is confusing with the two properties as well.

Seems it would be an improvement to change `setBases` to take `List<Object>`.
Does that resolve this for you?

jpallas:
Changing bases to be `List<Object>` would resolve the inconsistency, yes.  (Side note: I'm not sure if introducing a `BasedCategoryGraphics` subclass, similar to `BasedXYGraphics`, would be worthwhile.  It's only a few lines of common code.)  It won't help with the type-as-documentation aspect, sadly.

Given the way it's used, my dream of having a statically typed Scala interface that matches what is actually allowed seems out of reach.

scottdraves:
Yea this bit of the API isn't amenable to static typing, sorry, it wasn't designed with that in mind.
The problem is just the mixing of the lists and numbers within the list, ie allowing [1, [2, 3]].
I don't think this is very commonly used, so if it had a more verbose syntax that would be OK.
Could you define a helper class that used a subclass to workaround lack of explicit sum types?
SetBase method would be overloaded with the following types:
`Number`, `List<Number>`, `List<List<Number>>`, and `List<ListOrNumber>`?

scottdraves:
https://stackoverflow.com/questions/3508077/how-to-define-type-disjunction-union-types

jpallas:
[...]

PR #6359, opened 2017-11-21
#6245 extract common base class for category plotitems that have bases

jpallas:
This pulls out the common behavior, fixes the static type to reflect reality, and adds a runtime check that the values passed are of the expected types.  A test is also provided to verify and document legal and illegal combinations.

The category plot demos all still seem to work as expected.

#6245


scottdraves:
thanks!

Issue #6217, opened 2017-10-31
Gradle-like %classpath mvn?


altavir:
Is it possible to use a standard gradle-like syntax for maven imports. Meaning `commons-io:commons-io:2.5` instead of `commons-io commons-io 2.5`. The difference is purely aesthetic, but the first way is familiar for gradle users.

scottdraves:
Sure, that makes sense, thanks for the suggestion.

PR #6360, opened 2017-11-21
#6217 Gradle-like %classpath mvn

michalgce:


scottdraves:
sorry for the confusion but this should be an additional optional syntax, please keep all our documentation as it is, except add one cell that shows you can use the : if you want.  the magic should work either way.

scottdraves:
👍 
please bring up to date with master

PR #6365, opened 2017-11-21
Added badge to README.MD

michalgce:
We need to generate VERSION file (https://microbadger.com/labels Ad.2) which contains string with latest label of version, so i generate it with gradle file.

Badge will be updated after updated Dockerfile will be pushed into Docker Hub

### DIFF
--- a/kernel/base/build.gradle
+++ b/kernel/base/build.gradle
@@ -75,11 +75,23 @@ def gitGetHash() {
 
 def resourceDir = file("./src/main/resources")
 
+def gitGetCommitHash() {
+  return "git rev-list --tags --max-count=1".execute().text.trim()
+}
+
+def gitGetLatestTagVersion() {
+  def gitLatestTag = "git describe --tags " + gitGetCommitHash() + ""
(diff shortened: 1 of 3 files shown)


kernel/base/build.gradle
```diff
@@ -75,11 +75,23 @@ def gitGetHash() {
...
+}
+
+def mainDir = file("../../")
+
 task prepareBuildVerion() {
   doLast {
     new File(resourceDir, 'version').write(gitDescribe())
     new File(resourceDir, 'hash').write(gitGetHash())
     new File(resourceDir, 'build_time').write(new Date().format("yyyy-MM-dd HH:mm z"))
+    new File(mainDir, 'VERSION').write(gitGetLatestTagVersion())
```
scottdraves, line 94: why is this version different from the one computed by gitDescribe()?

michalgce: cause gitDescribe returns current branch name, 
```gitGetLatestTagVersion``` get latest Tag release (which currently is 0.8.1)

PR #6373, opened 2017-11-22
#6363 remove old nbviewer examples, fix current ones

MariuszJurowicz:


benmccann:
None of these are linked to from StartHere.ipynb. It seems like we should include them there to make them discoverable and add some context about what they intend to demonstrate

MariuszJurowicz:
We may add the manual on how to open those in nbviewer to see the widgets rendered as well.

scottdraves:
yes there is a missing piece.  the issue for getting these properly linked and explained is https://github.com/twosigma/beakerx/issues/6224

Issue #6299, opened 2017-11-12
image does not appear


scottdraves:
http://nbviewer.jupyter.org/github/twosigma/beakerx/blob/master/doc/contents/java_args.ipynb
the image does not appear.
nbviewer bug?

<img width="714" alt="screen shot 2017-11-11 at 11 20 24 pm" src="https://user-images.githubusercontent.com/963093/32695927-f6b90568-c736-11e7-9fac-42a1e11ea18e.png">

scottdraves:
how the nb appears if directly opened:
<img width="567" alt="screen shot 2017-11-11 at 11 22 41 pm" src="https://user-images.githubusercontent.com/963093/32695949-3e33b546-c737-11e7-9bb7-7e8078dce6be.png">

lmitusinski:
There is an issue created for nbviewer:
https://github.com/jupyter/nbviewer/issues/692

It seems that nbconverter cannot correctly convert attachment markups with --to html parameter and default configuration.

lmitusinski:
I found out that it also doesnt work on my beakerx instance if you try to download notebook as html.
As a soultion i tried to install "jupyter_contrib_nbextensions" and convert notebook using "html_embed" instead of "html" as --to parameter and it worked fine.

scottdraves:
https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/master/src/jupyter_contrib_nbextensions/nbconvert_support/embedhtml.py

lmitusinski:
PR in twosigma repo: https://github.com/twosigma/nbconvert/pull/3

scottdraves:
https://github.com/jupyter/nbconvert/pull/780

PR #6377, opened 2017-11-23
#6299 java_arg embedded image

michalgce:
Here is a result: https://nbviewer.jupyter.org/github/twosigma/beakerx/blob/54b8ed2eedc8370f1053f81f8d5b11bd7440a5a4/doc/groovy/java_args.ipynb


scottdraves:
i see that works but i am looking for how to make it work without working around the problem.
can this be fixed in nbviewer without changing the notebook?  is there already an issue filed on nbviewer for this?

michalgce:
@scottdraves I cannot find any issue about this in nbviewer repo.
It looks like ipynb has a problem with print image when data about it is saved in attachment section:
image

to make sure that is not connected with printing base64 i made a test and instead of paste image directly from clipboard i paste base64 code into cell and it seems to works correctly. 

image

and here is the result:
https://nbviewer.jupyter.org/github/twosigma/beakerx/blob/14b32da08255050eb75703b6ab32ebfd681f20b9/doc/groovy/java_args.ipynb

Issue #6323, opened 2017-11-15
logging


scottdraves:
what's the right way to get logging output?
we have a lot of slf4j logging but i have never seen the output of it.
where does it go?

PR #6383, opened 2017-11-24
#6323 Clean up logging.

michalgce:


kernel/base/build.gradle
```diff
@@ -39,7 +39,9 @@ dependencies {
   compile group: 'org.apache.commons', name: 'commons-lang3', version: '3.6'
   compile group: 'org.apache.commons', name: 'commons-text', version: '1.1'
   compile group: 'org.apache.maven.shared', name: 'maven-invoker', version: '3.0.0'
-  compile group: 'org.slf4j', name: 'slf4j-simple', version: '1.7.22'
+  compile group: 'org.slf4j', name: 'slf4j-simple', version: '1.7.25'
```
benmccann, line 42: You shouldn't have two logging implementations on the classpath. I would recommend switching this line from `slf4j-simple` to `slf4j-api`. That will allow you to use the slf4j logging API with the logback backend implementation

kernel/base/build.gradle
```diff
@@ -39,7 +39,9 @@ dependencies {
   compile group: 'org.apache.commons', name: 'commons-lang3', version: '3.6'
   compile group: 'org.apache.commons', name: 'commons-text', version: '1.1'
   compile group: 'org.apache.maven.shared', name: 'maven-invoker', version: '3.0.0'
-  compile group: 'org.slf4j', name: 'slf4j-simple', version: '1.7.22'
+  compile group: 'org.slf4j', name: 'slf4j-api', version: '1.7.25'
+  compile group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.3'
```
scottdraves: we need to stick with slf4j, not add logback.

kernel/clojure/src/main/resources/logback.xml
```diff
@@ -0,0 +1,27 @@
...
+      <pattern>%date %-5level [%class{0}.%M:%line] %msg%n</pattern>
+    </encoder>
+  </appender>
+
+  <appender name="FILE_CLOJURE" class="ch.qos.logback.core.FileAppender">
+    <encoder>
+      <pattern>%date %-5level [%class{0}.%M:%line] %msg%n</pattern>
+    </encoder>
+    <append>true</append>
+    <file>${user.home}/.jupyter/logs/jupyter-clojure-kernel.log</file>
```
scottdraves: don't write to this file, just write to stderr, like the python log from "jupyter notebook".

kernel/base/build.gradle
benmccann: Why do you say that? slf4j-api provides the logging API, so the way that users log a message doesn't change at all as a result of this PR. It's only the way that you configure logging that changes (with this change `logback.xml` is used to configure the logger).

From what I've seen, probably the most recommended way to do logging in Java is to use slf4j-api + logback. I contribute quite a bit to Play Framework, which is a popular Java web framework, and it probably took us a few releases over a couple of years to get logging to a place where it really worked well, We ended up using slf4j-api + logback there, so this seems like a fairly good change to me, since it aligns well with what I've seen from a good amount of experience with Java logging

scottdraves: i believe the log should go to stderr, so it appears with the rest of the jupyter log.  SimpleLogger in slf4j already supports that so another lib and a bunch of config files are not needed.

kernel/clojure/src/main/resources/simplelogger.properties
```diff
@@ -0,0 +1,9 @@
+# SLF4J's SimpleLogger configuration file
```
scottdraves: this is identical for all kernels so shouldn't it be in base instead of being duplicated?
also, all files need banner.

scottdraves:
i'm getting the following warning on startup:
```
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/Users/spot/src/tmp/beakerx/beakerx/beakerx/kernel/base/lib/slf4j-simple-1.7.25.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/Users/spot/src/tmp/beakerx/beakerx/beakerx/kernel/base/lib/slf4j-simple-1.7.22.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.slf4j.impl.SimpleLoggerFactory]
```

kernel/base/build.gradle
benmccann: These seem like separate issues. We have just as many config files in the current version of the PR with slf4j-simple as we did in the logback version of the PR. I saw your comment that we should try to reduce the number of config files. I'm guessing if we can do that for slf4j-simple then we can for logback as well. Also, logback can send logs to stderr. In terms of libraries, it's the same number of libraries on the classpath in both cases. I would generally prefer using logback to slf4j-simple still

michalgce:
@scottdraves you probably got cached downloaded dep please try to run gradle clean first

scottdraves:
looks good thx

Issue #6362, opened 2017-11-21
beakerxTest dirs leaking


scottdraves:
my java.io.tmpdir has a million directories like:
```
beakerxTest996896452129936091
beakerxTest997087850015478165
beakerxTest997197771838036941
beakerxTest998087701884264556
beakerxTest998241787285088430
beakerxTest998449102516708068
beakerxTest99893586885012043
beakerxTest999112053571151476
beakerxTest999685608223986928
```

PR #6397, opened 2017-11-27
delete beakerxTest dirs

EfimovVladimir:


EfimovVladimir:
#6362

scottdraves:
this looks like it will solve the problem but wouldn't it be better to call this function in the right place?
https://github.com/twosigma/beakerx/blob/356cf950582b9c79d68c29eab7eca1a9d8476bf2/kernel/base/src/main/java/com/twosigma/beakerx/evaluator/BaseEvaluator.java#L189
this might not be a test problem.  are the directories leaking even from normal usage?

scottdraves:
@jaroslawmalekcodete

jaroslawmalekcodete:
Instead of adding task to build.gradle (which should work of course but only for test run from console) we can try to add `path.toFile().deleteOnExit()` below https://github.com/twosigma/beakerx/blob/master/kernel/base/src/test/java/com/twosigma/beakerx/evaluator/EvaluatorTest.java#L70

Issue #6257, opened 2017-11-07
multiple classloaders breaks with this native lib


scottdraves:
![screen shot 2017-11-07 at 9 56 40 am](https://user-images.githubusercontent.com/963093/32499971-059a744c-c3a2-11e7-8404-35a255b02c23.png)

scottdraves:
The error is:
```
java.lang.ExceptionInInitializerError
	at org.nd4j.linalg.cpu.nativecpu.ops.NativeOpExecutioner.<init>(NativeOpExecutioner.java:53)
	at org.nd4j.linalg.factory.Nd4j.initWithBackend(Nd4j.java:6185)
	at org.nd4j.linalg.factory.Nd4j.initContext(Nd4j.java:6087)
	at org.nd4j.linalg.factory.Nd4j.<clinit>(Nd4j.java:201)
	at script15100665976831923612652.run(script15100665976831923612652.groovy:2)
	at com.twosigma.beakerx.groovy.evaluator.GroovyCodeRunner.run(GroovyCodeRunner.java:71)
Caused by: java.lang.RuntimeException: ND4J is probably missing dependencies. For more information, please refer to: http://nd4j.org/getstarted.html
	at org.nd4j.nativeblas.NativeOpsHolder.<init>(NativeOpsHolder.java:51)
	at org.nd4j.nativeblas.NativeOpsHolder.<clinit>(NativeOpsHolder.java:19)
	at org.nd4j.linalg.cpu.nativecpu.ops.NativeOpExecutioner.<init>(NativeOpExecutioner.java:53)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at java.lang.Class.newInstance(Class.java:442)
	at org.nd4j.linalg.factory.Nd4j.initWithBackend(Nd4j.java:6185)
	at org.nd4j.linalg.factory.Nd4j.initContext(Nd4j.java:6087)
	at org.nd4j.linalg.factory.Nd4j.<clinit>(Nd4j.java:201)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:348)
[...]

PR #6401, opened 2017-11-28
#6257:[native lib] skip resetting environment when adding jars and importing class

jaroslawmalekcodete:


scottdraves:
kick jenkins

Issue #6347, opened 2017-11-19
allow adding maven resolvers


scottdraves:
https://github.com/twosigma/beakerx/issues/6343
```
@GrabResolver(name='bintray', root='http://dl.bintray.com/6thsolution/apexnlp'),
```
make the equiv of this for `%classpath add mvn`

scottdraves:
`%classpath add repo ...`

jpallas:
Maybe a different subcommand than `add`?  The existing `add` commands both add things to the classpath, while this does not.  `%classpath config resolver`?

scottdraves:
good point.

PR #6403, opened 2017-11-28
#6347 Addding new repo to mvn config resolver.

michalgce:


kernel/base/src/main/java/com/twosigma/beakerx/kernel/magic/command/PomFactory.java
```diff
@@ -0,0 +1,60 @@
+package com.twosigma.beakerx.kernel.magic.command;
```
scottdraves, line 16: banner

kernel/base/src/test/java/com/twosigma/beakerx/kernel/magic/command/PomFactoryTest.java
```diff
@@ -0,0 +1,38 @@
+package com.twosigma.beakerx.kernel.magic.command;
```
scottdraves, line 16: all files

scottdraves:
looks good

Issue #6330, opened 2017-11-16
plots with specific range, show both ends


scottdraves:
run this cell:
```
new Plot(xBound: [7, 9], yBound: [6, 9]) << new Line(x: (1..8), y: (1..8)) << new Line(x: (8..10), y: [8, 7, 6])
```
<img width="632" alt="screen shot 2017-11-18 at 6 58 16 pm" src="https://user-images.githubusercontent.com/963093/32985868-77c29240-cc92-11e7-92e0-04cb26bca9b0.png">
in the result of the plot, "9" should be shown on both x and y axes.

piorek:
I have a  fix ready for specified case, but not in general.

Ticks and labels are based on grid lines. So in a specified case I've fixed it by including right|top grid line.
But let say I change:
```js
p = new Plot(xBound: [7, 9.4], yBound: [6, 9.4])
```
What then?

scottdraves:
doesn't need to be shown in that circumstance.  Please make a PR with what you have and I'll try it out.

piorek:
When i were looking if I've introduced any BC breaks i found out that:
- there are no ticks on right YAxis 
- top label is missing

I have to investigate this

![screenshot from 2017-12-01 15-31-41](https://user-images.githubusercontent.com/743617/33487467-9d839478-d6ad-11e7-96d1-63716f2c8a5c.png)

scottdraves:
missing ticks seems to be another issue. https://github.com/twosigma/beakerx/issues/6440

piorek:
fixed
![image](https://user-images.githubusercontent.com/743617/33553232-2cfcb30e-d8f8-11e7-9388-44353cc8ddfe.png)

PR #6435, opened 2017-12-01
#6330 plots with specific range, show both ends

piorek:


scottdraves:
if i pan the plot i am seeing a problem with cropping:
screen shot 2017-12-01 at 11 44 09 am

piorek:
issue with cropping fixed

Issue #5829, opened 2017-08-10
document column size limit


scottdraves:
big_table_test.ipynb

document that cells that are more than 1000 chars wide get truncated with ...

show a table with a column of really long strings that gets truncated like this, and explain it.

https://github.com/twosigma/beakerx/blob/e9b2e94800e504cd7c34e29a9fe0815d47db6966/beakerx/js/src/tableDisplay/tableScope.js#L2302

jmsdnns:
What path would you like for the doc?

scottdraves:
put it in here:
https://github.com/twosigma/beakerx/blob/master/doc/groovy/big_table_test.ipynb

jmsdnns:
Done. PR coming up.

PR #6441, opened 2017-12-01
#5829 Added documentation to describe max column lengths

jmsdnns:
Addresses #5829.

The change looks like this: 
screenshot from 2017-12-01 13-35-19


scottdraves:
kick jenkins

scottdraves:
🐬

Issue #6358, opened 2017-11-21
docker build has language warnings


scottdraves:
```
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
        LANGUAGE = "en_US.UTF-8",
        LC_ALL = "en_US.UTF-8",
        LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
```

scottdraves:
resulted in

```
    warning You are using Node "4.2.6" which is not supported and may encounter bugs or unexpected behavior. Yarn supports the following semver range: "^4.8.0 || ^5.7.0 || ^6.2.2 || >=8.0.0"

```

michalgce:
Fixed

PR #6448, opened 2017-12-04
 #6358 Fix docker language warnings.

michalgce:


scottdraves:
thanks, those warnings are gone but now we got a bunch like this:
```
/home/beakerx/kernel/base/src/main/java/com/twosigma/beakerx/kernel/magic/command/functionality/TimeMagicCommand.java:90: error: unmappable character for encoding ASCII
```

Issue #5643, opened 2017-07-12
make javadoc accessible


scottdraves:
like classic beaker does.
have a menu item that opens it up in a new tab.
run javadoc during build, and then just configure tornado to serve the right directory.

PR #6450, opened 2017-12-04
#5643 make javadoc accessible

rbidas:


beakerx/beakerx/static/html/settings_tab.html
```diff
@@ -15,15 +15,33 @@
...
-           <span id="beakerx_info" class="toolbar_info"></span>
+        <div class="col-xs-11 no-padding">
+            <span id="beakerx_info" class="toolbar_info"></span>
         </div>
-    </div><br/>
+        <div class="col-xs-1">
+           <span id="beakerx_java" class="toolbar_info">
+
+            <a class="beakerx_javadoc_link" href="beakerx/javadoc/index.html" target="_blank">
+            <input type="button" class="button" value="Java Doc"/>
```
scottdraves: instead of putting a button in the settings tab, add an item to the Help menu, like https://github.com/twosigma/beakerx/blob/ffd6f5cdf764814ce1bba815cf407b2dfcce261b/kernel/base/src/main/java/com/twosigma/beakerx/KernelInfoHandler.java#L73

scottdraves:
works great thx.  is there a way to filter classes from the javadoc though?
we really only want to show the runtime, not the kernel internals.

beakerx/beakerx/static/html/settings_tab.html
rbidas: Done

Issue #6459, opened 2017-12-05
remove internal classes from javadoc


scottdraves:
only runtime classes should be presented to the user.
anything that is internal to the kernel should be excluded.
basically, if it appears in any tutorial then it should be included.

PR #6463, opened 2017-12-06
#6459 remove internal classes from javadoc

rbidas:


scottdraves:
cool that's a huge improvement.
can you also add: com.twosigma.beakerx.mimetype, jupyter.Displayers, com.twosigma.beakerx.jvm.object

rbidas:
I add mimetype and jvm.object. jupyter.Displayers is in external lib.

Issue #6412, opened 2017-11-29
LoD menu breaks in lab


scottdraves:
<img width="582" alt="screen shot 2017-11-28 at 10 49 11 pm" src="https://user-images.githubusercontent.com/963093/33357129-70e53d02-d48e-11e7-9450-00ff5e04eeea.png">

PR #6467, opened 2017-12-06
#6412 fix LoD menu breaks in lab

piorek:


js/notebook/src/shared/style/bootstrap.less
```diff
@@ -0,0 +1,3 @@
+@import '~bootstrap/less/variables.less';
```
scottdraves: all files need banner

js/notebook/package.json
```diff
@@ -30,6 +30,8 @@
     "fork-ts-checker-webpack-plugin": "^0.2.8",
     "html-loader": "^0.4.5",
     "json-loader": "^0.5.4",
+    "less": "^2.7.3",
```
scottdraves: do we really need less?  we already have scss, can we just stick with one?

Issue #6429, opened 2017-11-30
add jupyterlab and jupyterhub to docker image


scottdraves:
i think it's as simple as
```
conda install -c conda-forge jupyterhub jupyterlab
```

michalgce:
@mjurowicz-xsolve after added jupyterlab 
![image](https://user-images.githubusercontent.com/1582983/33482877-6e44f6dc-d69b-11e7-8005-0a30d8328d9e.png)

after execute random groovy cell

scottdraves:
`jupyter labextension install @jupyter-widgets/jupyterlab-manager`

michalgce:
![image](https://user-images.githubusercontent.com/1582983/33550128-1c415f52-d8ed-11e7-96c9-aaef125471b9.png)

![image](https://user-images.githubusercontent.com/1582983/33550117-118b8416-d8ed-11e7-9efd-3196d839ddce.png)

MariuszJurowicz:
`jupyter labextension install @jupyter-widgets/jupyterlab-manager`
`jupyter labextension install beakerx-jupyterlab`

PR #6472, opened 2017-12-07
 #6429 Added jupyterlab and jupyterhub.

michalgce:


docker/setup.sh
```diff
@@ -15,12 +15,14 @@
...
 source activate beakerx
-conda install -y -c conda-forge ipywidgets
+conda install -y -c conda-forge ipywidgets jupyterhub jupyterlab
 
 rm -r /home/beakerx/beakerx/js/node_modules
 rm -r /home/beakerx/beakerx/js/dist
 
 (cd beakerx; pip install -e . --verbose)
+jupyter labextension install @jupyter-widgets/jupyterlab-manager
+jupyter labextension install beakerx-jupyterlab
```
scottdraves: this isn't right, it would install from npm.
we should install the code we just built from the local repository.
see the README for now to do that, basically need a path to js/lab directory.

Issue #6410, opened 2017-11-29
javascript disabled in lab


scottdraves:
SQL tutorial:
<img width="326" alt="screen shot 2017-11-28 at 10 06 59 pm" src="https://user-images.githubusercontent.com/963093/33356118-bfa220a0-d488-11e7-8720-f7388d1ccb3f.png">

scottdraves:
https://github.com/jupyterlab/jupyterlab/issues/3118

PR #6482, opened 2017-12-08
#6410 implement %%javascript magic for rendering the `application/javascript` mime

MariuszJurowicz:
This has to be tested heavily.
We don't have the requirejs on the jupyterlab yet.

### DIFF
--- a/beakerx/jslab/src/javascriptRendererExtension.ts
+++ b/beakerx/jslab/src/javascriptRendererExtension.ts
@@ -1,0 +1,62 @@
+/*
+ *  Copyright 2017 TWO SIGMA OPEN SOURCE, LLC
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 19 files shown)


scottdraves:
this works great but i am not sure i understand the new build requirements.
i had to run "npm install && npm run build".
1) why is a new build step required, we already have a "yarn install" run to compile this, right?
2) can we stick with yarn instead of npm?
3) whatever it is, this stuff has to be included into our regular build scripts before i can merge.

jmsdnns:
1. `npm run build` is an alias for `npm build`, and it does nothing unless you specify what "build" does in your package.json file, which https://github.com/twosigma/beakerx/pull/6482/commits/bfe75da56f600933749af2547ac9573460abd3b3 now does.

2. `yarn build` and `npm build` are _supposed_ to be equivalent.

MariuszJurowicz:
The separate build for the Lab extension is just the transpilation of the TypeScript sources to the js output. It can be done via npm or yarn but as this is published as the separate package i don't want to add an extra env requirements on the build process (just node and npm). The current approach lets developers change the sources and build the extension for themselves without depending on the beakerx notebook extension. But I think there are no obstacles to the running this build together with the beakerx notebook extension build.

MariuszJurowicz:
@scottdraves i've added the `prepublish` scripts so that you cand do just the `npm publish` and the npm will run the `yarn build:prod` for you before publishing to npm.
Also the build for lab is executed after the notebook build.

Issue #6478, opened 2017-12-08
heap size must be positive


scottdraves:
also show an error if the heap size is zero or negative, in addition to numeric.

<img width="287" alt="screen shot 2017-12-07 at 9 23 56 pm" src="https://user-images.githubusercontent.com/963093/33748205-2627422a-db95-11e7-85ea-6edd9f75a5e0.png">

piorek:
can be closed

Issue #6371, opened 2017-11-22
get rid of the OK button


scottdraves:
of our tab http://localhost:8888/tree#beakerx

sync continuously.

scottdraves:
when the values are written to the server, please show a yellow spinner that turns into a green check.

piorek:
done

PR #6487, opened 2017-12-11
#6371 get rid of the OK button

piorek:
Additionally because of code refactor it contains fix for:
- #6478 heap size must be positive


beakerx/beakerx/static/tree.js
```diff
@@ -264,107 +276,125 @@ define(function (require) {
...
+  }
+  function _setupFormEvents() {
+    $("#add_property_jvm_sett").click(_onJvmPropertyAddClickedHandler);
+    $("#add_option_jvm_sett").click(_onJvmOptionAddClickedHandler);
+    $('#heap_GB').keyup(InputChanged);
+  }
+  function _setupContinuousSync() {
+    var lastSyncedTs = null;
+    function synchronizeTick() {
+      setTimeout(function() {
```
scottdraves: wouldn't it be better to use _.debounce instead of implementing the same idea ourselves?

piorek: @scottdraves sorry i missed this comment

beakerx/beakerx/static/html/settings_tab.html
```diff
@@ -38,7 +46,11 @@
     <div class="beakerx_container">
         <form id="beakerx_jvm_settings_form" class="form-horizontal">
             <fieldset id="default_options">
-                <legend>JVM Options:</legend>
+                <legend>JVM Options: <span id="sync_indicator">
```
scottdraves: please move the sync indicator to the bottom, where the "Result" string is, and where the errors are shown.

scottdraves:
thanks that's better.

1) changes are not synced in all cases.  for example if i add a property, and then start typing into the property name, i can see the result changing at the bottom, but sync is not run.

2) when changes are synced, the focus is lost.  so if i am typing a property value, and i pause, then its synced (good) but i can't continue typing.

3) also, when in the cursor is in the property name field, and i press TAB, focus should move to the value for that property.

piorek:
Moved sync indicator

ad 2. and 3. 
fixed

ad 1. 
i had to disable syncing in that case until there is at least one character in value
because response from backend for payload with empty properties is not containing them

for egzample sending `{ a1: '1', b1: ''}` gives  `{ a1: 1 }`

scottdraves:
thanks.  after https://github.com/twosigma/beakerx/issues/6511 is fixed then we should be able to remove "rebuilding" the form, so focus is not lost.

scottdraves:
OK please bring up to date with master and continue.

piorek:
Done

scottdraves:
Great, this is very close.

the remaining problem: if the heap size field is bad, then it should not be sent to server.
right now i can enter "abc" into the heap size, and i  see the error reported in the UI.
but the value is still stored in ~/.jupyter/beakerx.json and will crash the server next time i create a notebook:
```
[E 10:55:52.015 NotebookApp] Failed to load kernel spec: 'clojure'
      File "/Users/spot/anaconda/envs/beakerx/lib/python3.6/site-packages/notebook/services/kernelspecs/handlers.py", line 57, in get
        d = kernelspec_model(self, kernel_name)
      File "/Users/spot/anaconda/envs/beakerx/lib/python3.6/site-packages/notebook/services/kernelspecs/handlers.py", line 22, in kernelspec_model
        spec = ksm.get_kernel_spec(name)
      File "/Users/spot/anaconda/envs/beakerx/lib/python3.6/site-packages/jupyter_client/kernelspec.py", line 205, in get_kernel_spec
        return self._get_kernel_spec_by_name(kernel_name, resource_dir)
      File "/Users/spot/anaconda/envs/beakerx/lib/python3.6/site-packages/jupyter_client/kernelspec.py", line 192, in _get_kernel_spec_by_name
        return self.kernel_spec_class.from_resource_dir(resource_dir)
      File "/Users/spot/anaconda/envs/beakerx/lib/python3.6/site-packages/jupyter_client/kernelspec.py", line 41, in from_resource_dir
	return cls(resource_dir=resource_dir, **kernel_dict)
      File "/Users/spot/src/beakerx/beakerx/beakerx/kernel_spec.py", line 23, in __init__
        args = EnvironmentSettings.read_beakerx_env_settings()
      File "/Users/spot/src/beakerx/beakerx/beakerx/environment.py", line 95, in read_beakerx_env_settings
	val = float(jvm_settings['heap_GB'])
    ValueError: could not convert string to float: 'abc'
```

piorek:
Should be OK now

beakerx/beakerx/static/tree.js
```diff
@@ -221,6 +214,22 @@ define(function(require) {
       }).val(value).keyup(InputChanged);
     },
 
+    normaliseHeapSize: function(val) {
+      if (val === '') {
+        return '';
+      }
+
+      var parsedVal = parseFloat(val);
```
scottdraves, line 222: this is broken, please do not change how valid numbers are identified, it's actually quite tricky.
`parseFloat("5a") = 5`

scottdraves:
also if i type a bad heap size, i can see the "sync" seems to run, it turns yellow and then goes green.
it shows a red error and the value in the config does not change, but why did the sync animation run?

beakerx/beakerx/static/tree.js
piorek: I did not change the way it was identified

see:
https://github.com/twosigma/beakerx/blob/27dab19f5c88a5e03cbb479a6da80a3bba020b73/beakerx/beakerx/static/tree.js#L54-L64
https://github.com/twosigma/beakerx/blob/27dab19f5c88a5e03cbb479a6da80a3bba020b73/beakerx/beakerx/static/tree.js#L98-L101

All i did with it was a little extraction refactor, because i needed that logic in two different places.

I added additional check 
- isFinite(val) will catch "5a"

piorek:
"sync" issue you mentioned is fixed now
  
Should there be some logic to prevent sending valid large numbers like "2e64"?

scottdraves:
i merged this by mistake and then reverted it.
there must have been an incorrect merge with master because now if i run this branch and enter "5" for the heap size i get a prefs file like this:
```
{
    "beakerx": {
        "heap_GB": "",
        "jvm_options": {
            "heap_GB": "5",
            "other": [],
            "properties": []
        },
        "version": 2
    }
}
```
it should be like this:
```
{
    "beakerx": {
        "jvm_options": {
            "heap_GB": 5,
            "other": [],
            "properties": []
        },
        "version": 2
    }
}
```

please fix this and make a new PR.

> Should there be some logic to prevent sending valid large numbers like "2e64"?

no

piorek:
Fixed and created new PR https://github.com/twosigma/beakerx/pull/6614

Issue #6209, opened 2017-10-27
add magic aliases 


scottdraves:
HTML should be an alias for html
js should be an alias for javascript

PR #6489, opened 2017-12-12
#6209 added aliases for javascript and html magic command

lmitusinski:
Default matchCommand logic has been separated to MagicCommandFunctionality interface as default methods.

 If magic command require to use multiple aliases it can override getMagicCommandAliases() or if it require custom command matching it can override matchCommand()


scottdraves:
This is good I like how `matchCommand` is factored out.
However, `%lsmagic` needs to report the new aliases.

scottdraves:
another possible approach: don't create the notion of "aliases" in the code at all.
just define two magics "js" and "javascript" that both happen to call the same implementation?

lmitusinski:
According to your comment i have created aliases as new magic commands by extending original magic commands and overwriting their names.

kernel/base/src/test/java/com/twosigma/beakerx/KernelTest.java
```diff
@@ -36,23 +36,7 @@
...
-import com.twosigma.beakerx.kernel.magic.command.functionality.HtmlMagicCommand;
-import com.twosigma.beakerx.kernel.magic.command.functionality.JavaScriptMagicCommand;
-import com.twosigma.beakerx.kernel.magic.command.functionality.LoadMagicMagicCommand;
-import com.twosigma.beakerx.kernel.magic.command.functionality.LsMagicCommand;
-import com.twosigma.beakerx.kernel.magic.command.functionality.TimeCellModeMagicCommand;
-import com.twosigma.beakerx.kernel.magic.command.functionality.TimeItCellModeMagicCommand;
-import com.twosigma.beakerx.kernel.magic.command.functionality.TimeItLineModeMagicCommand;
-import com.twosigma.beakerx.kernel.magic.command.functionality.TimeLineModeMagicCommand;
-import com.twosigma.beakerx.kernel.magic.command.functionality.UnImportMagicCommand;
+import com.twosigma.beakerx.kernel.magic.command.functionality.*;
```
scottdraves: wildcard is not allowed by our java style guide.

scottdraves: https://google.github.io/styleguide/javaguide.html#s3.3-import-statements

PR #6491, opened 2017-12-12
6340: add maven cache lock file

jaroslawmalekcodete:


scottdraves:
Let's think about the right way to do the locking: https://github.com/twosigma/beakerx/issues/6497

Issue #6477, opened 2017-12-07
groovy magic for python


scottdraves:
1) it should handle errors.  result should be the same as if you had an error in a regular groovy kernel notebook.
![screen shot 2017-12-07 at 2 00 37 pm](https://user-images.githubusercontent.com/963093/33733239-2920e622-db57-11e7-8e59-2f301d60420b.png)

2) it should handle widgets.
![screen shot 2017-12-07 at 2 02 59 pm](https://user-images.githubusercontent.com/963093/33733293-5bd819b4-db57-11e7-9509-9837ab59f8c6.png)

3) it should handle mime types.
![screen shot 2017-12-07 at 2 21 42 pm](https://user-images.githubusercontent.com/963093/33734118-ff0c9da6-db59-11e7-8e2d-12a8f226ddbe.png)

4) it should handle interrupt.
```
%%groovy
sleep(100*1000)
```
interrupt that.  then evaluate another %%groovy cell.

5) it should handle output.
![screen shot 2017-12-07 at 2 24 24 pm](https://user-images.githubusercontent.com/963093/33734462-6a078f4e-db5a-11e7-9665-b8e92673c052.png)

scottdraves:
dup of https://github.com/twosigma/beakerx/issues/5442

PR #6492, opened 2017-12-12
#6477 groovy magic for python

rbidas:


scottdraves:
that is a huge improvement, thank you!

this example just hangs for me though.  comms need to be 2-way?
```
%%groovy
beakerx.getCodeCells('tagme')
```

also good for testing: easyform and plot actions.

PR #6498, opened 2017-12-13
#6422 implement 'beaker.getcodecells' feature in Lab

MariuszJurowicz:


scottdraves:
I tried this with the doc/groovy/EasyForm.ipynb tutorial (because i don't see the UI for editing tags), and it worked, on the 1st try, but then i ran the easyform cell and ran the form, and then i tried to call getCodeCells() again and it just hung.

MariuszJurowicz:
@scottdraves 
The issue here is that it behaves exactly the same as in standard notebook.
So it's not the problem of the Lab implementation but maybe of something in lower level.

PR #6500, opened 2017-12-13
jarek/6340: move maven cache to conda evn

jaroslawmalekcodete:


kernel/base/src/main/java/com/twosigma/beakerx/kernel/CondaEnvCacheFolderFactory.java
```diff
@@ -0,0 +1,77 @@
...
+
+
+  public interface CondaPrefix {
+    String get();
+  }
+
+  public static class CondaPrefixSystem implements CondaPrefix {
+    @Override
+    public String get() {
+      return System.getenv(CONDA_PREFIX);
```
scottdraves: we need a way to get this even when not running conda. @jmsdnns

jmsdnns:
A Python programmer is very likely to use either Conda or Virtualenv and, from my experience, not likely to use the OS install of Python. So, we should check for Conda or Virtualenv and fall back to the original mechanism if we can't find either.

Like Conda, Virtualenv's function mostly by adjusting environment variables, so the mechanism in `CondaEnvCacheFolderFactory.java` would work. Might be worthwhile to rename this class `EnvCacheFolderFactory.java` and make it a general class.

For Virtualenv's, look for `$VIRTUAL_ENV`. Notice that this path goes to the root of the environment just like `$CONDA_PREFIX`.

```shell
$ workon sharknado
(sharknado) 
$ env | grep VIRTUAL_ENV
VIRTUAL_ENV=/home/jmsdnns/.virtualenvs/sharknado
...
(sharknado) 
$ deactivate
$ env | grep VIRTUAL_ENV
$
```

Here's the value for `$CONDA_PREFIX`, so we can see both at the same time.

```shell
$ source activate sharknado
(sharknado)
$ env | grep CONDA_PREFIX
CONDA_PREFIX=/home/jmsdnns/.conda/envs/sharknado
...
(sharknado)
$ source deactivate sharknado
$ env | grep CONDA_PREFIX
$
```

kernel/base/src/main/java/com/twosigma/beakerx/kernel/EnvCacheFolderFactory.java
```diff
@@ -0,0 +1,100 @@
...
+import static java.util.Arrays.asList;
+
+public class EnvCacheFolderFactory implements CacheFolderFactory {
+
+  public static final String PLEASE_SWITCH_TO_CONDA_ENV_OR_VIRTUAL_ENV = "Please switch to conda env or virtual env.";
+
+  public static final CondaPrefixSystem CONDA_PREFIX_SYSTEM = new CondaPrefixSystem();
+  public static final VirtualEnvPrefixSystem VIRTUAL_ENV_PREFIX_SYSTEM = new VirtualEnvPrefixSystem();
+
+  // "$CONDA_PREFIX/share/beakerx/maven/cache"
```
scottdraves: remove leftover

kernel/base/src/main/java/com/twosigma/beakerx/kernel/EnvCacheFolderFactory.java
```diff
@@ -0,0 +1,100 @@
...
+
+  private Path cache = null;
+  private List<EnvPrefix> condaPrefix;
```
scottdraves: just call this "prefix".  a virtualenv is not a conda env.

kernel/base/src/main/java/com/twosigma/beakerx/kernel/EnvCacheFolderFactory.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  public Path getCache() {
+    if (cache == null) {
+      cache = getOrCreateFile(getCondaPrefix() + "/share/beakerx").toPath();
+    }
+    return cache;
+  }
+
+  private String getCondaPrefix() {
```
scottdraves: also not well named but the important thing is, there might be no prefix at all, then should revert to current java.io.tempdir solution.

kernel/base/src/main/java/com/twosigma/beakerx/kernel/EnvCacheFolderFactory.java
```diff
@@ -0,0 +1,100 @@
...
+    String envPrefix = null;
+    for (EnvPrefix cp : condaPrefix) {
+      String prefixResult = cp.get();
+      if (prefixResult != null && !prefixResult.isEmpty()) {
+        envPrefix = prefixResult;
+        break;
+      }
+    }
+    if (envPrefix == null || envPrefix.isEmpty()) {
+      throw new RuntimeException(PLEASE_SWITCH_TO_CONDA_ENV_OR_VIRTUAL_ENV);
```
scottdraves: if neither is defined then use current tmp directory.

Issue #6499, opened 2017-12-13
Implement the comment-out cmd


MariuszJurowicz:
The comment-out cmd is not working in Lab.

PR #6515, opened 2017-12-15
#6499 add line comment feature to groovy cm mode

MariuszJurowicz:


scottdraves:
this isn't working for me.... cmd-/ does nothing.

MariuszJurowicz:
@scottdraves 
I have tested this in the fresh conda env and it did work for me (ctrl-/).
Maybe someone else can test.

scottdraves:
i tried it again with both cmd and ctrl, rebuilding in a fresh env.  same result.
mariusz can you write more exactly what you are doing to verify this works?
is there any other way to check if the function is active, like would it show up in the cmd palette?
@piorek can you please test this branch and see if it works for you?

MariuszJurowicz:
@scottdraves 
I've found a reason why you couldn't get this working properly: 
after building the lab extension from master branch, the `plugin.js` file was created in the `dist` dir. In `mariusz/6499` branch the `plugin.js` file was replaced by the `plugin` dir. But the lab was taking the old `plugin.js` file instead of `plugin/index.js` file as the `dist` folder wasn't cleaned up.
Now the build command for the lab extension cleans the `dist` dir at first and then builds extension from sources into `dist`.
Plz try to test this branch one more time.
You should have the `ctrl-/` working for groovy now.

Issue #6512, opened 2017-12-15
add flint demo


scottdraves:
https://github.com/twosigma/flint

this is another spark demo this time using flint in spark.

PR #6522, opened 2017-12-18
#6512 flint demo

lmitusinski:
Flint demo for scala
It requires building flint project and adding jar to classpath


scottdraves:
Thanks, this looks cool.  The zip file seems wrong though, it's 118 bytes and it's not actually zipped.
1) no need to zip the file.
2) give it a descriptive name, eg "flint-demo.csv".

i will look into finding a published version of built version.

Issue #6511, opened 2017-12-14
improve beakerx.json format


scottdraves:
currently we use javascript properties to represent java properties:
```
{
    "beakerx": {
        "jvm_options": {
            "other": [],
            "properties": {
                "x": "2"
            }
        }
    }
}
```
the problem with this is, order of properties is not guaranteed, duplicates are not allowed.
instead we should have an array of name/value objects, like this:
```
{
    "beakerx": {
        "jvm_options": {
            "other": [],
            "properties": [
                {"name: "x", "value": "2"}
            ]
        }
    }
}
```

we should support reading the old format so that if someone has a config file with the old format, they are transparently upgraded.

also, make sure that empty values are supported.  see https://github.com/twosigma/beakerx/pull/6487
which says
> sending { a1: '1', b1: ''} gives { a1: 1 }

PR #6537, opened 2017-12-19
#6511 improve beakerx.json format

rbidas:


beakerx/beakerx/environment.py
```diff
@@ -20,6 +20,7 @@
 default_config = """
 {
     "beakerx": {
+        "version": 1.1
```
scottdraves: here the version is floating point but below it is a string.
IMO integer is best.

Issue #6475, opened 2017-12-07
plot resize thumb missing in lab


scottdraves:
![screen shot 2017-12-07 at 10 46 39 am](https://user-images.githubusercontent.com/963093/33723923-f3678cea-db3b-11e7-9b4d-98ad844f2907.png)

scottdraves:
should be like:
![screen shot 2017-12-07 at 10 50 21 am](https://user-images.githubusercontent.com/963093/33724611-e6399372-db3d-11e7-9cb6-67b3c60c5cd5.png)

PR #6539, opened 2017-12-19
#6475 plot resize thumb missing in lab

piorek:


piorek:
@scottdraves @mjurowicz-xsolve there were couple issues 

1. when building extension from js/notebook to js/lab/lib 
image files are collected fom styles, but then were not included when publishing extension
fixed by modifing package.json files section

2. url-s generated when building extension are dont contain /lab/static prefix
only way i was able to fix this was by overriding css classes in lab

But that is not a good idea because i had to overwrite css by hand.

I experimented with webpacks `output.publicPath` and with `__webpack_public_path__` but was not able to get it done this way.

MariuszJurowicz:
I think we could get rid of the extra assets and use the `base64-inline-loader` to emit base64 encoded image data inline. It will produce slightly bigger js files.

piorek:
Thanks will try that

piorek:
Done

Issue #6550, opened 2017-12-20
multiple classpath magics output broken


scottdraves:
1) the outputs should be printed as soon as each magic finishes.  currently, i see the progress as some of them run, but the final output does not appear until they are all finished.
2) the outputs are appearing all on the same line.  result of each magic should appear on its own line.
![screen shot 2017-12-20 at 11 30 12 am](https://user-images.githubusercontent.com/963093/34217335-371abc68-e579-11e7-97ba-d443167108e5.png)

PR #6558, opened 2017-12-21
#6550 changed dsiplay of magic commands output

lmitusinski:


scottdraves:
The output looks good but there's a major problem which is the evaluation is now considered finished right away even though it is actually still running.

On master if you run that flint cell, the cell status stays "*" until it's done.
on this branch, the cell status becomes a number right away, and the various outputs appear afterwards.

lmitusinski:
Fixed, now output is displayed after each command but status stay unchanged until all magic commands in cell are completed.

Issue #6553, opened 2017-12-20
easyform radio buttons broken


scottdraves:
open doc/groovy/EasyForm.ipynb, run the cell "All Kinds of Fields", then click on a radio button.
the kernel dies with this stack trace:

```
Exception in thread "Thread-1" java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.String
        at com.twosigma.beakerx.widgets.selections.SingleSelectionWidget.setValue(SingleSelectionWidget.java:46)
        at com.twosigma.beakerx.easyform.EasyFormComponent.lambda$setupNewValue$0(EasyFormComponent.java:172)
        at java.util.Optional.ifPresent(Optional.java:159)
        at com.twosigma.beakerx.easyform.EasyFormComponent.setupNewValue(EasyFormComponent.java:172)
        at com.twosigma.beakerx.kernel.comm.Comm.handleMsg(Comm.java:224)
        at com.twosigma.beakerx.kernel.handler.CommMsgHandler.handleMsg(CommMsgHandler.java:50)
        at com.twosigma.beakerx.kernel.handler.CommMsgHandler.lambda$handle$0(CommMsgHandler.java:41)
        at com.twosigma.beakerx.handler.KernelHandlerWrapper.wrapBusyIdle(KernelHandlerWrapper.java:25)
        at com.twosigma.beakerx.kernel.handler.CommMsgHandler.handle(CommMsgHandler.java:40)
        at com.twosigma.beakerx.kernel.handler.CommMsgHandler.handle(CommMsgHandler.java:31)
        at com.twosigma.beakerx.socket.KernelSocketsZMQ.handleShell(KernelSocketsZMQ.java:179)
        at com.twosigma.beakerx.socket.KernelSocketsZMQ.run(KernelSocketsZMQ.java:158)
```

PR #6559, opened 2017-12-21
#6553 fixed radio button on easyfrom

lmitusinski:
Radio buttons values were indentified incorrectly by index


scottdraves:
unit tests fail!

Issue #6603, opened 2018-01-05
fix conversion of old json prefs file with missing heap size


scottdraves:
![screen shot 2018-01-05 at 9 45 53 am](https://user-images.githubusercontent.com/963093/34613845-466076a0-f1fd-11e7-8b83-bbc735ed3688.png)

followup to https://github.com/twosigma/beakerx/pull/6580

PR #6604, opened 2018-01-05
#6603 heap_gb field not required in old json format

lmitusinski:


beakerx/beakerx/environment.py
```diff
@@ -74,7 +74,7 @@ def _convert_to_version_2(beakerx_settings):
             }
             new_prop.append(prop)
         settings['properties'] = new_prop
-        settings['heap_GB'] = float(settings['heap_GB']) if settings['heap_GB'] else None
+        settings['heap_GB'] = float(settings['heap_GB']) if settings.get('heap_GB') else None
```
scottdraves: wouldn't it be better to say:
```
if settings.get('heap_GB'):
    settings['heap_GB'] = float(settings['heap_GB'])
```

Issue #6575, opened 2017-12-30
Zoom the x-axis and scale the y-axis


benmccann:
Zooming in zooms equally on the x and y axes. It would be nice if it zoomed the x-axis and then rescaled the y-axis according to the values showing on the chart. This is how most charts I've seen such as Yahoo Finance or Google Finance work

I have a chart with many small values on the left and then some large values on the right. I want to zoom in on the left to see those values. The problem is that they still remain tiny after zooming because the y-axis remains scaled for the largest values in the chart which are no longer showing

scottdraves:
interesting idea, will have to experiment with that.  for now, you can zoom differently in x and y by moving the mouse over an axis and then zooming to apply it just to that axis.  but for time plots at least doing it automatically seems like a win, as long as it is not too jarring.

PR #6608, opened 2018-01-06
#6575 Autoscale Y on mouse zoom

markhalonen:
This is my first contribution to this project.

If the mouse zoom is in the chart area (not in the axis) and all the plot types are "point", then it will auto zoom the Y axis to fit the data in the x range. 

Right now, there is no margin on the fit. I was thinking of using `getDefaultFocus()` from `plotUtils.js`, but it requires the `model` as the argument. So I think I'd have to create a copy of the model, delete points that are not included, then pass to `getDefaultFocus()` to get the margins. Perhaps there's a better way?


js/notebook/src/plot/plotScope.js
```diff
@@ -1714,7 +1714,24 @@ define([
...
+                d => d.elements.filter(el => el.x >= focus.xl && el.x <= focus.xr).map(el => el.y)
+              )
+            );
+            var minYValue = Math.min(...AllYValsInRange);
+            var maxYValue = Math.max(...AllYValsInRange);
+            focus.yl = minYValue;// - self.stdmodel.margin.bottom;
+            focus.yr = maxYValue;// + self.stdmodel.margin.top;
+            focus.yspan = focus.yr - focus.yl;
+        }
+        else if (my <= plotUtils.safeHeight(self.jqsvg) - self.layout.bottomLayoutMargin) {
```
benmccann: I think this `else if` should be on the previous line with the closing brace for consistency

js/notebook/src/plot/plotScope.js
```diff
@@ -1714,7 +1714,24 @@ define([
       }else{
         // scale only
         var level = self.zoomLevel;
-        if (my <= plotUtils.safeHeight(self.jqsvg) - self.layout.bottomLayoutMargin) {
+        var isAllPointPlots = this.stdmodel.data.length &&
+          this.stdmodel.data.every(plot => plot.type == "point");
+        if(my <= plotUtils.safeHeight(self.jqsvg) - self.layout.bottomLayoutMargin
```
benmccann: it looks like the rest of the file puts a space after `if` before the `(`

benmccann:
This is awesome Mark! My only big piece of feedback is that I think this should apply to other chart types as well (line, bar, area, etc.)

I'm very excited for this fix for https://github.com/twosigma/beakerx/issues/6575. Thanks

markhalonen:
I was able to use the existing `getRange()` method of the chart types. So this should support autozoom for all plots with `getRange()` (which seems to be all of them). If `getRange()` is not supported, it falls back to the default zoom behavior.

js/notebook/src/plot/plotScope.js
```diff
@@ -1714,7 +1714,29 @@ define([
...
         // scale only
         var level = self.zoomLevel;
-        if (my <= plotUtils.safeHeight(self.jqsvg) - self.layout.bottomLayoutMargin) {
+        var autoZoomSuccess = false;
+        if (my <= plotUtils.safeHeight(self.jqsvg) - self.layout.bottomLayoutMargin
+          && mx >= self.layout.leftLayoutMargin) {
+            // Zooming in the middle of the chart, autoscale Y
+            var data = this.stdmodel.data;
+            if (data.map(d => d.getRange ? true : false).every(b => b))
+            {
```
benmccann: should probably put this brace on the previous line for consistency

benmccann:
Awesome! This will be amazing to use!

scottdraves:
This is great work, thank you Mark.  I think this is useful but we want to keep the BeakerX experience consistent.  Can you make this behavior conditional on the Control key?

js/notebook/src/plot/plotScope.js
```diff
@@ -1714,7 +1714,28 @@ define([
...
+              var ranges = data.map(d =>
+                d.getRange(d.elements.filter(el =>
+                  el.x >= focus.xl && el.x <= focus.xr
+                ))
+              );
+              var minYValue = Math.min(...ranges.map(r => r.yl).filter(y => !isNaN(y) && isFinite(y)));
+              var maxYValue = Math.max(...ranges.map(r => r.yr).filter(y => !isNaN(y) && isFinite(y)));
+              if(!isNaN(minYValue) && isFinite(minYValue) && !isNaN(maxYValue) && isFinite(maxYValue)) {
+                autoZoomSuccess = true;
+                focus.yl = minYValue;
```
benmccann, line 1733: I think that we should add a check so that if the current base of the chart is 0 then we shouldn't change it

benmccann:
@scottdraves I often use a laptop, which doesn't have a mouse wheel. The current zooming is pretty tricky to use in that situation, so I'm a little nervous about adding another key to the mix. Do you have any thoughts on zooming on a laptop? Potentially we could add a plus and minus button that appear when you hover over the chart. Or I think that making this the default behavior without adding the Control key makes sense because this PR makes the behavior more consistent with most charts I've seen that have panning and zooming (e.g. yahoo finance, google finance, trading view, etc.). It seems like it wouldn't be terrible to change now since beakerx is still in beta whereas changing things like that will only get harder in the future

scottdraves:
What OS/laptop needs a modifier for zoom?  It would be fine on the mac laptops i often use (though on 2nd thought I would go for option not control as the modifier).  Some plots autoscale like that but many do not, and those are APIs like ours mostly do not autoscale (bokeh, plotly, bqplot).  The chartjs zoom plugin (yours) was the only one that did it like this.  So I like this feature but I prefer to make it an option.  Our current users are happy with how zooming works.  I think they would be even happier with the option to autoscale.

benmccann:
I use a trackpoint / eraser head type pointing stick. The zooming is designed to work much better with touchpad like on a mac. I wonder if there's a design that we can come up with that would be easy on both platforms

Btw, I just wanted to mention that I picked Chart.js because of the way it works and not that Chart.js works the way it does because I wrote it's code ;-) I only became involved in Chart.js recently and didn't develop any of it's original zoom functionality

jmsdnns:
Could a setting resolve the difference of opinion? Maybe it defaults to requiring a modifier key but something could be in beakerx.json that turns that off?

benmccann:
That would work well enough for my case

benmccann:
Bokeh offers this functionality as a chart option instead of using a modifier key. You can set the wheel zoom mode to either `wheel_zoom`,  `xwheel_zoom`, or `ywheel_zoom`.  bqplot's documentation is not as clear, but it also seems like it is controlled similarly

plot.ly will autoscale an axis when you double click on it

js/notebook/src/plot/plotScope.js
markhalonen: Can you clarify what you mean by "current base"?

scottdraves:
> Bokeh offers this functionality as a chart option instead of using a modifier key. 

this is a fine idea

js/notebook/src/plot/plotScope.js
benmccann: The minimum value on the y-axis

markhalonen: Ok, I'm not sure I understand. If the miniumum y-axis value (in the zoomed area) is 0, shouldn't it autoscale and set the lower bound to 0? I'm not seeing why it's a special case.

markhalonen:
Is a "chart option" an argument into a specific chart? Or is it something that goes into `beakerx.json`?

js/notebook/src/plot/plotScope.js
benmccann: I was thinking that the user might want to be able to for the minimum value on the y-axis to be 0 even if the smallest point is bigger than 0. And if they did so, then we wouldn't want to change the minimum value on the axis by zooming in that case. However, I'm not seeing that such an option exists right now, so we probably don't need to do anything after all. I think that the we add an option to manually set the minimum on the y-axis in the future then we should be sure to pay attention to that option here. But until we have such an option we probably don't need to do anything, so you can ignore my initial comment.

benmccann:
It's an argument you can you when creating a chart or option you set on the chart after creation. Not something in `beakerx.json`

markhalonen:
Ok, I've removed the alt key functionality and implemented a chart option named `autoZoom` instead.

benmccann:
Looks good to me! Only last suggestion would be to add a sentence about the new option to the docs in the zoom section: https://github.com/twosigma/beakerx/blob/master/doc/groovy/Charting.ipynb

scottdraves:
🐬 
thank you for doing the extra work and accommodating our perspective.

Issue #6546, opened 2017-12-20
Refactor the initializationCells extension feature


MariuszJurowicz:
Examine this feature in order to implement into Lab extension.

PR #6616, opened 2018-01-08
#6546 implement initialization cells on Lab

MariuszJurowicz:


scottdraves:
sorry i merged this before testing it back in classic nb.
it has a regression there, which is it prompts me about running the init cells and an untrusted notebook, even if there are no init cells.
just open the StartHere.ipynb to see the problem...

Issue #6586, opened 2018-01-04
publish gist errors out


scottdraves:
```
VM786:1 POST https://api.github.com/gists 403 (Forbidden)
(anonymous) @ VM786:1
send @ jquery.js:9566
ajax @ jquery.js:9173
do_publish @ GistPublish.js:64
mightThrow @ jquery.js:3583
process @ jquery.js:3651
setTimeout (async)
(anonymous) @ jquery.js:3689
fire @ jquery.js:3317
add @ jquery.js:3376
(anonymous) @ jquery.js:3709
Deferred @ jquery.js:3793
then @ jquery.js:3694
click @ GistPublish.js:75
i @ jquery.min.js:4
dispatch @ jquery.min.js:5
y.handle @ jquery.min.js:5
GistPublish.js:60 Forbidden
```

MariuszJurowicz:
@scottdraves 
I can't reproduce this bug. Maybe it was something on github?
It sounds like the authentication problem but we are publishing gists as anonymous so there is no authentication during publishing.

jmsdnns:
Reading this gave me an idea. It could be a pretty neat feature to let someone log in to github and post authenticated gists. Perhaps even some interaction with git repos hosted on github.

scottdraves:
Yup we should.  My guess is I am rate limited or something.

scottdraves:
In the meantime, we should at least present this error to the user in a reasonable way.
So if the the post fails, please show a popup with the error.

scottdraves:
https://github.com/twosigma/beakerx/issues/6594

PR #6628, opened 2018-01-10
#6586 show error dialog on gist publishing error

MariuszJurowicz:


scottdraves:
i disabled wifi and then published and got a blank error.  
screen shot 2018-01-10 at 9 20 09 am

MariuszJurowicz:
xhr does not fail with the error the console logs (`net:ERR_INTERNET_DISCONNECTED`);
The only way i can show some message is to check the `readyState` when xhr fails. If the state is `0` there is some kind of network/connection problem so i can show eg. `NETWORK_ERROR` msg.

Issue #6629, opened 2018-01-10
Implement gist publish feature for Lab


MariuszJurowicz:

scottdraves:
merged the anonymous version, now needs the dialog to support the token.

PR #6635, opened 2018-01-10
#6629 implement gist publish feature for Lab

MariuszJurowicz:


MariuszJurowicz:
@scottdraves 
The feature for saving the notebook state with the widgets state is not implemented in jupyterLab yet.
The publication will work but without saving the widgets state.
Is it OK?

scottdraves:
yes

Issue #6594, opened 2018-01-04
allow publishing non-anonymously


scottdraves:
the publish dialog, instead of just supporting anonymous gists, should support using your regular github identity via oauth.

MariuszJurowicz:
@scottdraves 

As i understood, to allow users for publishing gists into their accounts we need to register an app in github that will request for the `gists` oAuth scope.
To register the app we need to provide the:

- full URL to the application homepage,
- Authorization callback URL that can be changed by `redirect_uri` paramenter in the request.

The flow looks like we need to redirect to github page for authentication and the we are redirected back to our beakerx page with the `code` data that can be exchanged to `access_token`.
This token has to be used to sign the `/gists` POST call.

- who should register the app and provide such data,
- how should the configuration look like and where it should be kept,
- is it possible to redirect to dynamic url with different port like eg. in our local instances?
- is this solution secure enough like eg. what if someone's beakerx instance requests for different scopes and use it for some illegal actions? Should we even care about that?
  
We can also request in pop up for the user's Personal Access Token that is configured for `gists` scope and use it to sign up the creating gist request.

jmsdnns:
OAuth wouldn't make sense for us as we don't have a concept of user accounts in BeakerX.

Personal Access Tokens are the right approach here.

PR #6637, opened 2018-01-11
#6594 allow to publish personal gists

MariuszJurowicz:
Wait till https://github.com/twosigma/beakerx/pull/6628 is merged.


scottdraves:
cool, that works great!
however we need the access token to be persistent, stored in beakerx.json like the JVM options are.

MariuszJurowicz:
@scottdraves 
I'm not sure how the use case for storing the Personal Access Token should look like.
It can be useful if the single instance of beakerx is used by single user, but as we don't introduce user accounts it can be the security breach.

Maybe we can add the `remember` checkbox and store the token in session storage or local storage.

MariuszJurowicz:
UC: if user fills the token input, store it in beakerx.json and fill the input next time with the saved token. If user removes the token from input, clear the beakerx.json token that was saved before. If user fills the input with different token, overwrite the saved token with the new one.

jmsdnns:
You are right to be concerned about the personal access token being something that could be hijacked.

Do we do anything to help the user limit the scope of the personal access token? We only need the "create gists" scope, which would protect everything else in the user's github account.

You can see all the scopes here and see that "create gists" is one of the choices: https://github.com/settings/tokens/new

MariuszJurowicz:
@jmsdnns @scottdraves 
I've added the informatio about scopes to publish modal:

> Enter your Personal Access Token to publish gist to your github account.
> Make sure your Personal Access Token has the _gists_ scope _only_.
> You can read about scopes [here](https://developer.github.com/apps/building-oauth-apps/scopes-for-oauth-apps/)

js/notebook/src/extension/gistPublish/gistPublishModal.ts
```diff
@@ -0,0 +1,98 @@
...
+  static create(submitCallback, personalAccessToken = '') {
+    const modalContent = GistPublishModal.createModalContent();
+    const personalAccessTokenInput = modalContent.querySelector('input');
+
+    if (personalAccessTokenInput) {
+      personalAccessTokenInput.value = personalAccessToken;
+    }
+
+    dialog.modal({
+      title : 'Publish to a Github Gist',
```
scottdraves: should be capitalized "GitHub"

scottdraves:
screen shot 2018-01-19 at 9 13 01 am
screen shot 2018-01-19 at 9 13 17 am

scottdraves:
actually, i had forgotten to add the "gist" scope to the token.
so the error is not that informative but this is in fact the error github reports, so i don't see a good way to fix this.  the good news is, the error is reported nicely :)

Issue #6638, opened 2018-01-11
in python display of array of dicts crashes 


scottdraves:
```
TableDisplay([{'y1':4, 'm3':2, 'z2':1}, {'m3':4, 'z2':2}])
```
fails with
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-7-5e569ea9aeca> in <module>()
----> 1 TableDisplay([{'y1':4, 'm3':2, 'z2':1}, {'m3':4, 'z2':2}])

~/anaconda/envs/beakerx11/lib/python3.6/site-packages/beakerx/tabledisplay/tabledisplay.py in __init__(self, *args, **kwargs)
    184     def __init__(self, *args, **kwargs):
    185         super(TableDisplay, self).__init__(**kwargs)
--> 186         self.chart = Table(*args, **kwargs)
    187         self.model = self.chart.transform()
    188         self.on_msg(self.handle_msg)

~/anaconda/envs/beakerx11/lib/python3.6/site-packages/beakerx/tabledisplay/tabledisplay.py in __init__(self, *args, **kwargs)
     37             self.convert_from_dict(args)
     38         elif isinstance(args[0], list):
---> 39             self.convert_from_list(args, types_map)
     40 
     41         self.headersVertical = False

~/anaconda/envs/beakerx11/lib/python3.6/site-packages/beakerx/tabledisplay/tabledisplay.py in convert_from_list(self, args, types_map)
     87                 value_type = types_map.get(columnName)
     88 
---> 89                 row.append(self.convert_value(value, value_type))
     90             self.values.append(row)
     91 

~/anaconda/envs/beakerx11/lib/python3.6/site-packages/beakerx/tabledisplay/tabledisplay.py in convert_value(value, value_type)
    130                 converted_value = value
    131             else:
--> 132                 converted_value = value.item()
    133         elif value_type == "int64":
    134             converted_value = value.astype('str')

AttributeError: 'str' object has no attribute 'item'
```

PR #6650, opened 2018-01-12
#6638 fixed table display type conversion

lmitusinski:


beakerx/beakerx/tabledisplay/tabledisplay.py
```diff
@@ -121,7 +121,9 @@ def convert_from_pandas(self, args, types_map):
 
     @staticmethod
     def convert_value(value, value_type):
-        if value_type == "time":
+        if value == "":
```
jmsdnns, line 124: I'm curious if a falsy test would be helpful here. The current code checks specifically for value being an empty string, but if the check was `if value:` we would catch None, an empty list, and other falsy values too.

Not sure if that's desirable, but certainly worth mentioning.

scottdraves: empty list should appear as [].
none sounds right though, what does it do now?

jmsdnns: In the context of this function, it seems we return "string" for anything that doesn't match the specific checks. If we want to return "string" even when value is `None`, then the current code is all good. If we want to do something else when the value is `None`, we could do this at the at the end of the function.

Again, I'm not sure what the ideal behavior is, but I see a potentially unhandled case.

```
if value:
    return "string"
raise Exception("Unknown value type")
```

scottdraves: see also https://github.com/twosigma/beakerx/pull/6657 https://github.com/twosigma/beakerx/issues/6656

PR #6654, opened 2018-01-12
#6652 beakerx.json priviliges changed to private

lmitusinski:


scottdraves (commented):
Create it with the right permissions, don't fix it afterwards

beakerx/beakerx/environment.py
```diff
@@ -40,9 +40,8 @@ def __init__(self):
     @staticmethod
     def save_setting_to_file(content):
         pathlib.Path(paths.jupyter_config_dir()).mkdir(parents=True, exist_ok=True)
-        file = open(EnvironmentSettings.config_path, 'w+')
-        file.write(json.dumps(json.loads(content), indent=4, sort_keys=True))
-        file.close()
+        with fdopen(osopen(EnvironmentSettings.config_path, O_WRONLY | O_CREAT | O_TRUNC, 0o600), 'w+') as file:
```
scottdraves: i think O_WRONLY should be O_RDWR

scottdraves:
thank you, works well, but i think above would be an improvement.

Issue #6647, opened 2018-01-12
Right click works wrong for windows


EfimovVladimir:
First right click on the plot do nothing
Second right click on the plot open the menu
(reproduced only for Windows)

![image](https://user-images.githubusercontent.com/17744188/34867027-a7a1d4bc-f78f-11e7-9de8-26c93ec1b2a6.png)

PR #6669, opened 2018-01-16
#6647 fix right click on plot in ms edge

MariuszJurowicz:
This bug was caused by the different events dispatch order between MS Edge and other browsers.

### DIFF
--- a/js/notebook/src/plot/plotScope.js
+++ b/js/notebook/src/plot/plotScope.js
@@ -103,6 +103,7 @@ define([
     this.legendResetPosition = false;
     this.doNotLoadState = false;
     this.saveAsMenuContainer = null;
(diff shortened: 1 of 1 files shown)


scottdraves:
👍  tested on chrome/mac

EfimovVladimir:
failed on windows/chrome

MariuszJurowicz:
@EfimovVladimir 
Can you be more specific on what's not working?
does it work on edge for you?

EfimovVladimir:
for first right click nothing happens, but for second right click opens menu.
i.e. every second right click opens the menu.
also the double right click opens the menu.

EfimovVladimir:
after rebuilding js for our app the right click started to work fine, so
:+1: tested on chrome/windows

Issue #6611, opened 2018-01-08
easyform radio buttons bad default value


scottdraves:
run the "all kinds of fields" example, and click "Go!" right away without clicking on anything.
in the form, the 1st item is selected:
<img width="393" alt="screen shot 2018-01-08 at 12 06 59 am" src="https://user-images.githubusercontent.com/963093/34659572-1aa2ab0c-f408-11e7-8469-cb00443e3847.png">
however, the resulting values for the radio buttons are empty:
<img width="297" alt="screen shot 2018-01-08 at 12 04 20 am" src="https://user-images.githubusercontent.com/963093/34659533-cd197bb8-f407-11e7-85ef-0670b2afdcc0.png">

buttons should not be selected unless you actually click on them (or assign them a value).
expected behavior: unset should display as unselected.

PR #6670, opened 2018-01-16
#6611 easyform radio buttons bad default value

piorek:


piorek:
I had to introduce beakerx RadioButtons, and made corresponding  changest in java and python files, but now I think its purely backend issue.

in **pyton** notebook i can do:
```py
from beakerx import *
g = EasyForm("Field Types")
options = ["a", "b", "c", "d"]
g.addRadioButtons("Radio Buttons", options)
g["Radio Buttons"] = "a"
g
```
`"a"` option of radio is selected as expected
and then 
```py
g["Radio Buttons"] = None
```
no option of radio is selected

----

I've tried to do the same in **groovy**:
```java
g = new EasyForm("Field Types")
options = ["a", "b", "c", "d"]
g.addRadioButtons("Radio Buttons", options)
g["Radio Buttons"] = "a"
g
```
`"a"` option of radio is selected
but when I've tried to unselect radio options by 
``` java
g["Radio Buttons"] = null
```
nothing happens (i tried with `""` as well)

----

I believe that this is caused by java implementation of **SelectionWidget**.
`defaultIndex` is 0 and therefore it is not possible to select nothing

If that will be fixed overwriting `render()` method in js/notebook/src/easyForm/radioButtonsWidget.ts
probably won't be needed.

scottdraves:
Sorry I don't understand why we can't use the std radio button widget, can you explain please?

scottdraves:
I think you are right about groovy not being able to clear with null.  @lmitusinski can you please push a commit to fix that?

piorek:
After fix in java, std radio widget will be ok.

I had to introduce RadioButtonsView and RadioButtonsModel in `ts` to overwrite render method. When I did that I had to make corresponding changes on the `java` and `python` files.

While making those changes I realized that implementation differs and its not an issue to fix on front.

piorek:
The difference between python and java is the initial state of radio widget after creation:

**pyhon** 
index: None value: None
**java** 
index: 0, value: 1st option

piorek:
Cleaned PR, tested on python and groovy kernel

PR #6671, opened 2018-01-16
add cleaning of cell outputs for test notebooks

EfimovVladimir:


EfimovVladimir:
it should clear all cells output after tests done for notebook

Issue #6674, opened 2018-01-16
upgrade table UI from datatables to phosphor grid


scottdraves:
stage 1: replace the UI with https://github.com/phosphorjs/phosphor/issues/285
stage 2: make it streaming with https://github.com/apache/arrow/tree/master/js

scottdraves:
what about html format?
<img width="508" alt="screen shot 2018-02-27 at 10 44 19 pm" src="https://user-images.githubusercontent.com/963093/36769052-027b1f0c-1c10-11e8-8785-66d77a8b5191.png">

MariuszJurowicz:
@scottdraves this is really tricky as we are rendering to canvas.
There are some tricks to render the HTML into canvas eg. (https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Drawing_DOM_objects_into_a_canvas)
but those rendered objects won't be interactive even though.

scottdraves:
OK.  rendering to a non-interactive image is still pretty useful, we should do that for "html" format.
we should have an easy way to make URLs clickable, maybe an "url" format.

PR #6680, opened 2018-01-18
#6674 use phosphor datagrid to display TableDisplay data

MariuszJurowicz:


scottdraves:
🐬 
screen shot 2018-01-18 at 10 35 56 pm

js/notebook/src/tableDisplay/dataGrid/dataGrid.css
```diff
@@ -0,0 +1,80 @@
+.p-DataGrid {
```
scottdraves, line 1: banner

Issue #6619, opened 2018-01-09
What's going on with set[yY]AutoRange?


jpallas:
While working on the Scala getters and setters, I stumbled across a peculiar situation.  `setYAutoRange` is defined on `XYChart`, while `setyAutoRange` is defined in both `AbstractChart` and `XYChart`.

The definition in `AbstractChart` looks reasonable:
```java
  public AbstractChart setyAutoRange(boolean yAutoRange) {
    this.yAxis.setAutoRange(yAutoRange);
    sendModelUpdate(ChartToJson.serializeAutoRange(this.yAxis.getAutoRange()));
    return this;
  }
```

The definition of `setyAutoRange` in `XYChart` just delegates to `setYAutoRange`:
```java
  public XYChart setyAutoRange(boolean yAutoRange) {
    return this.setYAutoRange(yAutoRange);
  }
```

The definition of `setYAutoRange` in `XYChart` looks insane:
```java
  public XYChart setYAutoRange(boolean yAutoRange) {
    setXAutoRange(yAutoRange);
    return this;
  }
```

What is going on here?  

The Java bean-style property naming conventions invite confusion about property names that start with a one-letter "word", so the fact that both variants exist isn't surprising.  Groovy follows the [formal rules](http://www.oracle.com/technetwork/java/javase/documentation/spec-136004.html) implemented by [`Introspector`](https://docs.oracle.com/javase/8/docs/api/java/beans/Introspector.html#decapitalize-java.lang.String-), so it thinks `[gs]etYAutoRange` methods define a property named `YAutoRange` and not one named `yAutoRange`.  Users, of course, expect the name `yAutoRange` because it certainly doesn't occur to them that the "YA" in "YAutoRange" looks like the start of an acronym.  😭 

Since `XYChart` extends `AbstractChart`, my best guess is that the inconsistency in naming was noticed and something went horribly wrong in an attempt to fix it.  It seems like the fix should have been applied in `AbstractChart` but was instead applied in `XYChart` and just got messed up.

scottdraves:
Thanks yes this is broken!

scottdraves:
we should support both setYAutoRange and setyAutoRange, but not via insane code.

PR #6681, opened 2018-01-18
#6619 xy properties

jpallas:
This fixes the setyAutoRange bug and supplies (I think) all the missing upper/lower case versions of properties that start with X or Y.


jpallas:
Explicitly tagging #6619 because WTF github?

scottdraves:
awesome thank you, i just wrapped up 0.11.1 and will get to the PR backlog shortly.

Issue #6682, opened 2018-01-19
move away old prefs during install


scottdraves:
beakerx-install should move ~/.jupter/beakerx.json to ~/.jupter/beakerx.json.bak
because otherwise the file will be left world readable and github token might be exposed.

PR #6684, opened 2018-01-19
#6682 beaker config is moved to .bak by beakerx-install

lmitusinski:


scottdraves:
now that i think about this, it would be better to just chmod the file.
otherwise they will lose their configs every release.

lmitusinski:
beakerx-install changed to modify file privileges to private on instalation

beakerx/beakerx/install.py
```diff
@@ -127,6 +128,11 @@ def _install_magics():
     file.write("c.InteractiveShellApp.extensions = ['beakerx.groovy_magic']")
     file.close()
 
+def _set_conf_priviliges():
```
scottdraves: spelling error, should be "privileges".

beakerx/beakerx/install.py
```diff
@@ -127,6 +128,11 @@ def _install_magics():
...
+    config_path = os.path.join(paths.jupyter_config_dir(), 'beakerx.json')
+    if pathlib.Path(config_path).exists():
+        os.chmod(config_path, 0o600)
```
scottdraves: does this work on windows?

Issue #6691, opened 2018-01-22
Clarify docs for autoZoom


benmccann:
The docs say:

>If you'd like the y-axis to automatically scale to fit the data points, set the `autoZoom` parameter of the plot to `true`

I tried this with the code below:

```
sp500 = session.getDataSet(DataSetRequest.Builder.of("MULTPL/SP500_REAL_PRICE_MONTH").build());
sp500 = sp500.sortOn("Date")
def plot = new TimePlot(title: 'Price Chart for S&P 500', xLabel: 'Time', yLabel: 'S&P 500')
plot << new Line(x: sp500.column("Date").collect(), y: sp500.column("Value").collect(), color: Color.blue)
Color.red)
plot.autoZoom = true
```

It didn't seem to work:

>groovy.lang.MissingPropertyException: No such property: autoZoom for class: com.twosigma.beakerx.chart.xychart.TimePlot
	at script15165988516921202287947.run(script15165988516921202287947.groovy:4)
	at com.twosigma.beakerx.groovy.evaluator.GroovyCodeRunner.run(GroovyCodeRunner.java:71)

I tried a different way:

```
sp500 = session.getDataSet(DataSetRequest.Builder.of("MULTPL/SP500_REAL_PRICE_MONTH").build());
sp500 = sp500.sortOn("Date")
def plot = new TimePlot(title: 'Price Chart for S&P 500', xLabel: 'Time', yLabel: 'S&P 500')
plot << new Line(x: sp500.column("Date").collect(), y: sp500.column("Value").collect(), color: Color.blue)
Color.red)
plot.setAutoZoom(true)
```

Still got an error:

>groovy.lang.MissingMethodException: No signature of method: com.twosigma.beakerx.chart.xychart.TimePlot.setAutoZoom() is applicable for argument types: (java.lang.Boolean) values: [true]
	at script15165986766451720558403.run(script15165986766451720558403.groovy:4)
	at com.twosigma.beakerx.groovy.evaluator.GroovyCodeRunner.run(GroovyCodeRunner.java:71)

PR #6698, opened 2018-01-23
#6691 add autoZoom support to CombinedCharts

markhalonen:


markhalonen:
This code should've been in #6575, I overlooked it.

scottdraves:
thank you, but i wasn't able to get this to work on a combined plot.
can you post the exact code you ran, and how you interacted with it?

markhalonen:
If you were trying to use `autoZoom` with Python `CombinedChart`'s that wasn't working. I have fixed that in d198ea9

So the original PR (#6608) added `autoZoom` to regular charts. `CombinedChart`'s do not inherit from regular charts, so `autoZoom` had to added for `CombinedChart`'s in this PR

markhalonen:
And then it should work just like any other chart option: `cplot = CombinedPlot(xLabel= "Linear", autoZoom=True)`

Issue #6423, opened 2017-11-30
port BeakerX tree panel to Lab


scottdraves:
the brand, version, credit and the JVM options form.

PR #6711, opened 2018-01-24
#6423 port port BeakerX tree panel to Lab

piorek:
image


scottdraves:
cool, works well in Lab.
can you just change the name under which its listed from "BeakerX - Tree" to "BeakerX Options"?

piorek:
> can you just change the name under which its listed from "BeakerX - Tree" to "BeakerX Options"?

changed

scottdraves:
awesome, works great in nb and lab :)

one glitch: put a number in the heap field, like 10.  let it sync.
then click on the field and hit backspace twice.  let it sync.
i get a NaN in the preview line
screen shot 2018-02-01 at 10 28 11 pm

piorek:
fixed

Issue #6712, opened 2018-01-24
extra digits in table read from CSV


scottdraves:
<img width="528" alt="screen shot 2018-01-24 at 9 47 48 am" src="https://user-images.githubusercontent.com/963093/35338411-e283c53a-00eb-11e8-8610-c96940f5c015.png">
and yet there are only 4 digits in the file:
```
"m3","y30","y1","m6","y2","y10","y3","time","y5","y7","spread"
"7.8981","8.2586","7.9210","7.9562","8.0852","8.2067","8.1324","1990-01-30 19:00:00.000 -0500","8.1195","8.1962","0.3086"
```

scottdraves:
https://github.com/twosigma/beakerx/blob/master/kernel/base/src/main/java/com/twosigma/beakerx/table/TableDisplayConverter.java#L46
should be double.
and 
https://github.com/twosigma/beakerx/compare/jarek/6712_extra_digits

scottdraves:
neither of these things fixes the problem for me.

jaroslawmalekcodete:
please try to use "gradle clean build"

EfimovVladimir:
after I rebuild by "gradle clean build" this error was fixed for me

scottdraves:
thanks that did it and it worked for me too.  please make a PR.
i would like to know why this upgrade works though, ie what issue or code change is responsible?
i tried searching the issue tracker but couldn't find anything.

dsblank:
I think that the "problem" is that the float version isn't close enough to the representation that ends in zeros, so it looks like it has more precision, when it really has less. It's just that the float's precision isn't close enough.

See also:

https://stackoverflow.com/questions/22970240/parsefloat-vs-parsedouble

scottdraves:
yes, toString() on a float knows to stop when it reaches the ULP but it seems jackson is printing with something else that did not?

scottdraves:
https://github.com/FasterXML/jackson-databind/issues/707

dsblank:
Ah, good find!

PR #6721, opened 2018-01-25
#6712 number format changed to double for CSV number converter

lmitusinski:


scottdraves:
https://github.com/twosigma/beakerx/issues/6720

Issue #6697, opened 2018-01-22
rename `beakerx-install` to `beakerx install`


scottdraves:
also support `beakerx uninstall`
`beakerx --help`
`beakerx --version`
`beakerx` by itself should be the same as `jupyter notebook`

see https://github.com/twosigma/beakerx/issues/6667

jmsdnns:
@lmitusinski would love to catch up with you before you start. I imagine we're already thinking in the same direction, but I have some extra context I'd like to add to your thoughts.

scottdraves:
james pointed out that argparse has support for subcommands https://docs.python.org/3.5/library/argparse.html#sub-commands

jmsdnns:
Even though we discussed this, I'm going to add the context from our discussion here for the sake of this thread being complete to readers.

argparse calls subcommands "subparsers". You have the top level command, eg `beakerx`, and then if the next string entered matches the name of a subparser, the subparser is invoked for processing the rest of it.

Here is an example from a personal project called MarkUpUpDownDown. The top level command is here: https://github.com/MarkUpUpDownDown/markupupdowndown/blob/master/markupupdowndown/commands/__init__.py and we see that it imports other modules for the rest of the commands.

The `init_subparser` method is my own convention, so use it if you like it or don't if you prefer something else: https://github.com/MarkUpUpDownDown/markupupdowndown/blob/master/markupupdowndown/commands/push.py#L5 The main idea to take from `init_subparser` is how easy it can be to link multiple modules together into a robust command suite, should we need one.

As Scott suggested via conversation, we are probably fine with all of the commands together in a single file.

PR #6727, opened 2018-01-26
#6697 beakerx commands refactor

lmitusinski:


beakerx/beakerx/commands.py
```diff
@@ -0,0 +1,44 @@
+import argparse
```
scottdraves, line 14: banner

beakerx/beakerx/commands.py
```diff
@@ -0,0 +1,44 @@
...
+
+def bkr2ipynb_subparser(subparser):
+    bkr2ipynb_parser = subparser.add_parser('bkr2ipynb', help='converts beaker notebooks to ipynb format')
+    bkr2ipynb_parser.set_defaults(func=main)
+    bkr2ipynb_parser.add_argument('notebooks', nargs='+',
+                                  help="beaker notebooks to be converted. Enter *.bkr in case you want to convert all notebooks at once.")
+    return subparser
+
+def init_parser():
+    version = 'beakerx %s' % (beakerx.__version__)
```
scottdraves: print just the version do not include the name.

beakerx/beakerx/commands.py
```diff
@@ -0,0 +1,44 @@
+import argparse
+import sys
+import subprocess
+import beakerx
+from .install import install, uninstall
+from .bkr2ipynb import main
+
+def install_subparser(subparser):
+    install_parser = subparser.add_parser('install', help='installs beakerx extension')
```
scottdraves: help='installs BeakerX extensions'

beakerx/beakerx/commands.py
```diff
@@ -0,0 +1,44 @@
...
+    bkr2ipynb_parser.add_argument('notebooks', nargs='+',
+                                  help="beaker notebooks to be converted. Enter *.bkr in case you want to convert all notebooks at once.")
+    return subparser
+
+def init_parser():
+    version = 'beakerx %s' % (beakerx.__version__)
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--version', action='version', version=version)
+    parser.set_defaults(func=lambda x : subprocess.check_call(["jupyter", "notebook"]))
```
scottdraves: if you do it this way, then there are 2 problems:
1) arguments are not passed,
2) control-c is not passed to the underlying process, so any kernels the user started are leaked.
maybe instead "import jupyter" and call it as a function?

beakerx/beakerx/commands.py
jmsdnns: Additional context, the Python community standardized the output for versions in [PEP 440](https://www.python.org/dev/peps/pep-0440/).

Here is what Django does, and notice they reference PEP 440: https://docs.djangoproject.com/en/2.0/ref/django-admin/#determining-the-version

I see that my example from MarkUpUpDownDown [puts the app name in the version string](https://github.com/MarkUpUpDownDown/markupupdowndown/blob/master/markupupdowndown/commands/__init__.py#L14), so that's my bad for misleading you there. Sorry, @lmitusinski.

Issue #6598, opened 2018-01-04
shift + tab does not work for Scala, Groovy, and possibly other kernels. Works for Python.


ArnoldLipton:
This is my first foray into Jupyter Notebooks. My background is in Java and Scala development in classical IDEs (e.g. IntelliJ, Eclipse).

I have Jupyter Notebook 5.0.0 installed with only two nbextensions: jupyter-js-widgets and beakerx 0.9.1.

My Python 3 notebooks allow me to use the tooltip feature (i.e. if I were to click shit+tab inside a method's parenthesis, I get expected arguments). However my Scala and Groovy notebooks do not (but I can run the cells just fine).

Is this expected or am I missing something? If this is expected, how do folks author in the notebooks without the ability to discover method arguments like you can in a standard IDE?

Thank you.

scottdraves:
We haven't implemented docstring support on shift-tab yet, but version 0.10.0 (just released) does have a help menu item that brings up the javadoc for the runtime.

In general, since classfiles don't contain comments, the doc is normally not available:
https://stackoverflow.com/questions/32152489/why-jar-files-do-not-contain-documentation

what methods' doc were you looking for, BeakerX code or 3rd party libs?

ArnoldLipton:
Thanks for the prompt reply.

This is surprising especially to somebody who's lived in mature IDEs for a while; without docs easily accessible at authoring time how do folks discover APIs? Do they have the source code or Google open somewhere and just look for what they need? 

To answer your question, for the case at hand, I was looking for BeakerX's CsvPlotReader and SimpleTimePlot docs and ended up just going to the source code in GitHub (which really slows down authoring).

scottdraves:
Cool, the JavaDoc in 0.10.0 should help you with that.  And you are right, it should appear on shift-tab. https://github.com/twosigma/beakerx/issues/6599

i am not really familiar with stds for getting javadoc along with the regular jars, will have to look into how to improve that.  thank you for the question.

scottdraves:
http://maven.apache.org/plugins/maven-eclipse-plugin/examples/attach-library-sources.html

benmccann:
I like this suggestion. It'll be a really big help. The keyboard shortcut I'm most familiar with that's similar to this is Ctrl + Space for content assist in Eclipse

PR #6728, opened 2018-01-26
#6598 display javadoc on shift-tab

rbidas:


doclet/build.gradle
```diff
@@ -0,0 +1,25 @@
+group 'com.twosigma.beakerx'
```
scottdraves: banner

scottdraves:
move doclet directory into kernel directory so it can share the same gradlew

scottdraves:
i am hitting shift-tab with a classname open, but no window is open...
what am i missing?

benmccann:
I was wondering if this is documented anywhere yet? If not, it would be good to add it to the docs as well

scottdraves:
Cool, the help page opened up, but I only see the class name, not the actual doc?
Looks like it needs a bit more margin on top because the buttons are blocking it.

screen shot 2018-02-27 at 11 31 19 am 2

doclet/gradlew
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env sh
```
scottdraves: move the doclet into the kernel directory so that it shares the gradle stuff.

scottdraves:
cool thx.  during compilation i am seeing a bunch of errors like this:
```
    Constructing Javadoc information...
    /Users/spot/src/beakerx/kernel/base/src/main/java/com/twosigma/beakerx/handler/KernelHandler.java:23: error: package com.google.common.base does not exist
    import static com.google.common.base.Preconditions.checkNotNull;
                                        ^
    /Users/spot/src/beakerx/kernel/base/src/main/java/com/twosigma/beakerx/handler/KernelHandler.java:23: error: static import only from classes and interfaces
```
we should not even being generating javadoc for internal classes, just for the user-visible runtime classes.  the regular javadoc had a filter for this...  or maybe the errors can just be fixed, idk.

PR #6731, opened 2018-01-26
#5798 scala getters setters

jpallas:
This is the Scala plotting classes setters/getters work for #5798.  There were some dead ends before I arrived at a workable approach, which I think this is.

There's room for adding some convenience creation methods in companion object `apply` methods, but I wanted to get this much pushed out for review.

This *will* require changes to existing notebooks that used the previous constructors.  There was no reasonable way to make the constructors type safe.


jpallas:
The matchers from assertj just refused to work right in Scala, which is why I added Scalatest.  I would have been okay with writing the Scala tests in the same style as the Java tests if I could.

scottdraves:
thank you joe that sounds awesome i look forward to trying it out.
am i right that this is item 4 from https://github.com/twosigma/beakerx/issues/5798?
i think before you were planning on doing this with ScalaFX style, did that pan out?

jpallas:
Yes, this is the 4th item from #5798.  This is the ScalaFX-style approach described in https://github.com/twosigma/beakerx/issues/5798#issuecomment-345570360 after I decided that overloading the existing Java methods was not workable.  Most of the code is just forwarding Scala-style getters and setters to bean-style methods, with appropriate conversions between Java and Scala types where needed.  (I considered trying to use macros to reduce the boilerplate, but Scala macros are still in flux and definitely not for the faint of heart—I've never written one—and I'm not sure it would have substantially reduced the amount of code required.)

The goal I ended up with was to make Scala notebook plotting almost as concise as it is in Groovy (modulo dealing with static typing) without needing constructors with dozens of optional arguments.  I think this does a decent job.

All that said, if the plotting library gets replaced, this instantly becomes legacy code :vhs:.

scottdraves:
🐬 🐬 

this is amazing thank you.  effort will not be wasted even when/if https://github.com/twosigma/beakerx/issues/6572 happens for a couple of reasons:
1) we will continue to support the current (and now improved) API and,
2) any new API will be based on lessons of current API.

jpallas:
Thanks.  I know I got some education out of it, including a better understanding of Scala's type class pattern and a gentle reminder of why unit tests for one-line methods are sometimes necessary.

PR #6742, opened 2018-01-30
jarek/6617: add output widget

jaroslawmalekcodete:


scottdraves:
this works great but would like to expand it to work with rich output, not just stdout/err.
could output have a "display" method?
and catch the input to `display()`?

EfimovVladimir:
It looks good.  (run e2e tests).  Mime.ipynb and JavaWidgets.ipynb notebooks also works fine.

Issue #6432, opened 2017-12-01
Incorrect type inference results in date parsing error ("Unknown string format")


ghost:
| Version |
|-----------|
| 0.9.1     |

I've happened across a bug where BeakerX is trying to automatically parse a string field that looks "date-like" but that isn't actually a date (i.e. is an `[object]` type in the dataframe rather than a `datetime64[ns]` type). I came across this while trying to display a dataset that contains references to particular runways (string types containing integers, usually in the range 0-36). It parses the first few rows of my dataset happily enough (as these rows coincidentally are < 31 and therefore - I guess - get parsed as days). However, if BeakerX encounters a string that doesn't match this assumption then it will throw an error.

## Minimum working example

```python
import pandas as pd
from beakerx import *

df = pd.DataFrame({"runway": ["24", "36L"]})

TableDisplay(df)
```

## Error

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-3-a6495998b2cd> in <module>()
----> 1 TableDisplay(df)

/opt/anaconda3/lib/python3.6/site-packages/beakerx/tabledisplay/tabledisplay.py in __init__(self, *args, **kwargs)
    175     def __init__(self, *args, **kwargs):
    176         super(TableDisplay, self).__init__(**kwargs)
--> 177         self.chart = Table(*args, **kwargs)
    178         self.model = self.chart.transform()
    179 

/opt/anaconda3/lib/python3.6/site-packages/beakerx/tabledisplay/tabledisplay.py in __init__(self, *args, **kwargs)
     31         self.hasIndex = None
     32         if isinstance(args[0], DataFrame):
---> 33             self.convert_from_pandas(args, types_map)
     34         elif isinstance(args[0], dict):
     35             self.convert_from_dict(args)

/opt/anaconda3/lib/python3.6/site-packages/beakerx/tabledisplay/tabledisplay.py in convert_from_pandas(self, args, types_map)
    104                 value_type = types_map.get(columnName)
    105 
--> 106                 row.append(self.convert_value(value, value_type))
    107             if not isinstance(args[0].index, RangeIndex):
    108                 row[:0] = [self.convert_value(args[0].index.get_values()[index], type(args[0].index.get_values()[index]))]

/opt/anaconda3/lib/python3.6/site-packages/beakerx/tabledisplay/tabledisplay.py in convert_value(value, value_type)
    120     def convert_value(value, value_type):
    121         if value_type == "time":
[...]

PR #6753, opened 2018-02-01
#6432 fixed python date recognition

lmitusinski:
Date parsing errors are caused by false positive results from is_date function. I changed its implementation so now it is based on regex, not parser.


scottdraves:
closed in favor of https://github.com/twosigma/beakerx/pull/6757

Issue #6214, opened 2017-10-31
add UI options


scottdraves:
in the BeakerX tab of the tree view, add a new tabbed view.  the current JVM options would all be on the first, top tab "JVM Options".  Add a new tab "UI Options" with the following configs:
1) checkbox to auto close brackets.
2) checkbox to allow wide code cells.
3) checkbox to make our font changes.
4) checkbox to show publication UI (menu item and toolbar button)

by default 2 and 3 and 4 are checked.  status saved in beakerx.json.

scottdraves:
```
codecell.CodeCell.options_default.cm_config.autoCloseBrackets = false;
```

scottdraves:
```
<style>.container { width:100% !important; }</style>
```

scottdraves:
```
{
    "beakerx": {
        "jvm_options": {
            "heap_GB": 4,
            "other": [],
            "properties": [
                {
                    "name": "myproperty",
                    "value": "myval"
                }
            ]
        },
        "ui_options": {
            "auto_close": false,
            "improve_fonts": true,
            "wide_cells": true
        },
        "version": 2
    }
}
```

scottdraves:
for notebook only, not lab.

piorek:
@scottdraves 

I've done:
- wide_cells
- auto_close
- show_publication

Don't know what improve_fonts should do

PR #6764, opened 2018-02-02
#6214 add UI options

piorek:


scottdraves:
looks good but none of the options do anything.
also, they should be in a tab (with the JVM options in the 1st tab).
and align the checkboxes.

PR #6766, opened 2018-02-02
add e2e tests for TableDisplay for pandas

EfimovVladimir:


scottdraves:
the table we really need to test is from that regression:
```
import pandas as pd
from beakerx import *

df = pd.DataFrame({"runway": ["24", "36L"]})

TableDisplay(df)
```

EfimovVladimir:
done for regression

PR #6780, opened 2018-02-05
add e2e tests for publish notebook

EfimovVladimir:


scottdraves:
clear the widget state from the test notebook so it just has code source, and do a "run all" before publishing so it has the widgets to test.

EfimovVladimir:
done

Issue #6745, opened 2018-01-30
Adding password fields in EasyForm


bjonnh:
I almost managed to get Passwords added to EasyForm.

See: 
https://gist.github.com/bjonnh/2ee3fe9a263f382593c06078feaeb2e9

It uses the undocumented Password field from ipywidgets.

However, there needs to be some JS code that goes with it and would be similar to : 
https://github.com/twosigma/beakerx/blob/c29502ab9c836b5eadc3a98d1c47b063edcc95fc/js/notebook/src/easyForm/textWidget.ts

as i'm working on a remote instance of beaker, I can't do that from a notebook I guess and I will have to change the beaker code directly.

scottdraves:
Hi, thanks for the feedback.

You should be able to just include the ipywidet's Password widget into the easyform, and the Groovy Easyform API has that, but it was left off the the Python version by mistake.
I made an issue to fix it: https://github.com/twosigma/beakerx/issues/6746

or do you really want an BeakerX component for passwords?  that's also possible.

bjonnh:
That should do it yes. Thanks.

PR #6792, opened 2018-02-06
#6745 password beakerx widget

lmitusinski:


scottdraves:
hitting the enter key with the cursor in this field did not submit the form.

Issue #6773, opened 2018-02-03
output container test


scottdraves:
doc/groovy/OutputContainers.ipynb

PR #6796, opened 2018-02-07
#6773 Groovy Output Containers tests

wojciechowskim:
Connected with issue: #6773


scottdraves:
i added a visible result ot the 1st cell:
```
     "x = big(10)\n",
-    "OutputCell.HIDDEN"
+    "OutputCell.HIDDEN\n",
+    "10"
    ]
```
but the test still passed...

test/tests/groovy/outputContainersTest.js
```diff
@@ -0,0 +1,138 @@
...
+
+  describe('(Groovy) Cyclical Output Containers', function() {
+    it('Charts render correctly on every interval', function() {
+      cellIndex += 1;
+
+      var dtContainer = beakerxPO.runCellToGetDtContainer(cellIndex);
+
+      plotHelper.checkXLabel(dtContainer, '');
+
+      browser.pause(3000);
```
scottdraves, line 132: this doesn't look good.  instead it should just block until the next plot appears.
it should do that 5x until each plot has been confirmed to appear, and back to the 1st.

test/tests/groovy/outputContainersTest.js
```diff
@@ -0,0 +1,138 @@
...
+      expect(beakerxPO.getTabLabel(0).getText()).toBe(testValues.tabbedOutputScatter);
+      expect(beakerxPO.getTabLabel(1).getText()).toBe(testValues.tabbedOutputShortTerm);
+      expect(beakerxPO.getTabLabel(2).getText()).toBe(testValues.tabbedOutputLongTerm);
+      expect(beakerxPO.getTabLabel(3).getText()).toBe(testValues.tabbedOutputDate);
+    });
+
+    it('Scatter with history chart is displayed properly', function() {
+      var dtContainer = beakerxPO.runCellToGetDtContainer(cellIndex);
+
+      expect(dtContainer.isEnabled()).toBeTruthy();
```
scottdraves, line 87: i don't understand what this is really checking.  shouldn't it use the testValues and make sure the displayed tab  is the right one?

EfimovVladimir, line 87: for example, it will be right to check `svg` elements as `expect(svgElement.$('#i0.plot-point').isVisible()).toBeTruthy();` in this case

PR #6798, opened 2018-02-07
#6214 add UI options

piorek:


scottdraves:
cool this is much better.  
1) the UI Options and JVM Options sections should be in a tabbed container.
2) show publication should be enabled by default.
3) auto close brackets is being interpreted in reverse (if the checkmark is ON then brackets should be auto closed).
4) for wide cells, can you keep a 1cm margin on the left? don't use up quite all the space.
5) something is wrong with the default options.  if i delete ~/.jupyter/beakerx.json, run jupyter, and then create a new noteobook, then the wide cells are not applied.    but if i go into the options panel, click "wide cells" twice, and then create a new noteobok, the wide cells are applied.
6) remove the "improve fonts" checkbox until it's implemented.

piorek:
> 1. the UI Options and JVM Options sections should be in a tabbed container.

Tried to do it with phosphors TabPanel but having styling issues. 
Turns out its not designed to have Panel children and parents.

Still trying to make it work or use bootstraps tabbed component

> 2. show publication should be enabled by default.

done

> 3. auto close brackets is being interpreted in reverse (if the checkmark is ON then brackets should be auto closed).

should be ok now

> 4. for wide cells, can you keep a 1cm margin on the left? don't use up quite all the space.

margin added

> 5. something is wrong with the default options. if i delete ~/.jupyter/beakerx.json, run jupyter, and then create a new noteobook, then the wide cells are not applied. but if i go into the options panel, click "wide cells" twice, and then create a new noteobok, the wide cells are applied.

changed default json in python should be ok now

> remove the "improve fonts" checkbox until it's implemented.

checkbox hidden (commented)

Issue #6772, opened 2018-02-03
show null config test


scottdraves:
doc/groovy/showNullExecutionResult.ipynb

PR #6799, opened 2018-02-07
#6772: show null execution result tests

wojciechowskim:
Connected with issue #6772


scottdraves:
i removed the code that turns off showing null:
```
-    "com.twosigma.beakerx.kernel.Kernel.showNullExecutionResult = false;\n",
+    "//com.twosigma.beakerx.kernel.Kernel.showNullExecutionResult = false;\n",
```
and yet the test still passed, so something must be wrong.

Issue #6814, opened 2018-02-09
The "Publish..." button is disappeared


EfimovVladimir:
After last update from master the "Publish..." button is disappeared.

![image](https://user-images.githubusercontent.com/17744188/36027574-f71a5c16-0dac-11e8-9eb7-4ca0d3f154dd.png)

console:
![image](https://user-images.githubusercontent.com/17744188/36027728-75e5bedc-0dad-11e8-831c-dabd2165ca92.png)

reproduced on windows

PR #6815, opened 2018-02-09
#6814 fix The "Publish..." button is disappeared

piorek:


piorek:
to reproduce issue and test if this works:

edit beakerx.json and remove ui_options from it

EfimovVladimir:
:+1:  works fine

Issue #6730, opened 2018-01-26
update values in TableDisplay without reinstantiation


themeshter:
I am trying to use TableDisplay as part of an interactive Python cell. All is good so far - values get displayed as I want them to, so no concerns there.
The values in the table should change based on some user input - so the problem that I have is that I would like to programatically update the values without reinitializing the TableDisplay object?
 
I looked over the documentation and the only examples I found are based on constructing the table with the values preinitialized and then using it do display. Is it any example that shows how can you update the values or this feature is not supported? If this is not supported, can you point me out to what the best practice would be in this scenario (refreshing the data in a component)?

scottdraves:
Hi thanks for the question.  That should be possible, there is no direct documentation of that though, you are right.  Check out the use of `setDoubleClickAction` though, it does it.  I will look into expanding the doc on this point.

scottdraves:
this would be a nb like doc/groovy/PlotsPythonStyle.ipynb but that changes numeric and string values in the table.

scottdraves:
we have this for groovy, need it for python now too.

PR #6816, opened 2018-02-09
#6730: tableDisplay update value

jaroslawmalekcodete:


scottdraves:
on 2nd thought, i think i just want sendModel to work for the user.  not sure what's wrong this this:
screen shot 2018-02-13 at 1 06 51 pm

jaroslawmalekcodete:
Instead of mapListToUpdate[0]["a"] use tableToUpdate.values[0][0] = 99
like here 
screenshot from 2018-02-14 12-36-19

scottdraves:
OK cool, then we need a way to refer to the data by column name and row instead of just by index.

jaroslawmalekcodete:
updatecellbycolumnname

PR #6844, opened 2018-02-13
Add more colors

benmccann:
I was really surprised that purple didn't exist already. And then I kind of got on a roll. I mostly added blues, greens, reds because they have good contrast against white background


scottdraves:
thanks 👍

PR #6860, opened 2018-02-14
jarek/6852:  rename more packages

jaroslawmalekcodete:


scottdraves:
now that I look into this one more closely I changed my mind, it should be like this:

com.twosigma.beakerx.widget.*.X
→
com.twosigma.beakerx.widget.X

Issue #6722, opened 2018-01-25
make a demo of STIL integration (astronomy formats)


scottdraves:
http://www.star.bristol.ac.uk/~mbt/stil/

scottdraves:
https://github.com/Starlink/starjava

mbtaylor:
I'm author of starjava/stil/stilts, and looking at integrating them somehow or other with Jupyter has been on my to-do list for a while, but I'm not sure (a) how to get started or (b) how useful it would actually be.

It's still not close enough to the top of my to-do list to actually get down to doing it, but if any jupyter/beakerx expert wants to have a go, I'm very willing to pitch in with advice or assistance from the stilts/starjava side.

There is already a jython front-end to stilts (http://www.starlink.ac.uk/stilts/sun256/jystilts.html), which can pop up interactive plots, amongst a load of other table operations.  From a user-interface point of view that would be a good starting point.  

Thanks to @maartenbreddels for pointing me at this.

scottdraves:
Hi Mark great to hear from you.  Getting STIL to load was easy since you provide a jar. I just ran
```
%classpath add jar stil.jar
```
then I was then able to call `makeStarTable()` but I don't have a file to read.  Will definitely pursue this further, and report back here.

Indeed thank you Maarten for the intro.

mbtaylor:
Plenty of data files to choose from.  You probably have or can find some CSV files lying around; a couple of medium-sized astronomy FITS files (more efficient) if you want them are ftp://andromeda.star.bris.ac.uk/pub/star/data/dr5qso.fits, http://andromeda.star.bristol.ac.uk/data/tgas_source.fits.

scottdraves:
Cool, CSV works: https://nbviewer.jupyter.org/gist/anonymous/6f68bb70a73945aaea5b94c650fbb28d

Will see about writing an adaptor so the tables are displayed with our GUI automatically.

scottdraves:
here's a live demo:
https://mybinder.org/v2/gh/twosigma/beakerx/0.12.1?filepath=doc%2Fgroovy%2FSTIL.ipynb

PR #6873, opened 2018-02-16
improve STIL example (#6722)

mbtaylor:
Fix the STIL example so that it accesses the StarTable object
cell values directly rather than just reading all the cells
into an internal array and accessing that.

The StarTable object requires random access for this to work;
tables read from CSV files are not random-access by default,
so the Tables.randomTable method has to get called on it.

In fact this is just moving the copy-to-internal-array step
out of the groovy and into the STIL library, so in this case
it doesn't make a huge amount of difference.  However, if the
input file was e.g. a FITS file, it would mean that data access
was strictly on-demand to the mapped disk file, and so potentially
much more efficient, especially for large files.

### DIFF
--- a/doc/groovy/STIL.ipynb
+++ b/doc/groovy/STIL.ipynb
@@ -23,7 +23,7 @@
     "stilFile = System.getProperty(\"java.io.tmpdir\") + \"/stilFiles/stil.jar\"\n",
     "FileUtils.copyURLToFile(new URL(stilUrl), new File(stilFile));\n",
     "\n",
-    "%classpath add dynamic stilFile"
+    "%classpath add dynamic stilFile\n"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


scottdraves:
Hi Mark thanks that looks awesome, can you please send a CLA (see https://github.com/twosigma/beakerx/blob/master/CONTRIBUTING.md)?
thanks

mbtaylor:
OK I'll do it in the next few days.

mbtaylor:
I have sent the CLA as requested.

scottdraves:
that's awesome, but before it would actually display the table with the interactive widget.
what's the best way to bring that back?

mbtaylor:
I still see the interactive table widget after execution of the final cell, as before.  If you're not seeing it, maybe it means that the table isn't getting created for some reason?  Does something like `starTable.getColumnCount()` work? 

Not sure where else to go with that, I'm afraid my experience in remote debugging of jupyter notebooks is extremely limited.

scottdraves:
yes getColumnCount works but I am getting the regular text rep, as if the Displayer config didn't trigger.
we'll look into it.

scottdraves:
screen shot 2018-02-21 at 11 34 51 am

scottdraves:
could not reproduce this

PR #6874, opened 2018-02-16
jarek/6860: rename widget packages

jaroslawmalekcodete:


jaroslawmalekcodete:
followup to https://github.com/twosigma/beakerx/pull/6860

Issue #6809, opened 2018-02-08
sections of options should use tabs


scottdraves:
in the BeakerX panel of the tree view, the UI Options and JVM Options sections should be in a tabbed container.

scottdraves:
remaining from https://github.com/twosigma/beakerx/pull/6798

PR #6883, opened 2018-02-20
#6809 sections of options should use tabs

piorek:


scottdraves:
cool this looks good but there are some glitches:
1) when i first click on the "BeakerX" tab the JVM options are not appearing:
screen shot 2018-02-21 at 11 26 42 am
i have to click on the JVM options tab header for them to appear (or click on the UI tab and then back to JVM).
2) now that the title "JVM Options" appears in the tabs, no need to repeat it in the contents.  ditto for "UI Options".

piorek:
Should be OK now

scottdraves:
Thanks that worked but please address:

> now that the title "JVM Options" appears in the tabs, no need to repeat it in the contents. ditto for "UI Options".

that also means eliminating the ruler line that goes with the heading.

piorek:
Done

scottdraves:
Thanks, that's that's good but I still see a couple of problems:
1) only show the "Result: java ...." line (including the sync symbol) in the JVM options panel.
2) the panel contents are truncating the descenders (the "p") in the text of the labels:
screen shot 2018-02-24 at 3 37 58 pm

piorek:
done

Issue #6841, opened 2018-02-13
new test: MIME display


scottdraves:
https://github.com/twosigma/beakerx/blob/master/doc/groovy/Mime.ipynb

EfimovVladimir:
also need implement for jupyterlab

PR #6959, opened 2018-03-08
#6841 MIME display tests

wojciechowskim:
Connected with issue: #6841


scottdraves:
1st test has a race condition:
```
[chrome #0-0] Error: Expected 'α
[chrome #0-0] 2
[chrome #0-0] +η' to be 'α2+η'.
[chrome #0-0]     at UserContext.<anonymous> (/Users/spot/src/beakerx/test/js/groovy/mimeTest.js:48:57)
[chrome #0-0]     at new Promise (<anonymous>)
```
failed the 1st time i ran it, passed the 2nd.

mathjax is used to format the math, and it runs async after the cell output appears, which makes testing for it harder than regular outputs.

Issue #6957, opened 2018-03-08
Scala completion is not working correctly


jpallas:
This has been broken for a while, I think.  I'm not sure exactly since when, because the behavior is a little bit confusing.  Autocompletion sees the standard library and beaker-specific imports, but it doesn't see values defined by the user.  At least, that's my interpretation of the behavior.

So, concretely:  If the first cell contains
```
val xyzzy = "hello"
```
then we would expect that typing `xyz<TAB>` in the second cell would offer the completion `xyzzy`.  But it doesn't.

My intuition says this has something to do with `ScalaEvaluator`'s `acshell`.  I don't understand the Scala kernel autocomplete stuff, but it seems odd to have a separate evaluator just for completions.  Does this have something to do with class loaders?  I don't know.

jpallas:
At this point my theory is that the existence of `ScalaEvaluator.acshell` is a mistake (possibly based on some code that pre-dated the Scala 2.11 presentation compiler, but that's just a guess).  The presentation-compiler-based autocomplete shouldn't mess with the state of the interpreter.  Evaluating code specifically for the side-effects in the namespace (in `ScalaEvaluator.autocomplete`) is a serious error.  (Example: don't try hitting `abc<TAB>` on line 2 if line 1 says `System.exit(0)`!)

I'm going to try excising `acshell` completely and see if anything breaks.  If there _is_ any kind of weird class-loading interaction with completion, though, I'm not confident that the existing tests will expose it.

scottdraves:
Cool thanks Joe

PR #6964, opened 2018-03-11
Scala completion improvement

jpallas:
Per #6957, this simplifies the completion code and adds tests to verify:
 * autocomplete should not have side effects (previous implementation did)
 * autocomplete should operate in the same namespace context as the compiler (previous implementation did not)

It's possible that something subtle was involved in the previous implementation.  I haven't encountered anything odd in my testing, though.


scottdraves:
awesome, thank you Joe!

Issue #6926, opened 2018-03-02
onClick of CombinedPlot broken


scottdraves:
```
def xs = (1..5)
def cplot = new CombinedPlot(initWidth: 800)
def random = new Random()
def p1 = new Plot(useToolTip: false);
p1 << new Line(x: xs, y: [5, 3, 7, 4, 5], width: 4).onClick({info -> info.graphics.y[1] = random.nextInt(10)})
cplot.add(p1, 2)
def p2 = new Plot(title: "Bars", useToolTip: false)
p2 << new Bars(x: xs, y: [3, 5, 2, 3, 7], outlineColor: Color.black, width: 0.3).onClick({info -> info.graphics.y[info.index]++})
cplot.add(p2, 3)
cplot
```
clicking on the items has no effect.

PR #6966, opened 2018-03-12
#6926 fix onClick of CombinedPlot broken

piorek:


scottdraves:
the plot details do not seem to be setup.  see test/ipynb/groovy/PlotActionsTest.ipynb the cell with tag61 is giving `java.lang.NullPointerException: Cannot get property 'graphics' on null object`

piorek:
Checked this on master and that NullPointerException was already there. 
I think it's separate issue. 

Do i fix it here or in separate PR?

Issue #6921, opened 2018-03-01
enable UI Options in Lab


scottdraves:
the "Show Publication" option makes sense.
So does the one about auto-closing brackets, and auto-save.

Hide the others.

PR #6991, opened 2018-03-19
#6921 enable UI Options in Lab

piorek:


scottdraves:
in notebook, the margin of the of checkboxes is too small now:

screen shot 2018-03-19 at 9 57 10 am

scottdraves:
in Lab, the visible switches all work, great!  however the margin is also missing, and the text of the checkbox labels is bold but it should be normal weight:
screen shot 2018-03-19 at 10 34 20 am

piorek:
fixed both

PR #6993, opened 2018-03-19
jarek/6513: spark magic command

jaroslawmalekcodete:


scottdraves:
cool this works great.  needs the following frontend changes:
screen shot 2018-03-19 at 11 39 39 am

1) move the text boxes to the right so the labels are not truncated.
2) redo the toggle to show/hide the cluster progress to be more standard at match the rest of the jupyter design.  right now it looks like this:
screen shot 2018-03-19 at 11 44 59 am
screen shot 2018-03-19 at 11 45 14 am
a jquery accordion titled "Spark Progress" would work.

jpallas:
I don't want to fall into the "that's not how I would have done it" trap, but I'd like to understand the design decision to link Spark into the kernel.  Spark promises binary compatibility for minor versions (unlike Scala), so it should be possible to compile against 2.2.1 and run against 2.3 or later.

scottdraves:
Happy to have your input on the right way to do it & am open to changes.
This is definitely a preliminary implementation intended for feedback.
I like the idea of making it totally loadable, will look into it.

jpallas:
On `SparkContext` _vs_ `SparkSession`: the Spark docs say that `SparkSession` (defaulting to `spark` in the Spark shell) is the preferred entry point in Spark 2.x, and all the examples in the documentation use it.  My preference would be to have the Spark magic default to giving you the same environment that the Spark shell does (including the automatic imports).

In the above scenario, I'm not sure what the body of a non-empty `%%spark` cell should return.  A `SparkSessionBuilder`, I guess.

I am uncertain about multiple sessions in a single kernel.  Is it possible?  Does it have value?  `SparkSession.conf` allows dynamic reconfiguration of some parameters, and `SparkSessionBuilder.getOrCreate` says it will reconfigure an existing session if it finds one.

scottdraves:
@jaroslawmalekcodete can you please add to this display of spark RDDs (org.apache.spark.rdd.RDD) with the table widget?

also, switch from SparkContext to SparkSession.

jpallas:
> can you please add to this display of spark RDDs (org.apache.spark.rdd.RDD) with the table widget?

Some considerations here:
1. RDDs and Datasets are evaluated lazily, unlike Pandas DataFrames.  Automatically displaying any part of the contents could trigger an undesired expensive computation.  
2. For Datasets, the `show()` operation computes and prints the first _n_ (20 by default) rows as a plain text table with columns.  A `display()` operation could do the same thing as a table widget (I've prototyped this).
3. Dataset `show()` works because Datasets have an explicit schema with column names.  RDDs don't have an equivalent operation; they would have to display as just strings (e.g., an index column and a value column).

scottdraves:
actually put the dataset display on its own PR for this issue: https://github.com/twosigma/beakerx/issues/7040

kernel/sparkex/src/main/scala/com.twosigma.beakerx.scala.spark/Implicit.java
```diff
@@ -0,0 +1,21 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package com.twosigma.beakerx.scala.spark;
+
+public interface Implicit {
```
jpallas, line 18: Could this just be a String?

scottdraves:
The cluster progress looks great, thanks!
However there's a problem which is for really quick jobs, it introduces a lot of vertical jumping.
Not sure what the best solutions is but how about this:

the output is first drawn, the Spark Progress is closed.
if the job takes more than 3s then it opens.
and like it does now, it closes when the job is finished.

jpallas:
I've confirmed that having two sessions in the same kernel does not work.  I suggest removing `SparkSessionAlias` and just using the session name `spark` for compatibility with the spark shell.

(Unless the idea is that you disconnect one and connect to another?  How would that be useful?)

scottdraves:
thanks agreed -- the intention isn't to support more than one, which afaik is not supported by spark at all, but just to allow you to pick the name.  it should be an advanced config that is not normally visible.
we'll get there.

PR #7031, opened 2018-03-23
jarek: downgrade jackson to 2.6.5 and fix test

jaroslawmalekcodete:


Ben-Epstein:
Why is this still downgraded? Was there a dependency issue?

Issue #6839, opened 2018-02-13
new test: java properties and heap size


scottdraves:
https://github.com/twosigma/beakerx/blob/master/doc/groovy/JavaArgs.ipynb
should set these in the JVM options panel and then open the test notebook and confirm the values.

EfimovVladimir:
also need implement for jupyterlab

EfimovVladimir:
#7451

PR #7046, opened 2018-03-26
#6839: java heap size and prop tests

wojciechowskim:
Connected with issue: #6839


test/js/groovy/javaArgsTest.js
```diff
@@ -0,0 +1,59 @@
...
+
+var BeakerXPageObject = require('../beakerx.po.js');
+var beakerxPO;
+
+describe('(Groovy) Java Properties and Heap Size tests', function () {
+  beforeAll(function () {
+    beakerxPO = new BeakerXPageObject();
+    beakerxPO.runJupyterRoot();
+    beakerxPO.goToJVMOptions();
+    beakerxPO.setHeapSize('4');
```
scottdraves: use 5

test/js/groovy/javaArgsTest.js
```diff
@@ -0,0 +1,59 @@
...
+
+  var cellIndex;
+
+  describe('Java Properties and Heap Size', function () {
+    it('Proper max memory is displayed', function () {
+      cellIndex = 0;
+
+      var output = beakerxPO.runCodeCellByIndex(cellIndex);
+
+      expect(output.$('.output_subarea').getText()).toBe('3.817865216');
```
scottdraves: only check the first two digits

test/js/groovy/javaArgsTest.js
```diff
@@ -0,0 +1,59 @@
...
+
+  var cellIndex;
+
+  describe('Java Properties and Heap Size', function () {
+    it('Proper max memory is displayed', function () {
+      cellIndex = 0;
+
+      var output = beakerxPO.runCodeCellByIndex(cellIndex);
+
+      expect(parseFloat(output.$('.output_subarea').getText()).toFixed(1)).toBe('3.8');
```
scottdraves: does that still work with 5?

scottdraves:
To verify that it's really setting the values it should set and check each value twice.
the 1st time with 5GB and "cookies", and then again with 3GB and "cream".
it should end by removing the 3 and removing the property (by clicking the "x"), confirming the property is now undefined.

scottdraves:
this is great!  there's a small problem though which is it leaves the heap configured at 3GB.

i think the best solution to this is before the tests run (in the python) it should move my prefs file to a safe location, and then move it back after the tests run (including if they are interrupted with control-C).

PR #7078, opened 2018-03-28
#7008 add drag'n drop column reorder

MariuszJurowicz:


scottdraves:
This has a problem when you drag a narrow column past a wide column, the ordering flickers back and forth.  Can solve by not updating the table as you drag, just show where the column will drop, and update on drop not on drag.

Issue #7099, opened 2018-03-29
table filter function not working with column names


scottdraves:
the table filter function should not just bind $ to the value of the current column, but also bind the names of all the columns to their values.  so for example the following should work:
![screen shot 2018-03-29 at 10 32 21 am](https://user-images.githubusercontent.com/963093/38094976-42c0ddec-333d-11e8-91be-13f6d900ad97.png)

PR #7105, opened 2018-03-30
#7099 filter with column names

MariuszJurowicz:


scottdraves:
thanks, this fixes the names, but now when i click "x" the row of filter boxes does not close.

Issue #7100, opened 2018-03-29
Investigate fix/workaround for Scala import issue


jpallas:
In [#7040](https://github.com/twosigma/beakerx/issues/7040#issuecomment-376997044) I included a code snippet with the comment 
> `// I do not understand why this import is necessary`

I did some further investigation, and the answer seems to be [Scala issue SI-9881](https://github.com/scala/bug/issues/9881).  This issue was fixed in the 2.12 line but the fix was not back-ported to 2.11 (see [this pull request](https://github.com/scala/scala/pull/6195) and [this issue](https://github.com/scala/scala-dev/issues/451) on the status of a back port—the question is whether there will be any more 2.11 releases.)

One reason for the lack of urgency is that the Spark project did their own back-port/workaround.  This is covered in [SPARK-22393](https://issues.apache.org/jira/browse/SPARK-22393).

The net result is code that works in the Spark shell may not work in the Scala kernel.  The Spark back-port is pretty small, but it is intimate with the REPL, so it is probably fragile.  (But the Spark team is in the same boat.)  It is worth looking into whether the same retrofit will work with the Scala kernel.

scottdraves:
Thanks Joe that's really interesting.  I don't see why we couldn't do the same.

jpallas:
Urf.  I tried stealing the Spark code, and it goes in cleanly.  But they used a different test case than the Scala code uses, and it turns out for whatever reason that the Spark code doesn't actually cover everything the Scala test case checks.

So, if the goal is anything that works in Scala should work in the kernel, it fails.  If the goal is anything that works in `spark-shell` should work in the kernel, it succeeds.

Also, there's a meta-question about what to do with the license/copyright headers on the files that are copied-with-minor-changes from Spark.

jpallas:
Added a comment at [SPARK-22393](https://issues.apache.org/jira/browse/SPARK-22393?focusedCommentId=16421548&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16421548) documenting the problem with the retrofitted solution.  I'm hoping someone who understands the import-handling code might figure out how to fix it, although I did not reopen the issue.

jpallas:
I figured out that there was one line missing from the Spark retrofit.  Adding that line makes the test pass.  I attached that information to the Spark issue, in hopes of getting some attention from the author of the retrofit.

[...]

PR #7121, opened 2018-04-03
Scala import fix

jpallas:
Fix for #7100.  This includes a line that is missing from the Spark retrofit of SI-9881.


scottdraves:
https://github.com/apache/spark/pull/19846/files

scottdraves:
thanks will check this out.

Issue #7024, opened 2018-03-23
keyboard navigation past edge of table


scottdraves:
if you use the arrow keys to move the cursor out of the view,
the table should scroll to keep it visible.

PR #7128, opened 2018-04-06
#7024 keyboard navigation past edge of table

piorek:


scottdraves:
This works great except for down-arrow should only scroll by 1 line, not a whole page.
Ditto for right-arrow.

Issue #7148, opened 2018-04-10
typescript unit tests fail in docker


scottdraves:
```
    $ yarn run test && webpack --config webpack.dev.js && yarn run build-lab-extension
    $ TS_NODE_PROJECT=test mocha


      BeakerxDataGrid

      BeakerxCellRenderer
--- Logging error ---
Traceback (most recent call last):
  File "/opt/conda/envs/beakerx/lib/python3.6/logging/__init__.py", line 994, in emit
    stream.write(msg)
UnicodeEncodeError: 'ascii' codec can't encode character '\u2713' in position 8: ordinal not in range(128)
Call stack:
  File "/opt/conda/envs/beakerx/bin/pip", line 11, in <module>
    sys.exit(main())
  File "/opt/conda/envs/beakerx/lib/python3.6/site-packages/pip/__init__.py", line 233, in main
    return command.main(cmd_args)
  File "/opt/conda/envs/beakerx/lib/python3.6/site-packages/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/opt/conda/envs/beakerx/lib/python3.6/site-packages/pip/commands/install.py", line 342, in run
    prefix=options.prefix_path,
  File "/opt/conda/envs/beakerx/lib/python3.6/site-packages/pip/req/req_set.py", line 784, in install
    **kwargs
  File "/opt/conda/envs/beakerx/lib/python3.6/site-packages/pip/req/req_install.py", line 845, in install
    install_options, global_options, prefix=prefix)
  File "/opt/conda/envs/beakerx/lib/python3.6/site-packages/pip/req/req_install.py", line 1007, in install_editable
    show_stdout=False)
  File "/opt/conda/envs/beakerx/lib/python3.6/site-packages/pip/utils/__init__.py", line 683, in call_subprocess
    logger.debug(line)
Message: '    \u2713 should create the columnManager'
Arguments: ()
```

scottdraves:
the tests have been disabled until they work in docker.

PR #7162, opened 2018-04-12
#7148 fixed env encoding in docker build

lmitusinski:


scottdraves:
still getting the same error

Issue #7112, opened 2018-04-02
groovy magic breaks on plot actions


scottdraves:
run this cell in a py3 notebook:
```
%%groovy
def random = new Random()
def p = new Plot(showLegend: true, useToolTip: false);
p << new Line(x: [1, 2, 3], y: [2, 3, 4], width: 10, displayName: "line 1")
         .onClick({info -> info.graphics.displayName = "new name"})

p << new Line(x: [1, 2, 3], y: [5, 6, 7,], width: 10, displayName: "line 2")
         .onClick({info -> info.graphics.y[1] = random.nextInt(10)})
```
groovy starts & the plot appears.  good.

click on a line.

expected: the line moves.  current: nothing on screen, but on the terminal it says:
```
[IPKernelApp] WARNING | No such comm: e58b2c6e-80d2-495d-9f25-c051f832de5e
```

PR #7192, opened 2018-04-18
#7112 handling communication with groovy magic kernel

lmitusinski:


lmitusinski:
With this commit this issue should be fixed too:
https://github.com/twosigma/beakerx/issues/6785

Issue #7061, opened 2018-03-26
allow dragging to resize table


scottdraves:
when the mouse is over the bottom edge of the bottom row, it could turn to vertical drag arrows,
allowing the user to click and drag and change the number of rows displayed.

when the mouse is over the right edge of the right column (to the right of the vertical scrollbar, if any), it should allow resizing horizontally.

and mouse in the bottom-right corner can resize in both directions.

PR #7197, opened 2018-04-19
#7061 DataGrid resize feature

MariuszJurowicz:


scottdraves:
has some issues.  after using it for a while i got this error:
```
VM99 index.js:138226 Uncaught TypeError: Cannot read property 'region' of null
    at EventManager.handleMouseDown (VM99 index.js:138226)
```
sometimes the blue rect got left on after mouse up.
try resizing to make the table as short as possible, then drag to make it tall.

scottdraves:
if you drag quickly to vertically to increase the size, and the mouse leaves the output area, then the resize stops.

if you drag to vertically to decrease the size, and the mouse moves above the table and leaves the output area, then the table jumps to very large size.

when i resize horizontally, it looks good, but my change is immediately undone.  i guess the handler for horizontal resize has triggered.  if this is not easy to fix then just remove horizontal dragging.

PR #7202, opened 2018-04-19
#7130 groovy magic expanded to other languages

lmitusinski:


lmitusinski:
I've added new notebook in doc/python/KernelMagics where I created examples for creating plot in all supported kernels (for sql there is a display of data from table)

PR #7212, opened 2018-04-20
jarek/7057: %classpath add results as HTML widget

jaroslawmalekcodete:


scottdraves:
works well, but resulting output spacing is like 1.5 normal
screen shot 2018-04-20 at 10 32 28 am

scottdraves:
Would be nice if the standard existing HTML widget could have the styles set but I can't figure out how: 
https://github.com/jupyter-widgets/ipywidgets/issues/2056

Issue #7221, opened 2018-04-23
column type tooltips stay up too long


scottdraves:
here are two ways to reproduce.  in both cases step 1 is the same:
1) run 1st cell of doc/groovy/TableAPI.ipynb

then for method A:
2) open the y5 column menu.  as the mouse goes over the heading to click the menu, the type tooltip appears.
3) change the format to "exponential 5".
4) the tooltip is still up even though the mouse has left the table, you can scroll down and run other cells and the tooltip stays on the screen.

and method B:
2) hover over any column header.  the tooltip appears.
3) use the mouse wheel to scroll.  the whole notebook moves, and the mouse leaves the header.
4) the tooltip stays on the screen.

PR #7227, opened 2018-04-24
#7221 refactor tooltips, fix column data tooltip frozen

MariuszJurowicz:


scottdraves:
tips are often not appearing at all...

Issue #7224, opened 2018-04-24
if mouse leaves widget while scrolling, focus is lost


scottdraves:
in the table widget, if you click the thumb in the scroll bar and drag, the table scrolls.
if you drag so far the mouse leaves the widget, the green focus is lost, even if the mouse stays down, and the scroll is in progress.
expected: keep focus so i can let go of mouse and switch to scroll wheel.

PR #7229, opened 2018-04-24
#7224 keep focus while scrolling

MariuszJurowicz:


scottdraves:
sorry i didn't clearly explain this.
if the mouse is released outside of the table, then it should lose focus (on release, not on exit).
the situation that's a problem is when the mouse leaves the table, and goes back in with the button still down, and then is released.

scottdraves:
scrollBar/BeakerXScrollBar is missing

Issue #7301, opened 2018-05-02
width of html column


scottdraves:
looks like it's getting computed from the underlying html src, not the result of the render.
![screen shot 2018-05-02 at 2 25 34 pm](https://user-images.githubusercontent.com/963093/39541806-cdce6ac0-4e14-11e8-9bd8-82b67d60fee8.png)

seems like by default it should take the widest.  maybe the HTMLStringFormat class should have an optional width though so user can specify it for the canvas?

PR #7314, opened 2018-05-04
#7301 set proper width for html column

MariuszJurowicz:


scottdraves:
screen shot 2018-05-04 at 9 25 03 am

we need to compute the width from the final rendered html, not by estimating using text.
also, if it wraps need to not be truncated, the row should expand.

Issue #7274, opened 2018-05-01
foldout for classpath add mvn


scottdraves:
<img width="589" alt="screen shot 2018-04-30 at 9 47 52 pm" src="https://user-images.githubusercontent.com/963093/39458113-48687fce-4cc0-11e8-8c2a-a8da32389878.png">

this take up too much vertical space and is not that interesting.
please put the output into a foldout, same style as the spark progress.

PR #7325, opened 2018-05-08
#7274 foldout for classpath add mvn

MariuszJurowicz:


scottdraves:
Cool thanks, this is good.  In addition can you please:
1) Replace the label "Output" with "Added JARs"
2) When closed, it should show one line of the contents, with an ellipsis at the end if it's truncated.
3) Download progress should not appear in the foldout, that should only appear at the end, when it lists all the jars.
4) Remove the text "Downloaded: " from the progress.  This maybe better done by @jaroslawmalekcodete in the kernel. This should avoid wrapping the line in most circumstances:
screen shot 2018-05-08 at 5 39 35 pm

Issue #7243, opened 2018-04-26
test interrupting kernel in %%groovy magic


scottdraves:

EfimovVladimir:
When I interrupt kernel for %%groovy, what result I should expect?

Issue #7354, opened 2018-05-15
Wrong Python behavior after interrupting %%groovy magic


EfimovVladimir:
If we interrupt one cell the next cell will display result from last cell
![image](https://user-images.githubusercontent.com/17744188/40050459-3bead270-5840-11e8-8aec-7f149407ed2b.png)

If we try to interrupt infinite loop the cell will hangs
![image](https://user-images.githubusercontent.com/17744188/40108774-9860d300-5904-11e8-9fc5-0b35a546613b.png)

PR #7355, opened 2018-05-15
test interrupting kernel in %%groovy magic

EfimovVladimir:
#7243

### DIFF
--- a/test/ipynb/python/KernelMagicsTest.ipynb
+++ b/test/ipynb/python/KernelMagicsTest.ipynb
@@ -77,6 +77,19 @@
    ]
   },
   {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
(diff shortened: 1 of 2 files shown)


test/ipynb/python/KernelMagicsTest.ipynb
```diff
@@ -76,6 +76,19 @@
...
   },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%%groovy\n",
+    "i = 0;\n",
+    "while(i < 1000) {\n",
```
scottdraves: how about an infinite loop, `while(true) {sleep(100)}`?

EfimovVladimir: done.
btw the interrupting doesn't work  #7354

Issue #6805, opened 2018-02-07
always run pexpect test


scottdraves:
it seems our `jupyter console` test is only run if all the other tests pass.
it should always run.

PR #7356, opened 2018-05-15
#6805: always run pexpect test

wojciechowskim:
Connected with #6805


test/run_tests.py
```diff
@@ -101,6 +101,7 @@ def signal_handler(sgnl, frame):
     result = test_console.test_lsmagic()
 
 if result:
+    test_console.test_lsmagic()
```
scottdraves: the script should fail if the tests fail.
there should be only one call to test_lsmagic.
rename test_lsmagic to test_ipython.

PR #7373, opened 2018-05-17
#7372 fix SparkProgress labels width

MariuszJurowicz:


scottdraves:
i'm seeing ellipses in the boxes and they are changing size too.
use `local[100]` as the Master URL to reproduce this:

screen shot 2018-05-17 at 8 55 59 pm

if you open it then it looks correct:
screen shot 2018-05-17 at 9 01 16 pm

Issue #6903, opened 2018-02-27
java/groovy APIs for external widgets


scottdraves:
@maartenbreddels suggests this and it sounds like a good idea.
eg: ipyleaflet, pythreejs, ipyvolume.

maartenbreddels:
@SylvainCorlay @jasongrout @vidartf
It would be nice to see how this works to get a jleaflet/jvolume, and from there maybe vidar can see how to autogenerate code from pythreejs for jvm languages.

@scottdraves 
I'm assuming one only needs to target 1 jvm language (Java OR Scala OR Clojure) and get the other for free right?

scottdraves:
If we do java then we get them all, but scala has its own types so it can be improved by having a wrapper.  see https://github.com/twosigma/beakerx/tree/master/kernel/scala/src/main/scala/com/twosigma/beakerx/scala/chart

scottdraves:
Maarten volunteered to allow a subdirectory in the ipyvolume repository with the API.
then with a https://github.com/twosigma/beakerx/blob/master/jitpack.yml
file at the top level, the jars would get published in a way that would be easy to import to beakerx with the `%classpath add mvn` magic.

vidartf:
pythreejs is using autogeneration for code, so that might be possible.

Another project that might be relevant is https://github.com/vidartf/widget-gen/. It allows for generating widget stubs based on the serialization format that would be used. It doesn't translate any logic, but if your kernel side widgets are only for data input it should be easy to sync all the kernel side definitions.

vidartf:
PS: widget-gen is able to parse python widget definitions with traits as a source for auto-generation.

maartenbreddels:
I think that would be a good idea, that saves me trouble of maintaining the boilerplate part.

vidartf:
@scottdraves Do you think you could hash out some sort of template-like structure for a widget definition in java/scala? If i have such a starting point, I can incorporate support for it in widget-gen.

scottdraves:
Wow cool widget-gen will help a lot.
Yea let's do it :)

scottdraves:
progress
![screen-shot-2018-05-18-at-9 54 19-am](https://user-images.githubusercontent.com/963093/40310971-7a6c429a-5cdc-11e8-9d14-9ac073a151e5.png)

maartenbreddels:
Progress indeed! Nice to see, willl try this out really soon.

scottdraves:
pythreejs starting to work now too:
[...]

PR #7395, opened 2018-05-22
#6903 add doc for ipyvolume

scottdraves:


scottdraves:
this is missing `conda install -c conda-forge ipyvolume`

PR #7411, opened 2018-05-23
#7377: add link test to table display

wojciechowskim:
#7377


scottdraves:
the test should click on the link and confirm the web page opens.

PR #7422, opened 2018-05-24
jarek/7405: fix image example

jaroslawmalekcodete:


scottdraves:
how hard is it to do this without guava?

Issue #7326, opened 2018-05-09
improve spark status widget


scottdraves:
<img width="376" alt="screen shot 2018-05-08 at 10 42 18 pm" src="https://user-images.githubusercontent.com/963093/39792946-2b7bab96-5311-11e8-84b4-206b62731f6c.png">

* add a status indicator, empty gray for not connected, green bkg for connected, with text label being the master URL.
* replace the word "disconnect" with just the "x", making the button smaller.  move it to the right, with some margin from the other controls. 
* ideally would show a few cluster metrics (active tasks, storage memory, dead), updated every 1000ms if visible.
https://spark.apache.org/docs/latest/monitoring.html
has a REST API... whatever :4040/executors/ in the Spark UI uses.
<img width="1178" alt="screen shot 2018-05-14 at 11 06 01 pm" src="https://user-images.githubusercontent.com/963093/40034561-701bd7d0-57cb-11e8-8b6a-8846781706d9.png">
* put a box around it.

PR #7423, opened 2018-05-24
#7326 improve spark status widget

MariuszJurowicz:


js/notebook/src/SparkUI.ts
```diff
@@ -110,6 +165,11 @@ class SparkUIView extends widgets.VBoxView {
       });
     });
   }
+  
+  despose() {
```
jpallas: typo?

scottdraves: thanks https://github.com/twosigma/beakerx/issues/7425

Issue #7353, opened 2018-05-15
spark config improvement


scottdraves:
when the master URL starts with "local", the other 2 fields should be grayed out (their value will be ignored anyway).

PR #7436, opened 2018-05-28
#7353 Spark config improvements

MariuszJurowicz:


scottdraves:
cool, this is the right idea but a couple of small problems:
1) it should gray out for anything starting with `local`. See https://spark.apache.org/docs/latest/submitting-applications.html#master-urls
2) it should take effect immediately as i type.

PR #7455, opened 2018-05-30
[WIP] jarek/7393: add preview for spark dataset

jaroslawmalekcodete:


scottdraves:
awesome.  some improvements:

* button label "Preview 10 Rows"
* should load 10 not 1000
* after clicking the button, when showing the preview, should show a warning, same place/style as the "Note: table is too big to display...." message but it should read "Note: showing a preview of a non-materialized Spark RDD"

remove the button for returning to the shape-only table.

jpallas (commented):
Suggest explicit return types for `toJavaMap` and `toJavaCollection`

scottdraves:
Thank you Joe, what is the advantage of adding the types?

jpallas:
The [Scala style guide](https://docs.scala-lang.org/style/types.html#inference) doesn't demand it, since these methods are not part of the public interface.  But their entire purpose is to map from Scala types to Java types, so being explicit about the return types seems like a good idea.  The explicit type on the final `val` in those methods would then be redundant, but I'm not sure if it helps or hinders readability.

Some type aliases might help the overall readability.  Maybe something like
```scala
type TableMap = util.Map[String, Object]
type TableArray = Array[TableMap]
type TableCollection = util.Collection[TableMap]
```

scottdraves:
plus https://github.com/twosigma/beakerx/issues/7464

scottdraves:
👍

Issue #7456, opened 2018-05-30
Support Table Of Content Extension


LeeTZ:
from https://github.com/ipython-contrib/jupyter_contrib_nbextensions/tree/master/src/jupyter_contrib_nbextensions/nbextensions/toc2

PR #7457, opened 2018-05-30
#7456: add support for table of content

LeeTZ:


scottdraves:
https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/master/COPYING.rst
means we should add that line to our README, and explicitly mention this module.
ditto for the init cells, actually.  there is already a line for IPython, but this is IPython-contrib.

PR #7484, opened 2018-06-06
jarek/7470: handle failure in spark getOrCreate method

jaroslawmalekcodete:


scottdraves:
this works great but the error section is missing its title.
screen shot 2018-06-06 at 8 40 56 am

on the console it says:
```
2018-06-06 08:40:20:186 -0400 [Thread-0] ERROR SparkContext - Error initializing SparkContext.
org.apache.spark.SparkException: Could not parse Master URL: 'nothing://foo'
        at org.apache.spark.SparkContext$.org$apache$spark$SparkContext$$createTaskScheduler(SparkContext.scala:2767)
...
```
no sense showing them an confusing stack trace when the real info is hidden away.
unfortunately i can't see any reasonable way of capturing this info that spark logs.
so we should just have one line of text "Error creating SparkSession, see the console log for more explanation".  put this to the right of the "Start" button instead of the bottom.

Issue #7332, opened 2018-05-09
make spark profiles


scottdraves:
allow groups of spark UI properties to be named and saved and loaded.
followup to https://github.com/twosigma/beakerx/issues/7331

scottdraves:
where profile is an editable combo box.
<img width="714" alt="screen shot 2018-06-12 at 11 44 38 pm new" src="https://user-images.githubusercontent.com/963093/41329636-d46d4a06-6e9c-11e8-9ce0-f4aa2903ab72.png">
they get stored in the beakerx.json too.

PR #7528, opened 2018-06-14
#7332 SparkUI - saving and loading configuration profiles

lmitusinski:


scottdraves:
Cool this works -- how about these additions:

* This way of storing a dict in beakerx.json:
```
"spark_options": {
      "prof A": {
```
for the profiles, use an array of name/value objects, like we do for jvm properties:
```
{
    "beakerx": {
        "jvm_options": {
            "other": [],
            "properties": [
                {
                    "name": "a",
                    "value": "b"
                }
            ]
```

* If profile is not used, still save form values into beakerx.json, in its own section.

* Fields have different typeface size, "Profile" is bigger.
screen shot 2018-06-14 at 10 28 02 am

scottdraves:
Thanks that's great, after playing with it I think

* a [dropdown](https://github.com/jupyter-widgets/ipywidgets/blob/b58d616484565165a1654358b1561dc7245862da/docs/source/examples/Widget%20List.ipynb) will work better than a editable combo box.  That will probably fix the font issue, the alignment of the label, and be easier to use.

* But the user still needs a place to type a name for when they want to create a new profile.  So add a new button to the right of "save".  label is "+", it creates a new profile.  It should popup a modal dialog that asks what to name it, with OK/cancel buttons.

* Tooltip for "+": "Create new profile", tooltip for "x": "Delete this profile".  tooltip for "save": "save profile"

* the name of the current profile should be saved as well, so if i use the profile dropdown and select one, and click "start", then rerun jupyter and open the spark UI, the same profile should be named and loaded.

* therefor the beakerx.json format should be adjusted.  use this:
```
"spark_options": {
  "current_profile": "X"
  "profiles": [ {"name": "X", "config": {...}},
                {"name": "Y", "config": {...}}
              ]
}
```
* save button should be disabled if there is nothing to save.
* ~~"Default" should just be blank.~~

Issue #7550, opened 2018-06-20
jagged spark status indicator heights


scottdraves:
in notebook in the cell output:
<img width="202" alt="screen shot 2018-06-19 at 10 09 11 pm" src="https://user-images.githubusercontent.com/963093/41633567-8eaa3a08-740d-11e8-8cdc-c038e2668913.png">

in the toolbar:
<img width="200" alt="screen shot 2018-06-19 at 10 09 27 pm" src="https://user-images.githubusercontent.com/963093/41633572-94c137d4-740d-11e8-95f3-94506fa4a7b7.png">

also check jupyter lab.

scottdraves:
how about matching this height:
<img width="507" alt="screen shot 2018-06-19 at 10 22 43 pm" src="https://user-images.githubusercontent.com/963093/41634070-efa4689a-740f-11e8-9145-04f8953464bf.png">

PR #7566, opened 2018-06-22
#7550 jagged spark status indicator heights

piorek:


scottdraves:
looks like they are a little too tall to fit in the Lab toolbar:
screen shot 2018-06-22 at 11 03 35 am

PR #7569, opened 2018-06-22
Display images

ctrueden:
As [discussed on Gitter](https://gitter.im/twosigma/beakerx?at=5b2d40adaeb8fa0c0737ca47), it is nice for the `Image(...)` MIME function to accept `BufferedImage` and other `RenderedImage` objects.

This PR splits out the various `ImageIO` uses that convert between images and `byte[]` into a common `Images` utility class. It includes a simple unit test for that class. It also updates the `Image` MIME function to special case `RenderedImage` objects. Finally, we register a displayer for `RenderedImage` that computes a base64-encoded PNG for `RenderedImage` outputs by default!


scottdraves:
Thanks this looks great, will test shortly.

scottdraves:
🐬

Issue #7327, opened 2018-05-09
time/date format for int64


scottdraves:
<img width="322" alt="screen shot 2018-05-08 at 11 03 06 pm" src="https://user-images.githubusercontent.com/963093/39793653-69bfc3a8-5314-11e8-9c80-b6714b57d5a4.png">
when doc/scala/Flint.ipynb reads the CSV file.

PR #7588, opened 2018-06-27
#7327 time/date format for int64

piorek:


js/notebook/src/tableDisplay/dataGrid/DataFormatter.ts
```diff
@@ -226,15 +226,19 @@ export class DataFormatter {
...
+      format = formatForTimes.format;
+      valueModifier = formatForTimes.valueModifier;
+    }
 
     if (_.isObject(config.value) && config.value.type === 'Date') {
       return formatTimestamp(config.value.timestamp, this.timeZone, format);
     }
 
-    let milli = config.value * 1000;
+    let milli = config.value * valueModifier;
```
scottdraves: arithmetic with a 64-bit value must be done with big.js

Issue #7575, opened 2018-06-25
implement autotranslation for python


scottdraves:
followup to https://github.com/twosigma/beakerx/pull/7542

PR #7606, opened 2018-06-29
#7575 autotranslation for python

lmitusinski:


beakerx/beakerx/autotranslation.py
```diff
@@ -0,0 +1,15 @@
+from ipykernel.zmqshell import ZMQInteractiveShell
```
scottdraves, line 15: banner

scottdraves:
i am getting only string results.  it's missing the JSON decode or something.
try doc/groovy/GeneralAutotranslation.ipynb

Issue #7625, opened 2018-07-02
python nanoplot problems:


scottdraves:
```
from beakerx import *
p = NanoPlot()
p.add(Points(x=[10000000],y=[5]))
p.add(Points(x=[20000000],y=[5]))
```
another bug, this should work:
```
p = NanoPlot()
p.add(Points(y=[5]))
```

PR #7663, opened 2018-07-06
 #7625 fix python NanoPlot problems

piorek:


piorek:
@LeeTZ tested and fixed this. It should be ok now

Issue #7338, opened 2018-05-10
tables: add option to HTML format for fixed width columns


scottdraves:
currently the html format renders every cell to the DOM to measure its size, and then rerenders them all to an SVG to actually get the pixels.    it would be a lot faster to skip the DOM step.

add an optional parameter to the HTML format (similar to the digits for the floating point formats):
`TableDisplayStringFormat.getHTMLFormat(width:200)`

scottdraves:
should apply to image format as well

PR #7667, opened 2018-07-09
#7338 support fixed width for image and HTML columns

MariuszJurowicz:


piorek (approved):
Frontend part is OK.
Python and Java too as far as I can tell, but I'm not an expert.

PR #7671, opened 2018-07-10
jarek/7653: support autotranslation for clojure

jaroslawmalekcodete:


lmitusinski (changes requested):
It would be better to place the serializers in a dedicated package, like in the Scala:
"com.twosigma.beakerx.clojure.serializers" instead of "com.twosigma.beakerx.clojure.kernel"

lmitusinski (approved):

Issue #7610, opened 2018-06-30
python autotranslation of pandas dataframe


scottdraves:
```
%%python
from beakerx import beakerx
import pandas as pd
beakerx.bar = pd.read_csv('../resources/data/interest-rates.csv')
```
fails with
```
TypeError: Object of type 'DataFrame' is not JSON serializable
```

scottdraves:
the value appears in groovy now but as a dictionary it still should be converted to a proper TableDisplay class:
![screen shot 2018-07-05 at 9 48 36 am](https://user-images.githubusercontent.com/963093/42326966-9d00b906-8038-11e8-974e-62fb4643c1fb.png)

Issue #7645, opened 2018-07-04
TableDisplay corrupts data in groovy


altavir:
Conversation of map to TableDisplay corrupts map itself instead of creating a copy:

```Groovy
def b = ["a" : 2, "b" : 3, "c" : 4]
println(b)
println(b.a)
```
[a:2, b:3, c:4]
2

```Groovy
println(b)
println(b.a)
```

[type:TableDisplay, columnNames:[Key, Value], values:[[a, 2], [b, 3], [c, 4]], subtype:Dictionary]
null

scottdraves:
thanks for the report but i am not able to reproduce, even if i remove the "def", which allows the 2nd cell to run.  what am i missing?

altavir:
Sorry, it seems to I forgot to restart the notebook and forgot that  Groovy is copying links instead of values. The problem has something to do with autotranslation.
Try `beakerx.b = ["a" : 2, "b" : 3, "c" : 4]` and then `beakerx.b` in another cell. Second one is broken.

altavir:
Correction even `beakerx.a = a; beakerx.a` is broken.

scottdraves:
yes this is a deficiency, essentially values we get back from autotranslation are not converted to the right class, they are left in JSON.  Here's another case of it: https://github.com/twosigma/beakerx/issues/7610
Your case demonstrates it more clearly, actually.

PR #7673, opened 2018-07-10
#7610 #7625 fixed deserialization of table objects in autotranslation

lmitusinski:


jaroslawmalekcodete (approved):
works fine
one remark: we had a discussion with @lmitusinski  and we agreed that we should reconsider the refactoring of  serializers and deserializers,
we have a problem with understanding theses classes e.g BasicObjectSerializer.class and we think It would be good to talk about it in the near future.

lmitusinski:
This should be connected with issues #7610 and #7645, commit name mistakenly refers to nano plot issue.

PR #7674, opened 2018-07-10
java/7652: support autotranslation for java

jaroslawmalekcodete:


kernel/base/src/main/java/com/twosigma/beakerx/NamespaceClient.java
```diff
@@ -31,6 +31,8 @@
 
 public class NamespaceClient implements BeakerXClient {
 
+  public static final String NAMESPACE_CLIENT = NamespaceClient.class.getSimpleName() + ".getBeaker()";
```
LeeTZ: Better to expose "getBeakerX" instead of Beaker

LeeTZ (changes requested):
Better to expose "getBeakerX" instead of getBeaker

kernel/base/src/main/java/com/twosigma/beakerx/NamespaceClient.java
jaroslawmalekcodete: done

lmitusinski (approved):

LeeTZ (approved):

Issue #7654, opened 2018-07-04
support autotranslation for Kotlin


scottdraves:
idk even know what the syntax should be...

altavir:
Like that:
```kotlin
//set
beakerx["a"] = a
//push
beakerx.push("a", myObject)
//get
val a = beakerx["a"] as MyType
// get №2
val a = beakerx.pull<MyType>("a")
```
I suggest implementing both notations. Also it would be good to have custom serialization functions.

I can contribute this one if you make beakerx object somehow available for kotlin runtime.

LeeTZ:
Hi @altavir we added an initial version for Kotlin autotranslation. beakerx object is available in kotlin runtime, with get/set implemented. Haven't gone that far to custom serialization functions yet.
How do you see the importance of having push/pull? You are very welcome to test and contribute.

altavir:
I will test once stable release is out and will try to contribute if there is need. Sadly, I do not currently have time to play with development version of beakerx (last time I've tried, it took a lot of time)
On second thought, `push` is definitely not needed since there is `set` with the same functionality. `Pull` also is just alternative syntax. It is probably not needed, Anyone could add it later with extension function:
```kotlin
inline fun <reified T> Beakerx.get(key: String) = this[key] as T
```

~I am not sure, what you mean by custom serialization though.~

altavir:
Oops. Sorry, It was from my post. Now I need to remember what I meant by this :) It was a busy week.

PR #7681, opened 2018-07-13
#7654 support for autotranslation in Kotlin

lmitusinski:


jaroslawmalekcodete (changes requested):
It looks like autotranslation doesn't work in kotlin kernel
kotlinautotranslationerror

jaroslawmalekcodete (approved):

Issue #7461, opened 2018-05-31
easyform in outputcontainer


scottdraves:
this should make a form, not a table:
```
f = new EasyForm("Test Form")
f.addTextField("sometext")
o = new OutputContainer()
o.addItem(f, "Easy Form")
o
```

LeeTZ:
This is still buggy in tabbed output containers:


```
TableDisplay t = new TableDisplay(table);
def l = new TabbedOutputContainerLayoutManager()
l.setBorderDisplayed(false)
def o = new OutputContainer()
o.setLayoutManager(l)
o.addItem(plot1, "Scatter with History")
o.addItem(plot2, "Short Term")
o.addItem(plot3, "Long Term")
o.addItem(t, "1990/01")
f = new EasyForm("Test Form")
f.addTextField("sometext")
f
o.addItem(f, "easyform")
o 
```
![screenshot from 2018-07-17 17-26-47](https://user-images.githubusercontent.com/3425710/42846360-b297eb74-89e6-11e8-93ef-bda7e22e3532.png)

The easy form is added to every tab.

PR #7691, opened 2018-07-16
#7461 fix for displaying EasyForm in OutputContainer for JVM kernels

lmitusinski:


kernel/base/src/main/java/com/twosigma/beakerx/jvm/object/OutputContainerLayoutManager.java
```diff
@@ -54,6 +55,9 @@ public void setBorderDisplayed(boolean borderDisplayed) {
     if (item == null) {
       return handleNull();
     }
+    if (item instanceof EasyForm) {
```
jaroslawmalekcodete: I would suggest create an interface "WidgetItem" with getAsWIdget() method, 
then  EasyForm and Widget should implement the interface.
This will allow us to get rid of "if statement" for Widget and EasyForm.

jaroslawmalekcodete (approved):

PR #7694, opened 2018-07-16
jarek/7481: repr of scala unit object as MIMEContainer.HIDDEN

jaroslawmalekcodete:


lmitusinski (approved):

kernel/scala/src/main/scala/com/twosigma/beakerx/scala/evaluator/ScalaEvaluatorGlue.scala
```diff
@@ -139,6 +144,11 @@ class ScalaEvaluatorGlue(val cl: ClassLoader, var cp: String, val replClassdir:
     either
   }
 
+  private def isUnitObject() = {
+    val print = interpreter.lastRequest.lineRep.call("$print").toString.trim
+    print.equals("")
+  }
```
jpallas: The method name `isUnitObject` doesn't match what it actually does.  Maybe `isPrintStringEmpty`?  (and then use `isEmpty` instead of `equals("")`).  Or just compare the value to `()` and ignore the interpreter's notion of how things should be printed.

jaroslawmalekcodete: by isUnitObject I want to express whether the interpreter returns Unit object or not.
I used ```val print = interpreter.lastRequest.lineRep.call (" $ print "). toString.trim 
print.equals ("") ``` because I have not found a better way to do this.
The question is whether my solution meets the requirements and doesn't provide errors.
There is a test that checks if Unit object is represented by MIMEContainer.HIDDEN https://github.com/twosigma/beakerx/pull/7694/files#diff-7a177fbeb8a50f57cde772923fc6e1ccR60 
if you know use cases which we should handle or you see problems with this solution, please let us know, your input is welcome.

jpallas: I just did some exploration and the behavior of the interpreter is surprising.  Calling `$result` when the input is `"()"` returns `null` instead of a boxed `Unit`, so looking at the result value alone is not a solution.  I am persuaded that it makes sense to use `$print` to decide whether to hide the result.  That said, I still think the name `isUnitObject` is misleading, and would suggest changing it to `shouldHideResult`.  For a specific example, the input `"(): Any"` gives a result that **is** a unit object (with a wider type), but the interpreter does **not** suppress the output in that case.

Issue #5900, opened 2017-08-31
add publish menu item to plots and tables


scottdraves:
in plots, on the context menu in addition to "save as svg", png, and high dpi png, add "Publish..."
which would publish a notebook with just this output cell (like from the file menu) (after a confirmation dialog that says "publish just this plot to ..."

ditto for tables.

scottdraves:
This would be conditional on the "Show Publish UI" UI option.

PR #7716, opened 2018-07-24
#5900 add publish menu item to plots and tables

piorek:


piorek:
TODO:
- [x] Storing token in beakerx.json is not working
- [x] publish toolbar button in lab is missing passing token from dialog (looks like it wasnt fixed for lab after github disalowed anonymous gists)

Also publishing won't fully work in lab until https://github.com/twosigma/beakerx/issues/6893 is fixed

PR #7732, opened 2018-07-30
fix doc and test for Spark Scala

EfimovVladimir:


doc/scala/Spark.ipynb
```diff
@@ -44,11 +44,12 @@
...
    "outputs": [],
    "source": [
+    "import scala.math.random\n",
     "val NUM_SAMPLES = 10000000\n",
     "\n",
     "val count = spark.sparkContext.parallelize(1 to NUM_SAMPLES).map{i =>\n",
-    "  val x = Math.random()\n",
-    "  val y = Math.random()\n",
+    "  val x = random\n",
+    "  val y = random\n",
```
jpallas: Should keep the parentheses on the call to `random()`.   [Scala's convention](https://docs.scala-lang.org/style/method-invocation.html#arity-0) is that side-effecting nullary functions get called with an explicit empty parameter list.

EfimovVladimir: thanks

PR #7744, opened 2018-08-09
jarek/7428 add bkg and interrupt buttons to spark progress

jaroslawmalekcodete:


EfimovVladimir:
tested, all is good :+1:

Issue #7751, opened 2018-08-13
Can't use the custom GradientColor in Scala


EfimovVladimir:
It works fine on Groovy
![image](https://user-images.githubusercontent.com/17744188/44031132-b1079a36-9f0b-11e8-8b61-55ff3e9c1f41.png)

But we get the error on Scala
![image](https://user-images.githubusercontent.com/17744188/44031509-e82a9864-9f0c-11e8-9b17-a5de9856ac16.png)

PR #7752, opened 2018-08-14
#7751 add Scala GradientColor

jpallas:
This addresses #7751 by adding a Scala version of `GradientColor` (including removing the import of the Java version).  A companion object with an apply method is included for convenience.


LeeTZ:
Looks great, thanks

PR #7761, opened 2018-08-19
Fix docker build

JD557:
As far as I can tell, the current docker build is not working. This PR is an attempt to fix that and improve the Dockerfile.

Some of the changes are based on my personal preference. I'm open to feedback about those choices.

I tried to split this in multiple small commits to make the review easier (I'm not sure if you prefer a single large commit, but I can squash the changes later if you prefer).

Nevertheless, here's an overview of the main changes:

* **Add the `js/lab` directory as a build dep.:** `yarn install` was not being called for `js/lab`. (If you had already called `yarn install` during develop, you would not notice this);
* **Merge base docker image into the main image:** The [build cache](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache) should always cache those initial commands, so I think that the base image only makes the build process more complex without any gains;
* **Split install operations:** Use multiple `RUN` commands instead of a single one to avoid trashing the build cache if a single dependency is changed;
* **Add required files to the docker file:** Adds only the required files to the docker image. They are also split in a way that avoids trashing the build cache when only the documentation files are changed.


jpallas:
Also need to update the docker part of the Architecture and Code Overview in `README.md`.

LeeTZ:
Thanks @JD557 the new script looks great. I think the only remaining work is documenting

PR #7773, opened 2018-08-27
jarek/7718: accessible url arguments from notebook

jaroslawmalekcodete:


LeeTZ:
Looks good. let's add python support for that as well.

PR #7774, opened 2018-08-29
jarek/7472: ensure deterministic classpath order based on maven 'dependency:build-classpath' goal

jaroslawmalekcodete:


LeeTZ:
@jpallas Would you mind test it to see if it addresses the issue you mentioned? Thanks

jpallas:
I confirmed that the classpath generated is stable and that it seems to work for several additional Scala libraries that I spot-checked (breeze, vegas, tensorflow_scala).

kernel/base/src/test/resources/PomTemplateMagicCommand.xml
```diff
@@ -1,4 +1,3 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
jpallas: Unintentional deletion?

jaroslawmalekcodete: Unintentional deletion reverted

Issue #7743, opened 2018-08-09
Histogram Groovy has wrong scale for default mode


EfimovVladimir:
See http://localhost:8888/notebooks/doc/groovy/Histogram.ipynb (cell with Default is Overlap)
Setup initWidth == initHeight. After that we expect to see square area of chart.

![image](https://user-images.githubusercontent.com/17744188/43890791-61d29670-9bd0-11e8-8154-32212bbce4ae.png)

piorek:
@LeeTZ if you updated documentation as we discussed please close this one

PR #7783, opened 2018-09-05
#7743 Histogram Groovy has wrong scale for default mode

piorek:
@LeeTZ As we discussed 
- please update docs for initHeight, initWidth
- I made a fix so that plot/histogram behave same with showLegend: true|false

### DIFF
--- a/js/notebook/src/plot/PlotSize.ts
+++ b/js/notebook/src/plot/PlotSize.ts
@@ -55,7 +55,26 @@ export default class PlotSize {
 
     legendWidth = legendWidth ? legendWidth + this.scope.layout.legendMargin + 2 : 0;
 
-    return isLegendPlacedHorizontaly ? plotWidth - legendWidth : plotWidth;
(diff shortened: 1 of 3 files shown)


js/notebook/src/plot/PlotSize.ts
```diff
@@ -55,7 +55,26 @@ export default class PlotSize {
...
+
+  getPlotWithLegendHeight() {
+    const containerHeight = this.scope.jqcontainer.parents('.output_subarea').height();
+    const plotHeight = containerHeight && containerHeight < this.scope.layout.plotSize.height
+      ? containerHeight
+      : this.scope.layout.plotSize.height;
+    const legendPosition = this.scope.stdmodel.legendPosition.position;
+    // Logic based on updateLegendPosition method
+    const isLegendPlacedHorizontaly = (
+      (["LEFT", "RIGTH"].indexOf(legendPosition) !== -1)
```
LeeTZ: typo

PR #7797, opened 2018-09-18
jarek/7780: fix %classpath add mvn: NoSuchFileException under Win

jaroslawmalekcodete:


kernel/base/src/main/java/com/twosigma/beakerx/kernel/magic/command/FileServiceImpl.java
```diff
@@ -0,0 +1,14 @@
+package com.twosigma.beakerx.kernel.magic.command;
```
LeeTZ, line 16: add banners, the same below.

LeeTZ (changes requested):

kernel/base/src/main/java/com/twosigma/beakerx/kernel/magic/command/FileServiceImpl.java
jaroslawmalekcodete: done

LeeTZ (approved):

PR #7800, opened 2018-09-20
jarek/7718: url arg accessible from notebook python

jaroslawmalekcodete:


beakerx/beakerx/handlers.py
```diff
@@ -12,34 +12,40 @@
...
     def post(self):
 
-        def handle_response(response):
-            self.finish(response.body)
-
         data = tornado.escape.json_decode(self.request.body)
         content = json.dumps(data)
         params = json.loads(content)
+
+        type = params['type']
```
LeeTZ, line 41: This would raise a KeyError that breaks other post methods, like the spark interrupt button. Use dict.get instead.

PR #7828, opened 2018-10-16
Upgrade to Smile 1.5.2 for named variables

benmccann:
The code is a little shorter, but more importantly the regression output now uses the variable names instead of "Var 1", "Var 2", etc.


benmccann:
@lwhite1 @haifengl you guys might like this one

LeeTZ:
Thanks @benmccann !

Issue #7506, opened 2018-06-08
Timezone not showing correctly


aqi-jng:
```
index = ['2015-12-31T08:35:00-06:00',
        '2015-12-31T08:35:00-06:00',
        '2015-12-31T08:35:00-06:00',
        '2015-12-31T08:35:00-06:00',
        '2015-12-31T08:35:00-06:00',
        '2015-12-31T08:35:00-06:00',
        '2015-12-31T08:35:00-06:00']

series = pd.Series(index=index, data=0.0)
series.index = pd.to_datetime(series.index, utc=True).tz_convert('America/Chicago')
series.index.name = 'tm'

series
series.to_frame()
```

scottdraves:
Hi thank you for the report, will look into it.

piorek:
@LeeTZ close this issue please

PR #7841, opened 2018-10-25
#7506 Timezone not showing correctly

piorek:


beakerx/beakerx/tabledisplay/tableitems.py
```diff
@@ -43,9 +44,9 @@ class ColumnType(Enum):
 class DateType:
     type = "Date"
 
-    def __init__(self, value):
-        self.timestamp = parse(str(value)).timestamp() * 1000
-
+    def __init__(self, value, tz=None):
+        self.timestamp = parse(str(value)).replace(tzinfo=timezone.utc).timestamp() * 1000
```
piorek, line 48: @LeeTZ parsed date was treated as local time so timestamp was shifted.  Maybe it can be done better. Please take a look.

Issue #7315, opened 2018-05-04
other plot types should limit rows


scottdraves:
histograms, heatmaps, and treemaps should limit data on the server side.
right now if you try to compute a histogram of a large dataframe, it basically hangs and gives an IOPub error. we already have limits for tables and plots, should similarly protect all the plot types.
Please limit at 100k rows.  For histograms, like tables, it should truncate the data and still draw the plot, just with a preview.  the others should just fail.

PR #7845, opened 2018-11-02
#7315 other plot types should limit rows

piorek:


LeeTZ:
To mimic larger datasets, for heatmap you may try this by increasing the maxdepth:
```
from random import randint
maxdepth = 500
data_long = [[randint(1, 100) for x in range(maxdepth)] for y in range(maxdepth)]
```
For treemap, in our doc Treemap.ipynb, incresing the range to generate object 'big_node' would provide you a large number of nodes.

PR #7878, opened 2018-12-03
add a new option to make beakerx installed per-user

clouds56:


ildipo:
Repository was refactored

PR #7886, opened 2018-12-11
jarek/7276: fix failing tests in docker

jaroslawmalekcodete:


beakerx/setup.py
```diff
@@ -101,7 +101,7 @@
         get_data_files(os.path.join('beaker'))
     )],
     install_requires    = [
-        'notebook >=4.4.0',
+        'notebook ==5.6.0',
```
LeeTZ, line 104: Could we get around these tests without locking it to 5.6?

jaroslawmalekcodete: yes, we can but I fixed version because I had the following problem
dockererror
when I run beakerx docker image

jaroslawmalekcodete: thing to consider:
Fixed version of notebook sounds more reliable because when we have "notebook >=4.4.0" we don't really know which version of notebook we have in our docker image and we don't control the  moment when we switch between newer version.
Maybe we should switched to fixed version approach and create notebook version update issue every time when new notebook is released, then we have chance to check if our docker image works fine with newer version of notebook.

PR #7889, opened 2018-12-12
jarek/7884: setting user object before reducing nodes

jaroslawmalekcodete:


jaroslawmalekcodete:
Closed -> improved version here https://github.com/twosigma/beakerx/pull/7897

Issue #7832, opened 2018-10-18
Context menus not showing in mybinder


rdhyee:
I'm learning about how to use beakerx, specifically, how to use the Python bindings to add context menus a la <https://github.com/twosigma/beakerx/blob/master/doc/python/TableAPI.ipynb>. When I run this notebook on mybinder.org using the latest image (<https://mybinder.org/v2/gh/twosigma/beakerx/1.2.0?filepath=doc/python/TableAPI.ipynb>), the context menu doesn't show up. 

![image](https://user-images.githubusercontent.com/153266/47168724-74303f80-d2b6-11e8-8901-800d812927c1.png)

For a while, I was thinking that there is a bug in beakerx since I found that the context menus work in an earlier mybinder build of a previous version of beakerx (<https://mybinder.org/v2/gh/twosigma/beakerx/0.12.0?filepath=doc/python/TableAPI.ipynb>). 

![image](https://user-images.githubusercontent.com/153266/47168862-dd17b780-d2b6-11e8-819d-aa233a6d2c0f.png)

However, I then began to suspect that it's not a problem in beakerx per se but in how beakerx interacts with the other parts of the mybinder environment. Some evidence for this hypothesis: the context menus don't work in <https://mybinder.org/v2/gh/rdhyee/beakerx/0.12.0> \-- even though my fork of beakerx is identical to the original (<https://github.com/twosigma/beakerx/compare/0.12.0...rdhyee:0.12.0>). The key difference, I think, is that I built <https://mybinder.org/v2/gh/rdhyee/beakerx/0.12.0> on Oct 11, 2018 whereas <https://mybinder.org/v2/gh/twosigma/beakerx/0.12.0> was built a while ago (but I don't know when).

When I brought this issue up on the [binder gitter channel](https://gitter.im/jupyterhub/binder), Tim Head (@betatim) wrote:

> A more likely culprit in my experience is that something that a repo pulls in during the build has changed since the last build. the most common version of this problem is not pinning a dependency so when the same commit is rebuilt N months later you get a new version of the dependency and everything breaks

It'd be wonderful to fix this problem because it's handy to demonstrate beakerx fully using mybinder.org.

LeeTZ:
Thanks for the report I could confirm that it is not happening in a local conda build but on mybinder only. I will take a look into that.

PR #7900, opened 2018-12-18
#7832 Context menus not showing in mybinder

piorek:


js/notebook/src/plot/contextMenu/plotContextMenu.ts
```diff
@@ -30,7 +30,7 @@ export default class PlotContextMenu extends BkoContextMenu {
       ...createSaveAsMenuItems(this.scope),
     ];
 
-    new BeakerXApi(`${window.location.origin}/`)
+    new BeakerXApi(`${(Jupyter.notebook_list || Jupyter.notebook).base_url}`)
```
LeeTZ, line 33: Could you explain why this change fixes the menu issue?

piorek: There was uncaught exception caused by 404 while requesting /settings endpoint (needed to see if publishing is enabled).

This one is for plots but is exactly the same case.

PR #7909, opened 2019-01-02
#7908 fix ReferenceError: Jupyter is not defined

piorek:


piorek:
@LeeTZ @jaroslawmalekcodete found that I've broken master. This fixes it. So merge it ASAP.

Issue #7383, opened 2018-05-18
error in JavaScript for Plot (Lab, Python)


EfimovVladimir:
can reproduce on PlotJSAPI.ipynb
code:
```
%%javascript
var plot = new beakerx.Plot({
    title: 'Title',
    xLabel: "Horizontal", 
    yLabel: "Vertical",
    initWidth: 800,
    initHeight: 500
})
plot.display(this);
```
![image](https://user-images.githubusercontent.com/17744188/40246110-3f5021e4-5ad0-11e8-8226-d1784923aef4.png)

PR #7967, opened 2019-02-13
#7383 fix javascript plot api in Lab

piorek:


piorek:
@LeeTZ to use external libs like big.js in lab we have to do "same" tehnique we use in autotranslation examples when using d3

```
%%javascript
require.config({
     paths: {
        'big': '//cdnjs.cloudflare.com/ajax/libs/big.js/5.2.2/big'
     }
 });
require(['big'], function(Big) {
  window.Big = Big;
});
```

PR #8004, opened 2019-03-11
Updated README

erriapo:
Instruction to activate environment did not work before.

```
source activate labx
```
returned the error "activate: No such file or directory"


LeeTZ:
Thanks, this is related with conda version. We will keep it as it is now.

PR #8008, opened 2019-03-14
jarek/7985: fix println in scala and clojure

jaroslawmalekcodete:


EfimovVladimir:
was checked.
:+1:  works good

Issue #7975, opened 2019-02-20
content is placed over the Plot border (javaScript)


EfimovVladimir:
See http://localhost:8888/notebooks/doc/python/PlotJSAPI.ipynb notebook
Try to **move** the content of Plot. We can see that the content is placed over the border of Plot.
![image](https://user-images.githubusercontent.com/17744188/53083232-8abe0000-350f-11e9-8b60-875690175f88.png)

PR #8039, opened 2019-04-17
#7975 fix content is placed over the Plot border (javaScript)

piorek:


configuration.yml
```diff
@@ -7,7 +7,7 @@ dependencies:
   - notebook>=5.7.6
   - tornado>6
   - ipywidgets>=7.0.0
-  - nodejs
+  - nodejs=10.13.0
```
piorek: @LeeTZ had to do this because of #8028 (on clean env that I had its a must)

kernel/sparkex/src/test/java/com/twosigma/beakerx/EnableSparkSupportTest.java
```diff
@@ -115,7 +115,7 @@ public void sparkPreviewDisplayer() throws Exception {
     }
   }
 
-  @Test
+//  @Test
```
piorek: @LeeTZ had to do this because of #8028 without it could not go past 
```(cd beakerx; pip install -e . --verbose)```

configuration.yml
LeeTZ: Fixing isn't a long term solution, we should revisit this later

PR #8061, opened 2019-05-22
[WIP] #8056 separate table display

piorek:


piorek:
@LeeTZ 
This is extracted version of table display javascript/typescript files.
As you can see some files were duplicated

- src/contextMenu/*
- src/plot/*
- src/shared/*
- src/tree/*
and maybe others

What I'm trying to achieve now is move those files to its own folder/package.
I'm trying to use them then as a local package in package.json (never tried that before)

I started with BkoContextMenu.ts
and I probably misconfigured sth because after using it from local package I'm getting loader errors in our typescript tests

PR #8076, opened 2019-06-19
#8056 separate table display

piorek:


EfimovVladimir:
for jupyter lab are
installation errors when we run command `pip install -r requriments.txt --verbose` :
```> beakerx-jupyterlab@1.4.1 build:src /home/vova/beakerx/js/lab
    > rimraf dist && tsc --project src

    node_modules/@jupyter-widgets/base/lib/manager-base.d.ts(197,33): error TS1039: Initializers are not allowed in ambient contexts.
    node_modules/@types/backbone/index.d.ts(389,24): error TS2503: Cannot find namespace '_'.
    node_modules/@types/backbone/index.d.ts(390,24): error TS2503: Cannot find namespace '_'.```

piorek:
@EfimovVladimir should be ok now.

Please use `conda install -y -c conda-forge jupyterlab=0.35` for now. 1.0.0 was released and we are not ready for it yet.

EfimovVladimir:
classic notebook looks good

piorek:
@LeeTZ as we spoke on hangouts:

We have a problem to solve.

According to https://github.com/jupyter-widgets/ipywidgets/pull/2472 there is a _"split"_ in ipywidgets
- to support classic notebook and jupyterlab <1.0 we have to use @jupyter-widgets/base#^1
- to support jupyterlab>=1.0 we have to use @jupyter-widgets/base#^2

the solution mentioned there using ^1 || ^2 is AFAIK only partial because there is ipywidgets conda package

- we have to use ipywidgets<7.5 for classic notebook and old jupyterlab
- we have to use ipywidgets>=7.5 for jupyterlab 1.0

------

I've changed to @jupyter-widgets/base ^1 || ^2 and that part is working fine while using ipywidgets<7.5 but as soon as I try to use it with ipywidgets 7.5 I have a problem.

In many our views like EasyFormView we are using Bootstrap.View stuff like having this.$el but after switching to ipywidgets#7.5 it's not there at runtime at least not here 
https://github.com/twosigma/beakerx/blob/lukasz/8056_separate_table_display/js/notebook/src/EasyForm.ts#L66

maybe it's because of 
- where we are using it, 
- some changes in ipywidgets#7.5,
- some changes in jupyterlab#1.0 itself

maybe @afshin or someone else close to jupyterlab / ipywidgets knows something about this and can tell us where to look or what we can try.

Maybe it related to es target?

afshin:
@jasongrout @sylvaincorlay, is this understanding of the constraints accurate? If we want the table display to continue to support both JupyterLab and also classic (let's assume *older* versions of JupyterLab are not an issue) ... what versions do we need to target?

LeeTZ:
Both classic notebook and lab look good.

jasongrout:
> @jasongrout @SylvainCorlay, is this understanding of the constraints accurate? If we want the table display to continue to support both JupyterLab and also classic (let's assume _older_ versions of JupyterLab are not an issue) ... what versions do we need to target?

Just to follow up on this: 

To support classic and JupyterLab (including older versions of jlab), and older and newer versions of ipywidgets, just use base version `^1 || ^2`.

PR #8079, opened 2019-06-19
Demonstrate charting an arbitrary number of bands

### LINKED ISSUE #8078: Improve README's treasury inversion example
The README highlights a treasury inversion example: https://github.com/twosigma/beakerx#groovy-with-interactive-plotting

The source code for this example is here: https://github.com/twosigma/beakerx/blob/master/doc/groovy/Charting.ipynb

However, this example is not very realistic because it hard codes the dates for the inversions. I'd like to see an example of how you would calculate and display an arbitrary number of inversions or data points. 

The example code does `p1 << l1 << l2 << b1 << b2`. But what if I don't have `b1` and `b2`, but instead have `b1` - `bn`? Is there some way to plot these with a for-loop or similar?

benmccann:
Closes https://github.com/twosigma/beakerx/issues/8078

This makes the example more realistic by plotting an arbitrary number of treasury inversions. It wasn't clear to me initially how to plot bands in a loop.

Better yet would be to calculate the inversions instead of hardcoding them, but I'll leave that for someone else to handle


benmccann:
I've signed the CLA already: https://github.com/twosigma/beakerx/pull/5954#issuecomment-328174492

Issue #8083, opened 2019-06-26
Error: Circular Reference Detected on Table Display


JohnOmernik:
So this may appear to be similar to #7935 is is actually different. I was running a Dataframe into a BeakerX table and saw that I had some int64 columns and assumed they were the issue. They were not. 

However, I had two columns that were of type bool in df.dtypes. When I converted these two to type object, then the dataframe was displayed correctly. Thus what ever caused an issue with int64 probably needs to be looked at for type bool as well. 

This is on Pandas version 0.24.2 and BeakerX 1.4.0

JohnOmernik:
Example Code:
```
import pandas as pd
from beakerx import *
from beakerx.object import beakerx
#Display mode: TableDisplay Widget
beakerx.pandas_display_table()
df = pd.DataFrame(data=np.zeros((5,5)), index=pd.Int64Index([0, 1, 2, 3, 4], dtype='int64'))
df # Displays properly
```
```
df.dtypes
```
0    float64
1    float64
2    float64
3    float64
4    float64
dtype: object
```
df[2] = df[2].astype(bool)
df.dtypes
```
0    float64
1    float64
2       bool
3    float64
4    float64
dtype: object
```
df # Errors out
```

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/opt/conda/lib/python3.7/site-packages/IPython/core/formatters.py in __call__(self, obj)
    916             method = get_real_method(obj, self.print_method)
    917             if method is not None:
--> 918                 method()
    919                 return True
    920 

/opt/conda/lib/python3.7/site-packages/beakerx/runtime.py in f()
    467     def __get__(self, model_instance, model_class):
    468         def f():
--> 469             display_html(TableDisplay(model_instance))
    470 
    471         return f

/opt/conda/lib/python3.7/site-packages/beakerx/tabledisplay/tabledisplay.py in __init__(self, *args, **kwargs)
    211         super(TableDisplay, self).__init__(**kwargs)
    212         self.chart = Table(*args, **kwargs)
--> 213         self.model = self.chart.transform()
    214         self.on_msg(self.handle_msg)
    215         self.details = None

/opt/conda/lib/python3.7/site-packages/beakerx/utils.py in transform(self)
     53 
     54     def transform(self):
---> 55         model = json.dumps(self, cls=ObjectEncoder)
     56         return json.loads(model)
     57 

[...]

PR #8091, opened 2019-07-12
Add numpy.bool_ type to ObjectEncoder

rootty:
Implements:
Error: Circular Reference Detected on Table Display #8083

### DIFF
--- a/beakerx_base/beakerx_base/utils.py
+++ b/beakerx_base/beakerx_base/utils.py
@@ -199,6 +199,8 @@ class ObjectEncoder(json.JSONEncoder):
             return self.default(obj.tolist())
         elif isinstance(obj, np.int64):
             return self.default(obj.item())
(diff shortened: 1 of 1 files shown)


beakerx_base/beakerx_base/utils.py
```diff
@@ -199,6 +199,8 @@ def default(self, obj):
             return self.default(obj.tolist())
         elif isinstance(obj, np.int64):
             return self.default(obj.item())
+        elif isinstance(obj, np.bool_):
```
rootty: @LeeTZ @jaroslawmalekcodete I wonder if it's not worth to add case for https://docs.scipy.org/doc/numpy-1.15.0/reference/generated/numpy.generic.html to cover future numpy type casts?

LeeTZ: It's good for now

Issue #7953, opened 2019-02-01
setColumnFrozenRight() method doesn't work


EfimovVladimir:
try http://localhost:8888/notebooks/doc/python/TableAPI.ipynb
we can drag and drop frozen column
![image](https://user-images.githubusercontent.com/17744188/52120690-2b668180-262e-11e9-8830-2d5276dbd154.png)
as result the frozen column will be on another place
![image](https://user-images.githubusercontent.com/17744188/52120648-02de8780-262e-11e9-9e86-84271923ea22.png)

rootty:
https://github.com/twosigma/beakerx/blob/b3448d62915a825933c36c7fb42f72c676ce4525/js/notebook/src/tableDisplay/dataGrid/interface/IDataGridModelState.ts#L26

^--- according this comment, the feature was dropped in purpose

EfimovVladimir:
@LeeTZ  If we don't need this functionality we should clean source code and example in http://localhost:8888/notebooks/doc/python/TableAPI.ipynb notebook

PR #8095, opened 2019-07-17
Remove column frozen right functionality from backend

rootty:
Implements:
setColumnFrozenRight() method doesn't work #7953


LeeTZ:
please resolve the conflicts here

PR #8101, opened 2019-08-05
jarek: add endless scrollbar in tabledisplay

jaroslawmalekcodete:


noobymcnoob:
@jaroslawmalekcodete I have a question about this change if you don't mind.

Is it intended to support the situation in which I replace the entire underlying dataframe's values? For example, a table goes from 0 rows to 15.

I'm finding that with this code, streaming is fine but changing the number of rows doesn't seem to work.

Issue #8120, opened 2019-09-02
Output widget doesn't work properly


EfimovVladimir:
see http://localhost:8888/notebooks/doc/groovy/OutputWidget.ipynb notebook
We expect that outputWidget displays the output after the cell with code
 ```
%import com.twosigma.beakerx.widget.Output
out = new Output()
OutputManager.setOutput(out)
```
but not after the second cell
![image](https://user-images.githubusercontent.com/17744188/64103425-e862bd00-cd7a-11e9-81c6-6fed72e7551a.png)

PR #8123, opened 2019-09-09
jarek/8112: (and #8120) fix output widget plus preview in spark

jaroslawmalekcodete:


EfimovVladimir:
I checked.
It looks good.

Issue #8134, opened 2019-09-24
Add support for adding SQL query overrides to associated databases


Ben-Epstein:
Certain databases that have internal optimizers such as Derby and Splicemachine allow for passing overrides to the SQL query via specific comment structures:
In Derby: pass in `--DERBY-PROPERTIES constraint=cons1` after the FROM clause
In Splicemachine: pass in `--SPLICE-PROPERTIES constraint=cons1` after the FROM clause

The proposed fix for this is to modify [QueryParser.java](https://github.com/twosigma/beakerx/blob/master/kernel/sql/src/main/java/com/twosigma/beakerx/sql/QueryParser.java#L87)

To check for a list of accepted "overrides" and set `boolean mustBeDeleted` (the third parameter) to false if so.

I am happy to make the change and create the PR if desired.

Ben-Epstein:
PR available [here](https://github.com/twosigma/beakerx/pull/8135)

PR #8135, opened 2019-09-25
Issue #8134 FIX

Ben-Epstein:
This is a PR fix to [issue#8134](https://github.com/twosigma/beakerx/issues/8134)

I've added an environment variable that defaults to True, which will perform the default behavior of a LineCommentParsingState, which is to delete the comment from the query string before submitting the query to the Database. This change now allows anyone using beakerx against a database that utilizes comments for overrides ([Derby](https://db.apache.org/derby/docs/10.6/tuning/ctunoptimzoverride.html) and/or [Splicemachine](https://doc.splicemachine.com/developers_tuning_queryoptimization.html#queryhints))

This PR will have no effect on the behavior of the current deployment.

Help from @jramineni


kernel/sql/src/main/java/com/twosigma/beakerx/sql/QueryParser.java
```diff
@@ -18,8 +18,14 @@
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import com.twosigma.beakerx.util.BeakerXSystem;
 
 public class QueryParser {
+  // Defaults to true -- deletes the LineComment
+  String rd = BeakerXSystem.getenv("REMOVE_DASH_LINECOMMENT");
```
jramineni: Make these static, since these variables are referenced in static class/methods

jramineni (changes requested):

PR #8137, opened 2019-09-25
Fix for issue#8134

Ben-Epstein:
This is a PR fix to [issue#8134](https://github.com/twosigma/beakerx/issues/8134)

I've added an environment variable that defaults to True, which will perform the default behavior of a LineCommentParsingState, which is to delete the comment from the query string before submitting the query to the Database. This change now allows anyone using beakerx against a database that utilizes comments for overrides (Derby and/or Splicemachine)

This PR will have no effect on the behavior of the current deployment.

Help from @jramineni


jramineni (approved):

Ben-Epstein:
Is there any chance we can have an authorized user check this code?

LeeTZ:
Looks good, thanks for your contribution Ben!

PR #8146, opened 2019-10-10
jarek/8131:  upgrade scala version to 2.12.8 and spark *2_12 v2.4.4

jaroslawmalekcodete:


EfimovVladimir:
Tested.
It works fine

PR #8215, opened 2020-01-07
jarek/8205: heatmaphighlighters for string type as it works for integer type

jaroslawmalekcodete:


EfimovVladimir:
checked, it works good

PR #8229, opened 2020-01-30
jarek/8228: load the latest version %classpath mvn add magic command  when no version

jaroslawmalekcodete:
…oad latest version of dep.


ildipo:
Repository was refactored

Issue #8116, opened 2019-08-25
Groovy Autocomplete based on Runtime Introspection?


ssadedin:
I've been browsing the code and notice that there is quite sophisticated support for Groovy autocomplete based on a fully integrated parser / lexer that can infer the static type of expressions and determine their methods etc.  Using this I can get autocomplete for some things eg:

![image](https://user-images.githubusercontent.com/138868/63646752-319f8500-c75b-11e9-9ee6-b7848244a915.png)

However I find that this almost never works for anything other than highly explicit cases where the types are either declared or explicitly inferrable through static analysis. However this is not how Groovy code is normally written so in practice this is unfortunately limited in how often it actually works. For example, I can get autocomplete to work on a String:

![image](https://user-images.githubusercontent.com/138868/63646772-7f1bf200-c75b-11e9-94d4-deacad9b39fc.png)

But not a map:

![image](https://user-images.githubusercontent.com/138868/63646800-cc985f00-c75b-11e9-81c2-26c11a2fac21.png)

I am not sure if it's a bug that this doesn't work or just a consequence of the current parser and how it statically infers types.

Either way, I'm posting this issue to query whether consideration has been given to returning autocomplete results based not just on static parsing of the cell, but supplementing that with dynamic introspection of the properties and methods of the object under the cursor. This would allow for a huge increase in the possible autocomplete results and, I think, make the Groovy kernel much easier for beginners to get started with.

There would be some obvious downsides to doing this. Primarily, dynamic introspection of the properties can actually cause code execution in certain cases and that in turn can end up with the kernel even hanging if the code does something crazy or is blocked unexpectedly eg: on I/O.

Overall though, I think this would be a win - it is, after all, the behavior that is implemented in `groovysh`.

Would be great to hear thoughts on (a) whether this is feasible, or (b) desirable. If it's thought to be so, I'd be happy to (at least try) and have a go at implementing it. Alternatively if all this is acutally supposed to be implemented already and it just needs some diagnosing or bug fixing, I'd be happy to look at that too!

PR #8236, opened 2020-02-11
improved groovy autocomplete based on reflection

ssadedin:
This change adds improved autocomplete functions for Groovy notebooks.

See #8116 for some of the limitations of the current autocomplete approach. This change is a partial implementation of #8116. It supplements the existing autocomplete options by adding entries that are determined through runtime introspection of the variable under the cursor. 

For example, in the scenario below, the existing autocomplete cannot offer any suggestions, however this change enables the below:

image

This is quite useful with the plotting functionality as valid plot attributes can be suggested now which makes them much easier to learn:

image

Apart from actual JavaBean properties, the implemenation also supports Groovy Map notation so that map entries are autocompleted as well:

image

Finally, some intelligence has been added so that when the object under the cursor is an iterable or Map, various suggestions for valid Groovy methods on those types are included in the completions:

image

It would be great to be able to have this integrated to BeakerX as I think it makes a big difference ot the usability of the Groovy kernel.

### DIFF
--- a/kernel/base/build.gradle
+++ b/kernel/base/build.gradle
@@ -30,7 +30,7 @@ compileJava {
 
 dependencies {
   compile group: 'com.opencsv', name: 'opencsv', version: '3.10'
-  compile group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: '2.6.5'
+  compile group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: '2.10.3'
   compile group: 'com.github.jupyter', name: 'jvm-repr', version: '0.3.1'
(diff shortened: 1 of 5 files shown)


ssadedin:
Hi, I have signed the CLA previously in the past, though perhaps it was too long ago?

Thanks.

ssadedin:
Just noting I have added more reflection based autocompletions based so that method names and type signatures on objects are now included in the autocompletion results as well as properties.

ssadedin:
One more update - the reflection based autocomplate can now introspect inside lists to suggest completions for list-based expressions, eg:

image

PR #8257, opened 2020-03-31
lukasz-jarek/8189:  support for %%spark magic command in python + new SparkUIForm widget

jaroslawmalekcodete:


EfimovVladimir:
was tested
it works good

PR #8280, opened 2020-05-08
jarek/8274: update to groovy 3

jaroslawmalekcodete:


ildipo:
Repository was refactored

PR #8283, opened 2020-05-17
fix #334 - copy and paste off by one column in table display

### LINKED ISSUE #334: Saving notebook removes letter from page title
I'm running beaker in Chromium on Linux Mint 16. Every time I make a change and then save my notebook, another letter is removed from the page title displayed in the tab and at the top of the browser. After all of the letters are gone, the title becomes the full address (e.g. "http://127.0.0.1:8801/beaker/#/open?uri=file:config%2Ftutorial.bkr"). If I then make another change and save again, the title is replaced with a single asterisk, which remains in place no matter how many subsequent changes and saves I make. Presumably this is a bug in removing the "unsaved changes" asterisk.

### LINKED ISSUE #8166: Clipboard Copy Off-by-one Column in BeakerX table
Using the Groovy kernel I make a table like this:

```groovy
[
    [foo: 1, bar:2],
    [foo: 10, bar:20]
] * 5
```
This displays like so:

image

I then select some values from the `foo` column and choose "Copy" from the menu:

image

Choosing Copy:

image

Then I paste the values into the next cell - but only empty values are pasted:

image

However if I also select the column to the left:

image

Then it pastes the values I wanted originally:

image

The same behavior happens for the `bar` column. To copy the values from `bar` I have to also select the values from `foo`, which then copies only the `bar` column and not `foo` and `bar`.

Expected behaviour: the column I select should be copied when only that column is selected.

I am using Chrome with a checkout and build from master (commit #774fa4)

ssadedin:
This change fixes https://github.com/twosigma/beakerx/issues/8166

ie: that when "Copy Selected Columns" is used the actual columns copied are offest by one and mising a column from those selected in the table.

Note:  I did have to modify the tests to accept new behavior as a result of this change. To the best of my understanding however, the new test spec is correct and the old was incorrect. That is, it tests copying columns 0 - 1, and it should expect 2 columns in the output, not one. So I have changed it to expect the updated values.

### DIFF
--- a/js/beakerx_tabledisplay/src/tableDisplay/dataGrid/cell/CellManager.ts
+++ b/js/beakerx_tabledisplay/src/tableDisplay/dataGrid/cell/CellManager.ts
@@ -268,7 +268,7 @@ export default class CellManager {
     };
 
     function getStartIndex(has_index) {
-        return (has_index) ?0:1;
+        return (has_index) ?1:0;
     }
 
     function exportCells(cells, exportOptions) {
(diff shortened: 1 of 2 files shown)


ildipo:
Repository was refactored

